PR #39, opened 2021-06-09
*: add build + test instructions to DEVELOPING.md

asubiotto:
I figured these out from looking at .github/workflows/go.yml. It probably makes
sense to add these to the DEVELOPING.md doc for now. I think it might make
sense to move these instructions to a separate CONTRIBUTING.md doc in the
future and hence separate these from instructions to build an encore app.


eandre:
Thanks so much @asubiotto! I didn't get a chance to review this when you submitted it as we didn't have a CLA in place. I've just done another pass on `CONTRIBUTING.md` and took your suggestions to heart. I'm going to go ahead and close this since I think the updated version should address these issues (see #51). Thank you!

PR #41, opened 2021-06-10
Add support for custom paths for rpcs

Willyham:
### Background
There is currently no way to receive a 'dynamic' webhook which uses path parameters. For example, Prometheus' AlertManager has a webhook to receive alert information, but the only the webhook path can be set, and the body content can't be changed. If you wanted to run a single webhook on encore which received events from multiple AlertManager webhooks, you'd need to pass some kind of parameter in the webhook URL.

### Change
This adds support for the concept of extra params in rpc directives, with the syntax `x=y`, for example: `//encore:api private foo=bar`.

The only 'well-known' key which does anything is the `path` key. This PR adds the ability to set a custom route for `public raw` endpoints only. To avoid router collisions, the route is appended to the default RPC path. For example:

```
package fooservice
//encore:api public raw path=/foo/:id
func Bar(...) { }
```

Would result in the endpoint being served at `fooservice.Bar/foo/:id`.


Willyham:
Note that I've **not** tested this outside of unit tests, as I don't have all of the encore stuff set up locally yet. If the feedback on the direction of the PR is positive, I can set that up :)

eandre:
Thanks, this is really great work! I've taken this and finished it up. I'll push a new PR based on this one.

PR #45, opened 2021-07-16
cli/internal/codegen(ts): don't include body for GET requests

vilhelmmelkstam:
GET requests don't support body. However, the generated typesciprt api client always includes a body. As a result, you get the error when trying to make a GET request using the generated client: 
```
Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.
```

I solved it by using the fact that `JSON.stringify(undefined) === undefined`. This means that if `req` is undefined, the request won't have any body.

Encore does complain if there's no body on a POST request. However, this won't be a problem since all POST requests from the API client always include params as body.

I'm of course willing to change the solution if you feel like another approach would fit better.

### DIFF
--- a/cli/internal/codegen/ts.go
+++ b/cli/internal/codegen/ts.go
@@ -310,7 +310,7 @@ func (ts *ts) writeBaseClient() {
         let response = await fetch(this.baseURL + path, {
             method: method,
             headers: this.headers,
-            body: JSON.stringify(req)
+            body: req && JSON.stringify(req)
         })
         if (!response.ok) {
             let body = await response.text()
@@ -323,7 +323,7 @@ func (ts *ts) writeBaseClient() {
         let response = await fetch(this.baseURL + path, {
             method: method,
             headers: this.headers,
-            body: JSON.stringify(req)
+            body: req && JSON.stringify(req)
         })
         if (!response.ok) {
             let body = await response.text()
--- a/cli/internal/codegen/ts.go
+++ b/cli/internal/codegen/ts.go
@@ -310,7 +310,7 @@ func (ts *ts) writeBaseClient() {
         let response = await fetch(this.baseURL + path, {
             method: method,
             headers: this.headers,
-            body: req && JSON.stringify(req)
+            body: req !== undefined ? JSON.stringify(req) : undefined
         })
         if (!response.ok) {
             let body = await response.text()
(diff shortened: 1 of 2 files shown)


vilhelmmelkstam:
Closed and reponed due to accidental removal of my branch, my mistake!

klemmster:
I just ran into the same issue. Anything I can do here to get this merged?

cli/internal/codegen/ts.go
```diff
@@ -310,7 +310,7 @@ func (ts *ts) writeBaseClient() {
         let response = await fetch(this.baseURL + path, {
             method: method,
             headers: this.headers,
-            body: JSON.stringify(req || {})
+            body: JSON.stringify(req)
```
eandre, line 313: This works but the behavior that `JSON.stringify(undefined) === undefined` is not very intuitive to me. If I saw this code somewhere "in the wild" I would have guessed that there's a bug lurking somewhere.

I think a semantically equivalent but clearer change would be to do `req !== undefined ? JSON.stringify(req) : undefined` or similar. That way the fact that we're dealing with the case when `req` is undefined is made explicit.

(Same goes for all the other places below)

eandre (changes requested):
Thanks so much for the PR, and thanks for the reminder @klemmster! I think the semantic change is great but I think we'd be better off handling the `req === undefined` case explicitly. See my comment for a more thorough explanation of what I mean.

cli/internal/codegen/ts.go
vilhelmmelkstam: Good point! I decided to go with the shorter but equal expression `req && JSON.stringify(req)`.

eandre: They're not equivalent; there are [lots of values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) that are falsy in JS. We only want the `undefined` value to indicate that the request should be sent without a body, not any falsy value.

Let's go with the explicit and "clearly correct" `req !== undefined ? JSON.stringify(req) : undefined` over trying to find a shorter, equivalent expression that will puzzle future readers as to what the intent is.

vilhelmmelkstam: Sure, forgot about those that were still JSON-stringifyable and falsy! New commit pushed!

eandre (approved):
Thank you!

PR #47, opened 2021-08-02
cli/daemon/dash/dashapp: add required `baseUrl` to tsconfig

vilhelmmelkstam:
Using this is required as per [TypeScript docs](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping). Having no `baseUrl` also broke intellisense in my editor (Webstorm).


eandre:
Great, thank you!

PR #48, opened 2021-08-02
cli/daemon/dash/dashapp: add environment variable for daemon address

vilhelmmelkstam:
This allows for development of the dashapp without having to rebuild the dashapp and encore or changing the source code.

I'm not completely happy with the `DEVELOPING.md` file, as I didn't feel that the section belonged anywhere good. So feel free to suggest improvments or changing it.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -58,3 +58,19 @@ The Encore Compiler rewrites the source code based on the parsed
 Encore Syntax Tree to create a fully functioning application.
 It rewrites API calls & API handlers, injects instrumentation
 and secret values, and more.
+
+## Dashapp
+
+To run the dashapp in development mode (using `npm run dev`), the dashapp needs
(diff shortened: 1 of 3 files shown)


eandre:
This is great, thank you! I had a realization that a simpler solution would be to enable the daemon to listen on a fixed port for development. We opted for a random port in order to enable it to always start without failing on "port already in use". But for actually developing the Encore daemon itself, we could change that behavior. Maybe easiest would be an environment variable `ENCORE_DAEMON_DEV=1` (or flag `encore daemon -dev`) that changes the behavior.

eandre:
I finally got around to making a pass on the `DEVELOPING.md` (now `CONTRIBUTING.md`) as well as making the improvements we discussed here. I'll go ahead and close this PR since it's covered by #51 already. Thanks so much!

PR #58, opened 2021-08-14
Updated wireguard/windows dependency to fix #57

### LINKED ISSUE #57: Error building cli/encore from source in windows 10.
1. When I am building encore cli in windows, getting compilation error `..\..\daemon\dash\server.go:25:12: pattern dashapp/dist/*: no matching files found`

encore-error

So to compile this, I have removed this line https://github.com/encoredev/encore/blob/67aefe20466f686561c300dd1b8d6276a563b2bc/cli/daemon/dash/server.go#L25

2. After it compiles, then when I run `./encore.exe -h` it panics.
encore-panic

This seems issue in wireguard/windows dependency. So after I update wireguard from v0.3.1 to v0.4, it works. But I feel like this can have impact somewhere else as well.
encore-mod

Please let me know what will be the permanent solution for both the issues, and if upgrading wireguard is only solution then I will pull request for it.

Thanks.

frzam:
Fixing panic while `encore.exe -h` on windows


eandre:
Thanks @frzam! I just noticed the CLA website doesn't properly handle the sign-in link. To sign the CLA you can log in by heading to https://cla.encore.dev/api/oauth-begin. I'll fix the issue shortly.

frzam:
@eandre Yup website is not working. It is not letting me [submit]( https://cla.encore.dev/sign)

eandre:
Did you log in and fill in all the information in the form (and click 'agree')? The form seems to work for me.

frzam:
Done. Was having trouble while submitting in mozilla.

eandre:
Oh, that's strange, will investigate. Thanks so much for your contribution!

frzam:
@eandre  The problem is in https://pkg.go.dev/golang.zx2c4.com/wireguard/device#NewDevice this takes `conn.Bind` as well. In their previous release https://pkg.go.dev/golang.zx2c4.com/wireguard@v0.0.20201118/device#NewDevice, it was as per how you have implemented. So they are doing breaking changes.

So I checked if we downgrade to latest **release** that is `v0.0.20201118` then it compiles on both *windows* and *linux*

eandre:
Thank you very much!

PR #59, opened 2021-08-15
Updated golang.zx2c4.com/wireguard Fixes #57

### LINKED ISSUE #57: Error building cli/encore from source in windows 10.

frzam:
Please refer https://github.com/encoredev/encore/issues/57

### DIFF
--- a/cli/internal/wgtunnel/wgtunnel_darwin.go
+++ b/cli/internal/wgtunnel/wgtunnel_darwin.go
@@ -16,6 +16,7 @@ import (
 	"syscall"
 	"time"
 
+	"golang.zx2c4.com/wireguard/conn"
 	"golang.zx2c4.com/wireguard/device"
 	"golang.zx2c4.com/wireguard/ipc"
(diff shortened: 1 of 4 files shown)


frzam:
Windows build:
win-build
Linux build:
linux-build

eandre:
Thank you!

PR #79, opened 2021-08-30
docs(auth): add example on returning custom error messages on auth handlers

### LINKED ISSUE #75: Returning custom error messages on auth handlers
Hi,

I have a auth handler that looks something like this:

```go
//encore:authhandler
func AuthHandler(ctx context.Context, token string) (auth.UID, *UserData, error) {
	response, err := GetUserInternal(ctx, &GetUserInternalParams{
		token: token,
	})
	if err != nil || response == nil {
		log.WithError(err).Warning("Failed to authenticate user")
		return "", nil, nil
	}

	if response.User.Status == model.UserStatus_Pending {
		log.Warning("Authentication failed, user is still pending")
		return "", nil, nil
	}
	if response.User.Status == model.UserStatus_Denied {
		log.Warning("Authentication failed, user is denied")
		return "", nil, nil
	}

	return auth.UID(strconv.FormatInt(response.User.ID, 10)), &UserData{
		ID:       response.User.ID,
	}, nil
}
```

I would like to be able to return custom unauthenticated messages for each of the three returns I have so a user knows why their auth failed. If I omit the error and return an empty string, the error message is something like "auth token missing", which is not very actionable for my users. If I do return the error, the 401 error turns into a 500 error with my message shown.

I could mode this code to my endpoints instead, but it's a lot of duplicated code. I would like to be able to do something like returning a `&errs.Error` to customize the return code from an auth handler error.

wisdommatt:
Signed-off-by: Wisdom Matthew <talk2wisdommatt@gmail.com>

resolves #75


docs/develop/auth.md
```diff
@@ -77,6 +76,27 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
 }
 ```
 
+
+## Returning custom error messages on auth handlers
```
eandre, line 80: This feels too specific; I think what we want to communicate is more along the lines of "here's how to signal that the request isn't authenticated".

How about:

```suggestion
## Handling auth errors
```

docs/develop/auth.md
```diff
@@ -77,6 +76,27 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
...
+
+When you want to return a custom error message in the auth handler e.g To return an `unauthenticated` error message:
```
eandre, line 82: In the same vein as above, I think we should set the stage for the reader and give a bit more detail:

```suggestion
When a token doesn't match your auth rules (for example if it's expired, the token has been revoked, or the token is invalid), you should return a non-nil error from the auth handler.

Encore passes the error message on to the user when you use [Encore's built-in error package](errors), so we recommend using that with the error code `Unauthenticated` to communicate what happened. For example: 
```

docs/develop/auth.md
```diff
@@ -77,6 +76,27 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
...
+    "encore.dev/beta/errs"
+)
+
+// AuthHandler can be named whatever you prefer (but must be exported).
+//encore:authhandler
+func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
+    // if the token is invalid/expired you can return a custom error message
+    // using the code below.
+    return "", errs.B().Code(errs.Unauthenticated).Msg("auth token is invalid").Err()
+}
```
eandre, line 96: I think we can simplify this example slightly. How about:

```suggestion
//encore:authhandler
func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
    return "", &errs.Error{
        Code: errs.Unauthenticated,
        Message: "invalid token",
    }
}
```

docs/develop/auth.md
```diff
@@ -77,6 +76,27 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
...
+```
+
+
```
eandre, line 99: Finally it would be great to add a callout on security. Something like this:

```suggestion
<Callout type="important">

Note that for security reasons you may not want to reveal too much information about why a request did not pass your auth checks. There are many subtle security considerations when dealing with authentication, and we don't have time to go into all of them here. Whenever possible we recommend using a third-party auth provider; see [Using Firebase Authentication](../how-to/firebase-auth) for an example of how to do that.

</Callout>

**Review by eandre — COMMENTED** [review]
Thanks so much @wisdommatt for working on this! I noticed the Callout box (line 34-36 in the file) is likely the source of the confusion, since it reads:

```
If your auth handler returns an error, Encore responds with a `500 Internal Server Error`.
To signify that the token is not valid, return the empty string (`""`) as the `auth.UID`.
Encore will then respond with `401 Unauthorized.
```

This used to be the case, but now that we have support for structured errors we added better semantics. I think we can remove the callout box now.

I think your approach with adding a new section is the right one. I added some suggestions to flesh it out even more, please take a look!

**Comment by wisdommatt** [comment]
@eandre thanks for the code review, i will implement them ASAP

File: docs/develop/auth.md (hunk 4 above)
Comment (wisdommatt) [reply docs/develop/auth.md:103]: @eandre i just pushed a commit to improve on the changes

File: docs/develop/auth.md [hunk 5]
```diff
@@ -90,13 +91,21 @@ import (
...
+        Message: "invalid token",
+    }
 }
+
+
 ```
 
 
+<Callout type="important">
+Note that for security reasons you may not want to reveal too much information about why a request did not pass your auth checks. There are many subtle security considerations when dealing with authentication and we don't have time to go into all of them here. Whenever possible we recommend using a third-party auth provider; see [Using Firebase Authentication](../how-to/firebase-auth) for an example of how to do that.
```
eandre, line 105: There needs to be a newline between `<Callout type="important">` and the content (and similarly between the content and the closing tag `</Callout>`) for MDX parsing reasons. Don't ask me why :)

docs/develop/auth.md
```diff
@@ -90,13 +91,21 @@ import (
...
+</Callout>
+
+
```
eandre, line 108: Minor nitpick: can remove the extra newline here

docs/develop/auth.md
```diff
@@ -76,10 +76,11 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
...
+## Handling auth errors
 
-## Returning custom error messages on auth handlers
+When a token doesn't match your auth rules (for example if it's expired, the token has been revoked, or the token is invalid), you should return a non-nil error from the auth handler.
 
-When you want to return a custom error message in the auth handler e.g To return an `unauthenticated` error message:
+Encore passes the error message on to the user when you use [Encore's built-in error package](errors), so we recommend using that with the error code `Unauthenticated` to communicate what happened. For example:
 
 ```go
 import (
```
eandre, line 86: I think this code sample can drop the imports and the `// AuthHandler can be named whatever you prefer (but must be exported).` since the previous code samples cover these.

eandre (commented):
Thanks so much, looks much better. A few more minor tweaks and then it looks ready!

docs/develop/auth.md
wisdommatt: Okay, noted

docs/develop/auth.md
wisdommatt: Okay

docs/develop/auth.md
wisdommatt: Okay

wisdommatt: `encore.dev/beta/errs` is not imported by other code samples ?

docs/develop/auth.md
```diff
@@ -77,6 +76,32 @@ func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
...
+import "encore.dev/beta/errs"
+
+//encore:authhandler
+func AuthHandler(ctx context.Context, token string) (auth.UID, error) {
+    return "", &errs.Error{
+        Code: errs.Unauthenticated,
+        Message: "invalid token",
+    }
+}
+
```
eandre, line 95: Remove these two empty lines; they're part of a code block so would show up in the fully rendered doc.

eandre (commented):
Thanks so much @wisdommatt for the patience. Just one more small change and this is ready to merge.

docs/develop/auth.md
wisdommatt: @eandre i have removed the empty lines

eandre (approved):
Looks great, thank you!

PR #81, opened 2021-09-01
docs(develop/databases): add instructions on handling dirty database migrations

### LINKED ISSUE #68: Document how to handle dirty database migrations
When a database migration doesn't apply cleanly, `golang-migrate` (the library we use under the hood to manage database migrations) marks the migration state as "dirty" and refuses to continue. This can be fixed by editing the database table (fixing the schema problem and then running `UPDATE schema_migrations SET dirty = false`), but that's not documented anywhere. We should add this to [SQL Database documentation](https://github.com/encoredev/encore/blob/main/docs/develop/databases.md).

wisdommatt:
Signed-off-by: Wisdom Matthew <talk2wisdommatt@gmail.com>

resolves #68

### DIFF
--- a/docs/develop/databases.md
+++ b/docs/develop/databases.md
@@ -2,7 +2,6 @@
 title: Using SQL databases
 subtitle: Provisioning, migrating, querying
 ---
-
 Encore treats SQL databases as logical resources.
 This means using a database only requires you to [define the schema](#defining-a-database-schema)
 and then start using it. Encore takes care of provisioning the database, running
@@ -64,7 +63,6 @@ The Encore CLI comes with built in support for this:
 * Use `encore db shell [--env=<name>] <service-name>` to open a [psql](https://www.postgresql.org/docs/current/app-psql.html)
   shell to the database for `<service-name>` in the given environment.
   Leaving out `--env` defaults to the local development environment.
-
 * Use `encore db proxy [--env=<name>]` to create a local proxy that forwards any incoming connection
   to the database in the given environment.
   Leaving out `--env` defaults to the local development environment.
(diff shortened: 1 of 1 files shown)


docs/develop/databases.md
```diff
@@ -2,7 +2,6 @@
 title: Using SQL databases
 subtitle: Provisioning, migrating, querying
 ---
-
```
eandre, line 5: Keep this please

docs/develop/databases.md
```diff
@@ -64,7 +63,6 @@ The Encore CLI comes with built in support for this:
 * Use `encore db shell [--env=<name>] <service-name>` to open a [psql](https://www.postgresql.org/docs/current/app-psql.html)
   shell to the database for `<service-name>` in the given environment.
   Leaving out `--env` defaults to the local development environment.
-
```
eandre, line 67: Keep this please

docs/develop/databases.md
```diff
@@ -85,3 +83,8 @@ In the cloud, how the database is provisioned depends on the type of [Encore Env
 Encore automatically provisions databases to match what your application requires.
 Simply define a database schema ([see above](#defining-a-database-schema)) and Encore
 will provision the database at the start of the next deploy.
+
+
```
eandre, line 87: Don't need a double newline here

docs/develop/databases.md
```diff
@@ -85,3 +83,8 @@ In the cloud, how the database is provisioned depends on the type of [Encore Env
...
+## Handling dirty database migrations
+
+When a database migration doesn't apply cleanly, `golang-migrate` (the library we use under the hood to manage database migrations) marks the migration state as "dirty" and refuses to continue. This can be fixed by editing the database table (fixing the schema problem and then running `UPDATE schema_migrations SET dirty = false`).
```
eandre, line 90: I think this section needs to be expanded on to provide more of a thorough explanation of what dirty migrations are, why they exist and so on. We don't need to mention `golang-migrate` since it's an implementation detail, but we should cover the semantics in greater detail.

eandre (commented):
Thanks so much for getting started on this. My main piece of feedback (listed in a review comment as well) is that I think we should expand on this section and explain in greater detail what dirty migrations are and so on.

If you want to take a stab at this that would be super appreciated!

docs/develop/databases.md
wisdommatt: Okay, i will do that ASAP

DomBlack:
An updated version of this is currently live, thanks for the contribution @wisdommatt

PR #94, opened 2021-09-15
cli: add --port and --listen options to encore run, change to port 4000

eandre:
Adds --port and --listen command line options to encore run.
Also updates the default port to 4000, and removes the behavior
to automatically find the first open port, as it proved disruptive
rather than helpful.


kyeett:
🥇

Issue #98, opened 2021-09-20
Running tests shouldn't require Docker with no database


kblake:
I newly installed encore and created a new default hello world application.
I went to run the tests via `encore test ./...` and the output says `pull docker image: exit status 1`.

I'm on MacOS running Go 1.15.6. 

Any reasons why I'm getting this vs a pass or fail message?

kblake:
Nevermind. I had to have Docker running in order for the tests to run. So I'm good now.

eandre:
Thanks so much @kblake for the feedback! You actually shouldn't need Docker for tests (unless your application uses a database, which is deployed in Docker for local development), so that's a bug since you used the Hello World example. Reopening and will fix, thank you!

mateusnobre:
Hi Andre, can I take this one? I just found the bug following the documentation, from what I understood, we have to:
- Make sure the test passes on applications that don't use databases (specifically, the hello world example) without need to have docker installed or running

I'm trying to understand the CLI code and where exactly I have to change things (I'm new to open source), if you can guide me.

I was paying special attention to this [file](https://github.com/encoredev/encore/blob/4bb4ad775bab0ff93defdc9e0c0b24027b9a2f70/cli/daemon/run/tests.go), am I right looking in the right place?

Issue #49, opened 2021-08-03
Encore daemon doesn't detect restarted database container


eandre:
The Encore daemon caches information about the running database container. This works fine most of the time but causes problems when the Docker container (or Docker itself) is restarted outside of Encore's control, leading to Encore trying to connect to the wrong (ephemeral) container port.

It leads to errors like:
```
error: could not connect to the database for service foo/: create db mapping: failed to connect to admin database: failed to connect to `host=0.0.0.0 user=encore database=postgres`: dial error (dial tcp 0.0.0.0:49153: connect: connection refused)
```

A workaround is to restart the encore daemon (by running `encore daemon`) to clear the cache, but the daemon should detect this by itself instead.

scastrec:
Thanks to this, It solve this error too:
`  ❌ Running database migrations... Failed: create db notes: failed to connect to admin database: failed to connect to `host=0.0.0.0 user=encore database=postgres`: dial error (dial tcp 0.0.0.0:61483: connect: connection refused)`

eandre:
Yes, the bugfix in #113 should fix that. Before we tag a new release with that bugfix you can work around the issue by running `encore daemon` from your terminal to restart the daemon (which clears the cache). Hope that helps :)

PR #113, opened 2021-12-06
cli/daemon: database improvements

### LINKED ISSUE #107: Quiting db proxy don't close assigned port
When opening a proxy to the DB using `encore db proxy`, the encore daemon opens a port. However, when closing this using `ctrl-c`, the port don't close. The only way I've found for closing the port is to restart the daemon using `encore daemon`. This is true for both when specifying a port using the `-p`-flag and when encore assigns a random port for me.

This is on macOS 11.6, I have not tested this on any other OS.

```
➜ encore db proxy        
dbproxy: listening for TCP connections on localhost:57849
^C%      
                                                                                                                                                                                                                               
➜ lsof -i:57849
COMMAND   PID     USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
encore  86031 --------   20u  IPv4     device_id_here      0t0  TCP localhost:57849 (LISTEN)
```

horiajurcut:
Fixes #107 #98 #49

### DIFF
--- a/cli/daemon/run.go
+++ b/cli/daemon/run.go
@@ -236,13 +236,7 @@ func (s *Server) Test(req *daemonpb.TestRequest, stream daemonpb.Daemon_TestServ
 	}
 	s.cacheAppRoot(man.AppID, req.AppRoot)
 
-	var usesSQLDB bool
-	for _, svc := range parse.Meta.Svcs {
-		if len(svc.Migrations) > 0 {
-			usesSQLDB = true
-			break
-		}
-	}
+	usesSQLDB := requiresSQLDB(parse.Meta)
 	clusterID := man.AppID + "-test"
 
 	// Set up the database asynchronously since it can take a while.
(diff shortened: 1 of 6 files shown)


cli/daemon/run.go
```diff
@@ -236,38 +236,54 @@ func (s *Server) Test(req *daemonpb.TestRequest, stream daemonpb.Daemon_TestServ
 	}
 	s.cacheAppRoot(man.AppID, req.AppRoot)
 
-	setupCtx, setupCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	var usesSQLDB bool
+	for _, svc := range parse.Meta.Svcs {
```
eandre, line 240: This can be replaced with `usesSQLDB := requiresSQLDB(parse.Meta)`

eandre (commented):
Looks good! One minor nitpick but otherwise ready to ship.

cli/daemon/run.go
horiajurcut: Missed it! Thanks. It's replaced now

eandre (approved):
Nice!

PR #115, opened 2021-12-12
docs: document path conflicts & how to avoid them

### LINKED ISSUE #112: Cannot have paths /user/me and /user/:id/posts at the same time
Currently, it's not possible to create the following endpoints at the same time:
```
method=GET path=/user/:id
method=GET path=/user/:id/posts
```

Paths with parameters as the first part are also conflicting with any other non-parametrical paths:
```
method=GET path=/users
method=GET path=/:username/posts
```

It'd be very nice to support these kinds of paths without conflicts, would it be possible?

ValeriaVG:
- Adds a note regarding path naming conflicts
- Adds recommendations on how to avoid conflicts by prefixing path parameters

Closes #112

### DIFF
--- a/docs/develop/services-and-apis.md
+++ b/docs/develop/services-and-apis.md
@@ -106,6 +106,21 @@ func UpdateBlogPost(ctx context.Context, id int, post *BlogPost) error {
 }
 ```
 
+<Callout type="important">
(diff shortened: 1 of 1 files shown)


eandre (approved):
Thanks, this is great. Both succinct and clear!

PR #117, opened 2022-01-19
docs: add instructions on how to handle Postgres errors

vilhelmmelkstam:
We ran into the problem of how to handle specific database errors in Encore and after some digging, we found a good solution. We thought others might be interseted as well and therefore wrote a short instruction on how to do it. 

If you feel like it should belong somewhere else or if shomething should change, please feel free to make any changes or let me know!


DomBlack:
Hey @melkstam we've since added [sqlerr](https://pkg.go.dev/encore.dev@v1.13.4/storage/sqldb/sqlerr) to the runtime, which provides this information without exposing driver specific data (Allowing Encore to upgrade it's driver without breaking your code).

So I'm going to close this for now, however please reopen it if it doesn't solve your issue

PR #119, opened 2022-02-09
Disable generic tests when not Go 1.18

DomBlack:
This commit uses Go build tags to disable two tests which now
rely on Go 1.18 generics which are causing test failures when we
test on Go 1.17 or below.


compiler/internal/codegen/type_resolver_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build go1.18
```
eandre, line 1: Are some of the tests applicable to Go 1.17? Would be nice to run those in that case.

Issue #95, opened 2021-09-17
Proposal: Cron support


eandre:
## Summary
We propose a simple mechanism for scheduling recurring tasks ("cron jobs").

## Motivation
Encore today provides great support for building APIs that execute some business logic in response to some external trigger (a user taking an action in a frontend, a webhook being called, and so on). But there's no built-in way to internally trigger something to happen. While there are many such use cases that we want to support natively, the simplest and most flexible is support for cron jobs.

## Design
Cron's original design is incredibly simple: specify when (or how frequently) a command should be executed. But when translated to building reliable systems, it turns out there are [many nuances](https://sre.google/sre-book/distributed-periodic-scheduling/):

* If execution fails, what failure mode is appropriate? Running the cron job more than once, or possibly not running it at all? (see [The Two Generals' Problem](https://en.wikipedia.org/wiki/Two_Generals%27_Problem))
* What to do about partial failures? What if a cron job performs two tasks, and one succeeds but the other fails?
* If a cron job is supposed to be launched every minute, and it takes more than a minute to complete, what should happen? Should two instances run concurrently? Should the second launch be skipped? Should the delay be measured from the stop time of the previous job, or from the start of it?

And so on. What seemed simple at the surface turned out to be quite complex. At the same time we'd like to get a solution out sooner rather than later, so it's important that the design choices are (1) reasonable defaults, and (2) forward-compatible to support different trade-offs above, over time.

To provide a portable implementation, we propose to start with a design that's based on the [Kubernetes CronJob](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/) design. Specifically to make the following assumptions:

* Jobs are expected to be idempotent, meaning it's possible the same job is run more than once
* If a job is still running by the time the next job is supposed to start, another instance is started and both jobs run concurrently

### Configuration
In order to define cron jobs, we propose to extend the `encore.app` configuration file (which is in JSON format) with a new `cron` key:

```js
"cron": [
	{
		// Required keys:
		"endpoint": "service.Endpoint",
		"schedule": "*/1 * * * *", // or "@hourly" etc

		// Optional keys (with defaults):
		"timezone": "UTC", // time zone the schedule is based on
		"prod": true,      // whether to run in prod environments
		"dev": true,       // whether to run in dev environments
		"local": true,     // whether to run for local development
	},
	...
]
```

[...]

PR #122, opened 2022-02-15
Add Cron Job support

horiajurcut:
This pull request adds cron job support, based on the original proposal in #95 but with some surface-level changes in terms of how cron jobs are defined.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -754,10 +754,10 @@ func (p *parser) isCronIntervalAllowed(val int) (suggestion int, ok bool) {
 
 	if idx == len(allowed) {
 		return allowed[len(allowed)-1], false
-	} else if idx == 0 {
-		return allowed[0], false
 	} else if allowed[idx] == val {
 		return val, true
+	} else if idx == 0 {
+		return allowed[0], false
 	} else if abs(val-allowed[idx-1]) < abs(val-allowed[idx]) {
 		return allowed[idx-1], false
 	}


parser/parser.go
```diff
@@ -703,6 +718,41 @@ func (p *parser) validateApp() {
...
+		1, 2, 3, 4, 5, 6, 8, 9, 10, 12, 15, 16, 18, 20, 24, 30, 32, 36, 40, 45,
+		48, 60, 72, 80, 90, 96, 120, 144, 160, 180, 240, 288, 360, 480, 720, 1440,
+	}
+	idx := sort.SearchInts(allowed, val)
+
+	if allowed[idx] == val {
+		return val, true
+	} else if idx == len(allowed) {
+		return allowed[len(allowed)-1], false
+	} else if abs(val-allowed[idx-1]) < abs(val-allowed[idx]) {
```
eandre, line 756: `idx` could be `0` and then this would panic

parser/parser.go
```diff
@@ -542,6 +473,217 @@ SpecLoop:
...
+							if minutes < 1 {
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: duration must be one minute or greater, got %d", minutes)
+								return nil
+							} else if minutes > 24*60 {
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: duration must not be greater than 1440 minutes (1 day), got %d", minutes)
+							} else if suggestion, ok := p.isCronIntervalAllowed(int(minutes)); !ok {
+								suggestionStr := p.formatMinutes(suggestion)
+								minutesStr := p.formatMinutes(int(minutes))
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: 24 hour time range (from 00:00 to 23:59) needs to be evenly divided by the interval value (%s), try setting it to (%s)", minutesStr, suggestionStr)
+							}
```
eandre, line 589: These error cases should presumably `return nil` like the other error cases

parser/parser.go
```diff
@@ -542,6 +473,217 @@ SpecLoop:
...
+							minutes := dur / minute
+							if minutes < 1 {
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: duration must be one minute or greater, got %d", minutes)
+								return nil
+							} else if minutes > 24*60 {
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: duration must not be greater than 1440 minutes (1 day), got %d", minutes)
+							} else if suggestion, ok := p.isCronIntervalAllowed(int(minutes)); !ok {
+								suggestionStr := p.formatMinutes(suggestion)
+								minutesStr := p.formatMinutes(int(minutes))
+								p.errf(kv.Value.Pos(), "cron.JobConfig.Every: 24 hour time range (from 00:00 to 23:59) needs to be evenly divided by the interval value (%s), try setting it to (%s)", minutesStr, suggestionStr)
```
eandre, line 588: It would be helpful to provide the suggestion in the correct syntax instead of just "1 hour 30 mins" or whatever.

parser/parser.go
```diff
@@ -749,10 +752,12 @@ func (p *parser) isCronIntervalAllowed(val int) (suggestion int, ok bool) {
 	}
 	idx := sort.SearchInts(allowed, val)
 
-	if allowed[idx] == val {
-		return val, true
-	} else if idx == len(allowed) {
+	if idx == len(allowed) {
 		return allowed[len(allowed)-1], false
+	} else if idx == 0 {
+		return allowed[0], false
```
eandre, line 758: This could still be valid, if `allowed[0] == val`

eandre (approved):

PR #134, opened 2022-02-16
docs/develop: add cron job docs

horiajurcut:


docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type Response struct {
+	Message string
+}
+```
+
+<Callout type="important">
+
+We only support using infrastructure creation at the package level, any other call location will result in a compilation error.
```
DomBlack, line 61:
```suggestion
We only support creating cron jobs at the package level, any other call location will result in a compilation error.
```

I think we maybe need a separate page to talk about our infrastructure creation syntax. This page is specifically about cronjobs and people might get confused

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,89 @@
...
+
+</Callout>
+
+## Cron job configuration
+
+- `cron.NewJob(ID, ...)` - This ID allows you to rename the variable _or_ move the infrastructure resource without having it being destroyed/recreated. It would cause a compilation error to have two resources of the same type with the same ID within the app.
+- `Name` - The Cron job user friendly name, will auto complete based on the **ID** if not set.
+- `Every` - `time.Duration` that determines how often the Cron job is executed; <sup>\*</sup>*if you want more control over the execution schedule we support traditional [cron expressions](https://en.wikipedia.org/wiki/Cron#CRON_expression) by setting the `Schedule` field in the `JobConfig` struct; we support either **Every** or **Schedule** but not both*.
```
DomBlack, line 69: Should we do an example of this format too?

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,89 @@
...
+- `func Foo(ctx context.Context) (*Response, error)` – when you only return a response
+- `func Foo(ctx context.Context) error` – when you need neither request nor response data
+
+## Monitoring Cron jobs
+
+After deploying your app, all Cron jobs defined in code will be shown in your Encore dashboard.
+
+You'll be able to filter them by environment, as well as see if they're active or not, what's the status of their last execution and when are they scheduled to run next.
+
+![Cron Jobs UI](/assets/docs/cron.png)
```
DomBlack, line 89: Is this already committed? I don't see it in this PR

horiajurcut: This work is on the cron-jobs branch in the webapp2 repo

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,89 @@
...
+      Endpoint: Cron
+    }
+*/
+var _ = cron.NewJob("cronjobexample", cron.JobConfig{
+	Name:     "Cron Job Example",
+	Every:    5 * time.Minute,
+	Endpoint: Cron,
+})
+
+//encore:api public path=/cron
```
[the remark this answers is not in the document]
horiajurcut: This needs to be called by Google Cloud Scheduler, so it needs to be public.

docs/develop/cron-jobs.md
horiajurcut: Yes

docs/develop/cron-jobs.md
horiajurcut: The image as well. All the assets are in the `www` repo.

.gitignore
```diff
@@ -8,4 +8,7 @@
 /.vscode
 
 # Build artifact that must be placed alongside go files for Windows
-*.syso
\ No newline at end of file
+*.syso
+
+# JetBrains
+.idea
```
DomBlack, line 14: Just a heads up; you can add this to `~/.gitignore`

```
/.idea
```

And in `~/.gitconfig`
```
[core]
	excludesfile = /Users/dom/.gitignore
```

Then it will be automatically ignored in all your repo's

horiajurcut: ✅

docs/develop/cron-jobs.md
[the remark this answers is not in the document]
eandre: I believe we can make this work with private endpoints as we're not calling it directly from Google Cloud Scheduler but from our platform

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,107 @@
...
+---
+
+You can use a `CronJob` to run jobs on a time-based schedule. These automated jobs run similarly to [Cron](https://en.wikipedia.org/wiki/Cron) tasks.
+
+Cron jobs are useful for creating periodic and recurring tasks, like running a database query or sending emails. Cron jobs can also schedule individual tasks for a specific time, such as if you want to schedule a job for a low activity period.
+
+<Callout type="important">
+
+**Cron jobs** have limitations and idiosyncrasies:
+- In certain circumstances, a single cron job can create multiple jobs. Therefore, jobs should be **idempotent**.
```
eandre, line 12: It's not clear what it means for a single job to create multiple jobs. How about:
```suggestion
- In certain circumstances, a single execution can result in your cron job endpoint being called more than once in order to provide at-least-once semantics. As a result jobs should be **idempotent**, meaning calling the cron job endpoint multiple times should behave the same way as calling it once.
```

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,107 @@
...
+title: Cron Jobs
+---
+
+You can use a `CronJob` to run jobs on a time-based schedule. These automated jobs run similarly to [Cron](https://en.wikipedia.org/wiki/Cron) tasks.
+
+Cron jobs are useful for creating periodic and recurring tasks, like running a database query or sending emails. Cron jobs can also schedule individual tasks for a specific time, such as if you want to schedule a job for a low activity period.
+
+<Callout type="important">
+
+**Cron jobs** have limitations and idiosyncrasies:
```
eandre, line 11:
```suggestion
Keep in mind when using cron jobs in your application that:
```

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,107 @@
...
+- In certain circumstances, a single cron job can create multiple jobs. Therefore, jobs should be **idempotent**.
+- The actual job needs to be expressed as a **public** API endpoint, that doesn't have any request data.
+- You can uniquely identify a Cron job resource using its ID; this makes it easier to rename or move infrastructure around without recreating or destroying it.
+
+</Callout>
+
+## Defining a Cron job
+
+With Encore you define cron jobs directly in your code, using our infrastructure declaration syntax.
```
eandre, line 20:
```suggestion
Just like other Encore concepts, defining a cron job is done straight in your application code. There's no infrastructure to manage — the Encore Platform automatically manages, schedules, executes and monitors the cron jobs for you. And it works exactly the same way across all the cloud providers we support.
```

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,107 @@
...
+  You can also add code snippets in your description,
+  which will be picked up by our parser and displayed
+  in our web UI.
+
+    var cfg := cron.JobConfig{
+      Name:     "Cron Job Example",
+      Every:    5 * time.Minute,
+      Endpoint: Cron
+    }
+*/
```
eandre, line 44: The comment is a bit meta; it talks about itself and the code snippet example is an example of itself. Let's turn this whole example into a real-world example. https://jvns.ca/blog/2021/07/08/writing-great-examples/ has some great ideas on this topic :)

docs/develop/cron-jobs.md
```diff
@@ -0,0 +1,107 @@
...
+
+</Callout>
+
+## Cron job configuration
+
+- `cron.NewJob(ID, ...)` - This ID allows you to rename the variable _or_ move the infrastructure resource without having it being destroyed/recreated. It would cause a compilation error to have two resources of the same type with the same ID within the app.
+- `Name` - The Cron job user-friendly name, will auto complete based on the **ID** if not set.
+- `Every` - `time.Duration` that determines how often the Cron job is executed; <sup>\*</sup>*if you want more control over the execution schedule we support traditional [cron expressions](https://en.wikipedia.org/wiki/Cron#CRON_expression) by setting the `Schedule` field in the `JobConfig` struct; we support either **Every** or **Schedule** but not both*.
+```go
+// Define the cron job execution schedule using the Every field
```
eandre, line 74: Same thing with these examples; let's make them more realistic with real-world examples.

PR #155, opened 2022-03-18
compiler/runtime: update dependencies to fix go1.18 compat

eandre:
See https://github.com/json-iterator/go/releases/tag/v1.1.12


DomBlack (commented):
LGTM

PR #160, opened 2022-04-01
Add Azure Documentation

DomBlack:
This commit adds Azure to the list of cloud providers
and documents how to link Azure to an app as well as why
we require Owner level permissions rather than just Contributor
within the Azure subscription.

### DIFF
--- a/docs/deploy/own-cloud.md
+++ b/docs/deploy/own-cloud.md
@@ -13,7 +13,7 @@ This functionality also lets you easily deploy a hybrid or multi-cloud applicati
 Each [environment](/docs/deploy/environments] your application has can be configured to use a different cloud provider.
 
 <Callout type="important">
-Please note, that while Encore will provision infrastructure within your cloud account, we do not remove infrastructure
+Please note, that while Encore will provision infrastructure within your cloud account, for safety reasons Encore does not destroy infrastructure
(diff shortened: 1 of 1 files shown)


docs/deploy/own-cloud.md
```diff
@@ -10,13 +10,21 @@ This gives you enormous flexibility, letting you use Encore for improving your p
 while maintaining the existing trust relationship you have with your cloud provider of choice.
 This functionality also lets you easily deploy a hybrid or multi-cloud application, if desired.
 
+Each [environment](/docs/deploy/environments] your application has can be configured to use a different cloud provider.
+
+<Callout type="important">
+Please note, that while Encore will provision infrastructure within your cloud account, we do not remove infrastructure
+once it's no longer required. This means if you disconnect your app from your cloud provider or delete the environment
+within Encore, you will still have to manually remove the infrastructure that was created by Encore.
```
eandre, line 18: Perhaps:
```suggestion
Please note, that while Encore will provision infrastructure within your cloud account, for safety reasons Encore does not destroy infrastructure
once it's no longer required. This means if you disconnect your app from your cloud provider or delete the environment
within Encore, you will still have to manually remove the infrastructure that was created by Encore.
```

eandre (approved):

PR #162, opened 2022-04-02
CONTRIBUTING.md: Fix instructions for building from source

klemmster:
when following the instructions on a freshly checked out repository the compilation fails with

```bash
go build ./cli/cmd/encore
cli/daemon/dash/server.go:26:12: pattern dashapp/dist/*: no matching files found
```

since the `dashapp/dist` the command is trying to embedd isn't build yet and doesn't exist.
Running the dashapps build command creates the files and the compilation runs through


eandre (approved):
Nice, thank you very much for fixing this!

PR #163, opened 2022-04-04
Create opensource.md

marcuskohlberg:
Add information about Open Source to docs.


docs/community/opensource.md
```diff
@@ -0,0 +1,18 @@
+---
```
eandre, line 1: I would name the file `open-source.md` so that the URL becomes `/open-source` and not `/opensource` but looks good otherwise!

eandre (approved):

PR #165, opened 2022-04-06
compiler: support build tags, static linking

eandre:
Add support for specifying additional build tags and the ability
to static link the resulting binary. These changes enable building
more self-contained binaries that can be used in 'scratch' images,
reducing startup time and image sizes.


ekerfelt (approved):
Very nice!

PR #166, opened 2022-04-07
compiler: fix static linking toggle

eandre:
It needs to be broken up into two separate arguments
for the go tool to understand it properly.


ekerfelt (approved):
LGTM

PR #174, opened 2022-04-19
Improve local postgres setup

klemmster:
I've had a surprising failure trying to deploy my code with Postgres RowLevelSecurity(RLS) to encore cloud.

Since locally the encore user was running with super user privileges, where row level security is ignored, I created a new user/role which each connection would switch to, so that RLS would work.
Pushing this code to the cloud failed because the user running the migration isn't allowed to create users/roles.

This PR changes the local setup to closer the cloud based behavior:
a) there are two users now in the local setup: The `postgres` superuser and the `encore` 'create db' user
b) the version of postgres is updated to allow users with 'create db' permission to install trusted extensions. That feature was introduced in postgres 13, the cloud is running 14, local env was running 11

I don't expect this to be merged as this stage:
~~a) it's not the cleanest, the user creation code should only be run once, and not 'piggy backed' on the create db call.~~
b) the current migration path is: stop the cluster and delete the volumes, re-create the cluster with the encore cli so the permissions are in sync with the code. As is, it's not a hands-off experience, but rather breaking experience. The init scripts in the dockerfile won't run if a data directory is already present, so the encore user might or might not be there.

for a) ~~I'd think a cleaner solution would be a thin wrapper around the postgres image that contains an initialization script ( as described in https://hub.docker.com/_/postgres) to create the encore user during the `docker run` phase of the cluster creation.~~
The code is there already: https://github.com/klemmster/encore-postgres, also manually pushed it to dockerhub.
You seem to have a docker registry already, so I've created the PR to get the conversation rolling.

Please let me know what your thoughts are on this topic or if there are open questions.

### DIFF
--- a/cli/daemon/sqldb/cluster.go
+++ b/cli/daemon/sqldb/cluster.go
@@ -124,7 +124,7 @@ func (c *Cluster) Start(log runlog.Log) error {
 				"-d",
 				"-p", "5432",
 				"--shm-size=1gb",
-				"-e", "POSTGRES_USER=encore",
+				"-e", "POSTGRES_USER=postgres",
 				"-e", "POSTGRES_PASSWORD=" + c.ID,
 				"-e", "POSTGRES_DB=postgres",
 				"--name", cname,
--- a/cli/daemon/sqldb/db.go
+++ b/cli/daemon/sqldb/db.go
@@ -198,7 +198,7 @@ func (db *DB) connectAdminDB(ctx context.Context) (*pgx.Conn, error) {
 	var err error
 	for i := 0; i < 40; i++ {
 		var conn *pgx.Conn
-		conn, err = pgx.Connect(ctx, "postgresql://encore:"+db.Cluster.ID+"@"+hostPort+"/postgres?sslmode=disable")
+		conn, err = pgx.Connect(ctx, "postgresql://postgres:"+db.Cluster.ID+"@"+hostPort+"/postgres?sslmode=disable")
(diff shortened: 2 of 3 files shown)


klemmster:
@eandre Could you have a look at this (or nudge the right person of the encore team). For some reason, I can't request a reviewer in githubs interface.
I'd love to get some feedback and see how we can move this forward so I can stop using my custom build again :)

eandre:
Sorry @klemmster for dropping the ball on this, that's entirely my bad. Let me have a look!

eandre:
@klemmster, I just pushed #211 which addresses a series of long-standing issues with Encore's SQLDB handling. It includes setting up a more best-practices setup for roles and permissions, including breaking down the roles into:

* `encore-admin` — admin user, for running migrations
* `encore-write` — write user, for the running service
* `encore-read` — read user, for read-only use cases

As a starting point Encore will only use the `encore-admin` to minimize the difference with the existing setup.

Can you please have a look and see if this addresses your concerns? I opted to do this as part of the bootstrapping inside the Encore daemon as opposed to a separate Docker image as it would require us to maintain and keep another image updated.

klemmster:
@eandre I won't have a chance to really check/test it for a few days.
The PR looks like a big improvement on a quick glance. 
I'll have to adjust the RLS user on my end, but that's not a big deal. 
Thank you so much :+1: 

(If it unexpectedly doesn't work out, I'll just call out for '#help' in slack :) )

PR #178, opened 2022-04-20
testing fixes

eandre:
- compiler: bring test config in line with build
- compiler/internal/codegen: ensure runtime gets imported
- parser: support _test packages


DomBlack (approved):
Looks great!

PR #180, opened 2022-04-20
cli: compiler: runtime: merge cloud runtime and local runtime

eandre:
This commit combines the two runtime implementations Encore has provided
into a single runtime. This is done in order to simplify future
development by reducing the amount of duplication necessary to implement
new functionality.

APIs internal to Encore are hidden from the API docs through the use
of the "encore_internal" build tag, which the compiler now always
provides. Additionally, in order to switch on/off certain functionality
for local development only there is now a "encore_local" build tag
which is passed in by default by the daemon when compiling for
local development.

Finally to bring the runtimes to parity with each other we now
generate auth keys in the daemon when starting up the app locally,
in order to exercise the production code paths. Similarly the local
runtime outputs JSON logs instead of preformatted console logs,
and the conversion to console formatting happens in the CLI.


runtime/go.mod
```diff
@@ -3,25 +3,40 @@ module encore.dev
 go 1.18
 
 require (
-	github.com/felixge/httpsnoop v1.0.2
+	github.com/felixge/httpsnoop v1.0.1
```
DomBlack, line 6: It looks like we've done some downgrades in here; we should stick to the newer versions of `httpsnoop` and `pgx`, `pgconn` & `pgproto3`

DomBlack (dismissed):
Small nit comment; but otherwise looks good

DomBlack (approved):

Issue #171, opened 2022-04-16
Proposal: Support retrieving application and request metadata


eandre:
We've had several requests for more metadata about both the running Encore application and its environment, as well as about the currently running request. This proposal provides the suggested API for accomplishing these goals. Feedback is very much appreciated!

## Proposed API

```go
package encore // import "encore.dev"

// CurrentOp reports information about the current operation being handled
// by the calling goroutine. It never returns nil.
//
// It is safe to call concurrently from multiple goroutines and correctly handles
// the same server processing multiple concurrent requests.
func CurrentOp() *Operation

// Operation represents an incoming HTTP request or API call being handled by Encore.
type Operation struct {
	// Type is the type of operation this represents.
	Type OperationType
	// Service is the name of the service being called.
	Service string
	// Endpoint is the name of the endpoint being called.
	Endpoint string
	// Start is the time when the API endpoint was called to handle the request.
	Start time.Time
	// Path is the HTTP request path for the request.
	Path string
	// PathParams are the path parameters parsed from the request path.
	PathParams PathParams
	// Payload is the request payload parsed from the request body.
	// It is of the same type as the API endpoint's payload parameter.
	Payload any
}

type OperationType string

const (
    OpNone    OperationType = "none"
    OpAPICall OperationType = "api-call"
)

// PathParams contains the path parameters parsed from the request path.
type PathParams []PathParam

// PathParams represents a parsed path parameter.
type PathParam struct {
	Name  string // the name of the path parameter, without leading ':' or '*'.
	Value string // the parsed path parameter value.
}

// Get returns the value of the path parameter with the given name.
// If no such parameter exists it reports "".
func (p PathParams) Get(name string) string

// Meta reports metadata about the running Encore application.
// It never returns nil.
func Meta() *Metadata

// Metadata contains metadata about the running Encore application.
type Metadata struct {
	// AppID is the ID of the Encore application.
	// If the application is not linked to the Encore Platform it is "".
	AppID string

	// APIBaseURL is the base URL for calling this API.
	//
	// For local development it is "http://localhost:<port>", typically "http://localhost:4000".
	//
	// If a custom domain is used for this environment it is returned here, but note that
	// changes only take effect at the time of deployment while custom domains can be updated at any time.
	APIBaseURL string

	// EnvName is the name of the environment.
	// For local development it is "local".
	EnvName string

	// EnvType is the type of environment.
	EnvType EnvType

	// Cloud is the cloud provider this is running in.
[...]

PR #198, opened 2022-05-04
Add App & Request Metadata

DomBlack:
This adds the beginning of Proposal #171 which adds the ability
for running Encore applications to retrieve metadata about the application
itself and the current request.

In this commit we've added the code for returning the `CurrentOp()`, however
the application metadata call (`Meta()`) is still a work-in-progress.

I've also merged the two runtime config files together, by getting the compiler
to use the combined runtime. This allows us to ensure that the two config's structs
can never get out of sync; as now there is just one!


cli/daemon/run/run.go
```diff
@@ -457,10 +467,18 @@ func (r *Run) generateConfig(p *Proc, params *startProcParams) *config.Runtime {
...
 	return &config.Runtime{
-		AppID:         r.ID,
-		AppSlug:       r.AppSlug,
+		AppID:   r.ID,
+		AppSlug: r.AppSlug,
+		APIBaseURL: url.URL{
+			Scheme: "http",
+			Host:   fmt.Sprintf("localhost:%d", params.RuntimePort),
+		},
+		DeployID:      fmt.Sprintf("clidpl_%s", xid.New()),
```
DomBlack, line 476: We use `dpl` for deployments from our platform, so I've used `clidpl` to indicate this is a CLI deployment.

The reason this is set, rather than blank is so we more closely mimic a cloud-deployed application

cli/daemon/run/run.go
```diff
@@ -261,13 +269,15 @@ func (r *Run) buildAndStart(ctx context.Context, parse *parser.Result) (p *Proc,
...
-		CgoEnabled:        true,
-		EncoreRuntimePath: env.EncoreRuntimePath(),
-		EncoreGoRoot:      env.EncoreGoRoot(),
-		Parse:             parse,
-		BuildTags:         []string{"encore_local"},
+		Revision:              parse.Meta.AppRevision,
+		UncommittedChanges:    parse.Meta.UncommittedChanges,
+		WorkingDir:            r.params.WorkingDir,
+		CgoEnabled:            true,
+		EncoreCompilerVersion: fmt.Sprintf("EncoreCLI/%s", version.Version),
```
DomBlack, line 276: This format mirrors our existing User-Agent

cli/daemon/run/run.go
eandre: We didn't get a chance to discuss this one on Friday, but how about `run_%s` instead for local dev? `clidpl` is a bit terse (and a real consonant soup!)

runtime/meta.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+		Deploy: DeployMeta{
+			ID:   config.Cfg.Runtime.DeployID,
+			Time: config.Cfg.Runtime.DeployedAt,
+		},
+	}
+}
+
+func init() {
+	if config.Cfg != nil {
```
eandre, line 37: I assume this check is only to work outside of Encore apps? Maybe worth documenting

runtime/meta.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// EnvDevelopment represents a long-lived cloud-hosted, non-production environment, such as test environments.
+	EnvDevelopment EnvironmentType = "development"
+
+	// EnvEphemeral represents short-lived cloud-hosted, non-production environments, such as preview environments
+	// that only exist while a particular pull request is open.
+	EnvEphemeral EnvironmentType = "ephemeral"
+
+	// EnvLocal represents the local development environment when using 'encore run' or `encore test`.
+	EnvLocal EnvironmentType = "local"
```
eandre, line 117: Thoughts on removing the `Env` prefix on these? They seem specific enough to be unlikely to clash with something else down the line

runtime/request_impl.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"time"
+
+	"encore.dev/runtime"
+)
+
+var applicationStartTime = time.Now()
+
+// CurrentRequest returns the Request that is currently being handled by the calling Go routine
```
eandre, line 13:
```suggestion
// CurrentRequest returns the Request that is currently being handled by the calling goroutine
```

runtime/request_noop.go
```diff
@@ -0,0 +1,13 @@
+//go:build !encore_internal
```
eandre, line 1: Should we remove this tag for now?

runtime/runtime/config/config.go
```diff
@@ -94,6 +112,14 @@ type SQLServer struct {
...
 	Host string `json:"host"`
 
+	// MinConnections is the minimum number of open connections to use
+	// for this database. If zero it defaults to 2.
+	MinConnections int `json:"min_connections"`
+
+	// MaxConnections is the maximum number of open connections to use
+	// for this database. If zero it defaults to 30.
+	MaxConnections int `json:"max_connections"`
+
```
eandre, line 122: Was this moved by accident? They belong on the database struct

runtime/runtime/request.go
```diff
@@ -30,6 +33,7 @@ var (
 )
 
 var json = jsoniter.Config{
+	IndentionStep:          2,
```
eandre, line 36: What does this change do?

runtime/runtime/request.go
```diff
@@ -51,21 +55,23 @@ type Type byte
 
 const (
 	RPCCall     Type = 0x01
-	AuthHandler Type = 0x02
+	AuthHandler      = 0x02
```
eandre, line 58: Should this not be using `Type`?

runtime/runtime/rtc.go
```diff
@@ -1,3 +1,5 @@
+//go:build encore_internal
```
eandre, line 1: I'd say we should revert these tags and add them in later when we figure out how to merge the `encore.dev` module into the runtime

runtime/runtime/config/config.go
DomBlack: I had a merge conflict this morning when rebasing against main and it looked like they had been moved into the server struct.

Will move them back

runtime/runtime/request.go
DomBlack: Changes the formatting of the output to be pretty formated, rather than minimal (i.e. so the output looks the same as our hand written JSON responses)

runtime/meta.go
DomBlack: There are two `Local`'s already in this package, `CloudLocal` and `EnvLocal`.

I'd prefer to keep the prefix tbh as then it doesn't become confusing what `Local` and `CloudLocal` are

DomBlack:
> We didn't get a chance to discuss this one on Friday, but how about run_%s instead for local dev? clidpl is a bit terse (and a real consonant soup!)

I'm ok with this change, - I was just going for a `cli` prefix on the platform's ID prefix, so it's still relatable but clear it's differently scoped. Will however go for `run`

eandre (approved):

eandre (dismissed):

eandre (dismissed):

eandre (dismissed):

runtime/runtime/setup.go
```diff
@@ -146,12 +151,27 @@ func (srv *Server) scrapeMetrics(w http.ResponseWriter, req *http.Request) {
...
+		Code    string `json:"code"`
+		Message string `json:"message"`
+		Details any    `json:"details"`
+	}{
+		Code:    "ok",
+		Message: "Your Encore app is up and running!",
+		Details: struct {
+			AppRevision    string `json:"app-revision"`
+			EncoreCompiler string `json:"encore-compiler"`
+			DeployId       string `json:"deploy-id"`
```
eandre, line 167: JSON case convention is either `snake_case` or `camelCase` and looking through our code base we're primarily using `snake_case`.

```suggestion
			AppRevision    string `json:"app_revision"`
			EncoreCompiler string `json:"encore_rcompiler"`
			DeployId       string `json:"deploy_id"`
```

eandre (approved):

PR #199, opened 2022-05-05
runtime: add graceful shutdown support

eandre:
Serverless environments such as Cloud Run that frequently scale up
and down can cause unclean PostgreSQL connection shutdowns.

The PostgreSQL server can take a long time (many minutes) to detect this
and release the connection slow. This can lead to out-of-connection
errors even when there are no active connections.

Fix this by handling SIGINT/SIGTERM signals and gracefully shutting down
when detected.

### DIFF
--- a/runtime/runtime/config/config.go
+++ b/runtime/runtime/config/config.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
 )
@@ -58,6 +59,11 @@ type Runtime struct {
 	TraceEndpoint string          `json:"trace_endpoint"`
 	AuthKeys      []EncoreAuthKey `json:"auth_keys"`
 	SQLDatabases  []*SQLDatabase  `json:"sql_databases"`
+
+	// ShutdownTimeout is the duration before non-graceful shutdown is initiated,
+	// meaning connections are closed even if outstanding requests are still in flight.
+	// If zero, it shuts down immediately.
+	ShutdownTimeout time.Duration `json:"shutdown_timeout"`
 }
 
 func (r *Runtime) Copy() *Runtime {
--- a/runtime/runtime/shutdown.go
+++ b/runtime/runtime/shutdown.go
@@ -22,6 +22,16 @@ var shutdown = struct {
 	completed: make(chan struct{}),
 }
 
+// RegisterShutdown registers a shutdown handler that will be called when the server
+// is gracefully shutting down.
+//
+// The given context is closed when the graceful shutdown window is closed and it's
+// time to forcefully shut down. force.Deadline() can be inspected to learn when this
+// will happen in advance.
+//
+// The shutdown is cooperative: the process will not exit until all shutdown handlers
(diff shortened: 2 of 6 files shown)


runtime/runtime/config/config.go
```diff
@@ -58,6 +59,11 @@ type Runtime struct {
 	TraceEndpoint string          `json:"trace_endpoint"`
 	AuthKeys      []EncoreAuthKey `json:"auth_keys"`
 	SQLDatabases  []*SQLDatabase  `json:"sql_databases"`
+
+	// ShutdownTimeout is the duration before non-graceful shutdown is initiated,
+	// meaning connections are closed even if outstanding requests are still in flight.
+	// If zero, it shuts down immediately.
+	ShutdownTimeout time.Duration `json:"shutdown_timeout"`
```
DomBlack, line 66: Are we defaulting to zero?

runtime/runtime/shutdown.go
```diff
@@ -0,0 +1,76 @@
...
+	once      sync.Once     // to trigger shutdown logic only once
+
+	mu       sync.Mutex
+	handlers []func(force context.Context)
+}{
+	initiated: make(chan struct{}),
+	completed: make(chan struct{}),
+}
+
+func RegisterShutdown(fn func(force context.Context)) {
```
DomBlack, line 25: Could you document this func please.

It’s not clear what the “force” context is for or why it’s called force

runtime/storage/sqldb/sqldb.go
```diff
@@ -262,6 +261,17 @@ func getPool(name string) *pgxpool.Pool {
 	if err != nil {
 		panic("sqldb: setup db: " + err.Error())
 	}
+	runtime.RegisterShutdown(func(force context.Context) {
+		done := make(chan struct{}, 1)
+		go func() {
+			pool.Close()
+			close(done)
+		}()
+		select {
```
DomBlack, line 270: I get why here, but this is a bit of an annoying design to hve to do this for each registered shutdown function.

DomBlack (changes requested):

runtime/storage/sqldb/sqldb.go
eandre: Yeah it is. The alternative is for the `doShutdown` function to just ignore the fact that the shutdown handlers are still running and just proceed with shutting down after the timeout. That would be less code, but also worse? In this way the code can decide how to best shutdown gracefully.

Note that not every use case is doing this; the HTTP server doesn't need to. It depends on the semantics of the API. In the two sqldb-related cases there is no way of specifying the deadline and the methods block potentially forever, hence this code.

runtime/runtime/shutdown.go
eandre: Good point, will do

runtime/runtime/config/config.go
eandre: Currently yes, will add code on the platform side to specify a default. Originally I had a default of 10s if this is zero, but then you can't actually specify a timeout of zero (unless we change the field to be a `*time.Duration` and use `nil` to represent the default of 10s).

DomBlack (approved):

PR #203, opened 2022-05-07
cli/cmd/go-encore-wrapper: new command

eandre:
This adds a new command, go-encore-wrapper, that is designed
to wrap the "go" command and forward certain commands to
the "encore" CLI, like running tests.

The purpose is to build a better integration with tools like
vscode-go that only support replacing the "go" tool with
another tool that has the same command structure.

With this change, users will be able to configure vscode-go
to use Encore when appropriate simply by defining:

    "go.alternateTools": {"go": "go-encore-wrapper"}


DomBlack (approved):

eandre:
Let's hold off on merging this until we know more about where we're going with https://github.com/golang/vscode-go/issues/2238.

PR #207, opened 2022-05-10
compiler/internal/codegen: fix raw endpoint wrappers

eandre:
The generated wrappers were emitting bad code for raw endpoints.

This primarily showed up in tests, as it's rare for one endpoint
to call another raw endpoint.

Fix the codegen and add tests to catch any future regressions.


eandre:
This turned out to be trickier than expected, due to our new support for path parameters and how that interacts with raw endpoints. Closing this for now; will re-open when I've had a think on how to best handle this.

PR #212, opened 2022-05-13
Update development dashboard branding.

DomBlack:
This commit updates the dashboard branding to be in line with the new Encore branding.

It also adds a link to the Encore documentation website from the navigation menu.

# Screenshots

image

image


DomBlack:
> I think the full logo is a bit too much on the desktop view

I disagree; I think the purple button on the API explorer is the focus. But I've swapped it back in the desktop mode away 😄

eandre (approved):

PR #213, opened 2022-05-13
parser/encoding: Add request payload utility

ekerfelt:
Adds a utility to describe how an RPC request payload should be encoded

### DIFF
--- a/parser/encoding/request.go
+++ b/parser/encoding/request.go
@@ -1,0 +1,138 @@
+package encoding
+
+import (
+	"github.com/cockroachdb/errors"
+
+	"encr.dev/parser"
+	v1 "encr.dev/proto/encore/parser/schema/v1"
+)
+
+// ParameterLocation is the request/response home of the parameter
+type ParameterLocation int
+
+const (
+	Header ParameterLocation = iota // Parameter is placed in the HTTP header
+	Query                           // Parameter is placed in the query string
+	Body                            // Parameter is placed in the body
+)
+
+// tagToLocation is used to translate struct field tags to param locations
+// if overrideDefault is set, the default HTTP param location will be changed
+// if the tag matches the paramLocation, the param name will be override with the
+// tag name
+var tagToLocation = map[string]struct {
+	location        ParameterLocation
+	overrideDefault bool
+}{
+	"query": {
+		location:        Query,
+		overrideDefault: true,
+	},
+	"qs": {
+		location:        Query,
+		overrideDefault: true,
+	},
+	"header": {
+		location:        Header,
+		overrideDefault: true,
+	},
+	"json": {
+		location:        Body,
+		overrideDefault: false,
+	},
+}
+
+// RequestEncoding expresses how a request should be encoded for a specified set of HTTPMethods
+type RequestEncoding struct {
+	// The HTTP methods these field configurations can be used for
+	HTTPMethods []string
+	// Contains metadata about how to marshal a HTTP parameter
+	Fields []*ParameterEncoding
+}
+
+// ParameterEncoding expresses how a parameter should be encoded
+type ParameterEncoding struct {
+	// Location (e.g. header, query, body) where the parameter should be marshalled from/to
+	Location ParameterLocation
+	// The location specific name of the parameter (e.g. cheeseEater, cheese-eater, X-Cheese-Eater
+	Name string
+	// The underlying field definition
+	Field *v1.Field
+}
+
+// DescribeRequest combines a payload v1.Struct and a list of HTTP methods and returns
+// a list of RPCRequest metadata. RPCRequest specifies which of the requested HTTP method
+// it can be used for.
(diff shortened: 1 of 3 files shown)


parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"github.com/cockroachdb/errors"
+
+	"encr.dev/parser"
+	v1 "encr.dev/proto/encore/parser/schema/v1"
+)
+
+// ParamLocation is the request/response home of the parameter
+type ParamLocation int16
```
eandre, line 11: This can probably be just `int`; the efficiency doesn't really matter and `int16` tricks you into thinking the size matters

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+
+const (
+	Header = 1 // Parameter is placed in the http header
+	Query  = 2 // Parameter is placed in the query string
+	Body   = 3 // Parameter is placed in the body
```
eandre, line 16: These should specify the `ParamLocation` type (`Header ParamLocation = 1` etc) to make them typed constants

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+	"json": {
+		location:        Body,
+		overrideDefault: false,
+	},
+}
+
+// RequestEncoding expresses how a request should be encoded for a specified set of HttpMethods
+type RequestEncoding struct {
+	// The HTTP methods these field configurations can be used for
+	HttpMethods []string
```
eandre, line 48: In Go terms like HTTP and ID should be all uppercase (`HTTP`, `ID`, not `Http`/`Id`), so `HTTPMethods`

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+	// Contains metadata about how to marshal a http parameter
+	Fields []*ParameterEncoding
+}
+
+// ParameterEncoding expresses how a parameter should be encoded
+type ParameterEncoding struct {
+	// Location (e.g. header, query, body) where the parameter should be marshalled from/to
+	Location ParamLocation
```
eandre, line 56: A bit strange to see `ParameterEncoding` but `ParamLocation`. Probably `Param` is clear enough that we should use that instead of `Parameter` everywhere?

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+	// The location specific name of the parameter (e.g. cheeseEater, cheese-eater, X-Cheese-Eater
+	ParamName string
```
eandre, line 58: Since the type is `ParameterEncoding` (or `ParamEncoding`) this can probably be just `Name`

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
+package encoding
```
eandre, line 1: A few tests would be nice!

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+	default:
+		return name
+	}
+}
+
+// describeParam returns an RPCField which falls back on defaultLocation if no parameter
+// (e.g. qs, query, header) tag is set
+func describeParam(defaultLocation ParamLocation, field *v1.Field) (*ParameterEncoding, error) {
+	rpcField := &ParameterEncoding{defaultLocation, formatName(defaultLocation, field.Name), field}
+	var usedTag string
```
DomBlack, line 120: Maybe rename to be clearer you didn't intend to set it outside of the overrideDefault?

```suggestion
	var usedOverrideTag string
```

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+	ParamName string
+	// The underlying field definition
+	Field *v1.Field
+}
+
+// DescribeRequest combines a payload v1.Struct and a list of http methods and returns
+// a list of RPCRequest metadata. RPCRequest specifies which of the requested http method
+// it can be used for.
+func DescribeRequest(payload *v1.Struct, httpMethods ...string) ([]*RequestEncoding, error) {
+	methodsByLocation := make(map[ParamLocation][]string)
```
DomBlack, line 67: Can we add the word default here?

```suggestion
	methodsByDefaultLocation := make(map[ParamLocation][]string)
```

parser/encoding/request.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"encr.dev/parser"
+	v1 "encr.dev/proto/encore/parser/schema/v1"
+)
+
+// ParamLocation is the request/response home of the parameter
+type ParamLocation int16
+
+const (
+	Header = 1 // Parameter is placed in the http header
```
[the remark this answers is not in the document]
ekerfelt: 👍

parser/encoding/request.go
ekerfelt: Indeed! Can I do a follow up PR to not block mr Black?

eandre (approved):

DomBlack (approved):

PR #217, opened 2022-05-17
Add encoding.DescribeResponse

ekerfelt:
This commit adds a new `DescribeResponse` function to the encoding package.
This function's job is to describe the http wire representation of a Response type

The output of this function can then be used for both the client generation
and for the code generation inside Encore request handlers for applications.

### DIFF
--- a/parser/encoding/rpc.go
+++ b/parser/encoding/rpc.go
@@ -20,14 +20,8 @@ const (
 	Body                            // Parameter is placed in the body
 )
 
-// tagToLocation is used to translate struct field tags to param locations
-// if overrideDefault is set, the default HTTP param location will be changed
-// if the tag matches the paramLocation, the param name will be override with the
-// tag name
-var tagToLocation = map[string]struct {
-	location        ParameterLocation
-	overrideDefault bool
-}{
+// requestTags is a description of tags used for requests
+var requestTags = map[string]tagDescription{
 	"query": {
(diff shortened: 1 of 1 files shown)


parser/encoding/rpc.go
```diff
@@ -69,43 +91,47 @@ func (e *RPCEncoding) RequestEncodingForMethod(method string) *RequestEncoding {
...
 
-// ParameterEncoding expresses how a parameter should be encoded
+// ParameterEncoding expresses how a parameter should be encoded on the wire
 type ParameterEncoding struct {
 	// Location (e.g. header, query, body) where the parameter should be marshalled from/to
 	Location ParameterLocation
 	// The location specific name of the parameter (e.g. cheeseEater, cheese-eater, X-Cheese-Eater
 	Name string
+	// Whether the name should be case-sensitive
+	CaseSensitive bool
```
DomBlack, line 115: When sent to `true` the standard library JSON (i.e. the one we use in the client generator) will still be case insensitive. Is this ok? (i.e. what's the use case fo having this be true?)

DomBlack (dismissed):
One small comment

parser/encoding/rpc.go
ekerfelt: Good point. I'll remove it

DomBlack (approved):

PR #218, opened 2022-05-17
Shared Go Code Generator for Marshalling

DomBlack:
This commit ports all the `typeDecoder` logic from compiler/internal/codegen/codegen_main.go
into a new shared package called `gocodegen` and adds a new set of methods for encoding to
strings for use in the Go client.

I've not removed or modified the existing compiler to use this new package yet. But that
will be a future commit once we've finished adding the header support to the compiler.

## Example Usage Generation
```go
type SvcAllInputTypes[A any] struct {
	A    []time.Time `header:"X-Alice"`         // Specify this comes from a header field
	B    int         `query:"Bob"`              // Specify this comes from a query string
	C    bool        `json:"Charile,omitempty"` // This can come from anywhere, but if it comes from the payload in JSON it must be called Charile
	Dave A           // This generic type complicates the whole thing 🙈
}

type SvcHeaderOnlyStruct struct {
	Foo []int `header:"X-Foo"`
	Bar bool  `header:"X-Bar"`
}

func (c *svcClient) GetRequestWithAllInputTypes(ctx context.Context, params SvcAllInputTypes[int]) (resp SvcHeaderOnlyStruct, err error) {
	// Convert our params into the objects we need for the request
	reqEncoder := &marshaller{}

	headers := map[string][]string{"X-Alice": reqEncoder.FromTimeList(params.A)}

	queryString := url.Values{
		"Bob":  []string{reqEncoder.FromInt(params.B)},
		"c":    []string{reqEncoder.FromBool(params.C)},
		"dave": []string{reqEncoder.FromInt(params.Dave)},
	}

	if reqEncoder.LastError != nil {
		err = fmt.Errorf("unable to marshal parameters: %w", reqEncoder.LastError)
		return
	}

	var respHeaders http.Header
	respHeaders, err = callAPI(ctx, c.base, "GET", fmt.Sprintf("/svc.GetRequestWithAllInputTypes?%s", queryString.Encode()), headers, nil, nil)
	if err != nil {
		return
	}

	// Copy the unmarshalled response body into our response struct
	respDecoder := &marshaller{}

	resp.Foo = respDecoder.ToIntList("Foo", respHeaders.Values("X-Foo"), false)
	resp.Bar = respDecoder.ToBool("Bar", respHeaders.Get("X-Bar"), false)
	if respDecoder.LastError != nil {
		err = fmt.Errorf("unable to unmarshal headers: %w", respDecoder.LastError)
		return
	}

	return
}
```

## Example Struct Generation
```go
// marshaller is used to marshal requests to strings and unmarshal responses from strings
type marshaller struct {
	LastError error // The last error that occurred
}

func (e *marshaller) FromInt(s int) (v string) {
	return strconv.FormatInt(int64(s), 10)
}

func (e *marshaller) FromTime(s time.Time) (v string) {
	return s.Format(time.RFC3339)
}

func (e *marshaller) FromTimeList(s []time.Time) (v []string) {
	for _, x := range s {
		v = append(v, e.FromTime(x))
	}
	return v
}

func (e *marshaller) FromBool(s bool) (v string) {
	return strconv.FormatBool(s)
}

func (e *marshaller) ToInt(field string, s string, required bool) (v int) {
	if !required && s == "" {
		return
	}
	x, err := strconv.ParseInt(s, 10, 64)
	e.setErr("invalid parameter", field, err)
	return int(x)
}

func (e *marshaller) ToIntList(field string, s []string, required bool) (v []int) {
	if !required && len(s) == 0 {
		return
	}
	for _, x := range s {
		v = append(v, e.ToInt(field, x, required))
	}
	return v
}

func (e *marshaller) ToBool(field string, s string, required bool) (v bool) {
	if !required && s == "" {
		return
	}
	v, err := strconv.ParseBool(s)
	e.setErr("invalid parameter", field, err)
	return v
}

func (e *marshaller) ToTime(field string, s string, required bool) (v time.Time) {
	if !required && s == "" {
		return
	}
	v, err := time.Parse(time.RFC3339, s)
	e.setErr("invalid parameter", field, err)
	return v
}

func (e *marshaller) ToTimeList(field string, s []string, required bool) (v []time.Time) {
	if !required && len(s) == 0 {
		return
	}
	for _, x := range s {
		v = append(v, e.ToTime(field, x, required))
	}
	return v
}

// setErr sets the last error within the object if one is not already set
func (e *marshaller) setErr(msg, field string, err error) {
	if err != nil && e.LastError == nil {
		e.LastError = fmt.Errorf("%s: %s: %w", field, msg, err)
	}
}
```

### DIFF
--- a/internal/gocodegen/marshalling.go
+++ b/internal/gocodegen/marshalling.go
@@ -1,0 +1,473 @@
+package gocodegen
+
+import (
+	"strings"
+
+	"github.com/cockroachdb/errors"
+	. "github.com/dave/jennifer/jen"
+
+	schema "encr.dev/proto/encore/parser/schema/v1"
+)
+
+const (
+	lastErrorField = "LastError"
+)
+
+// MarshallingCodeGenerator is used to generate a structure has methods for decoding various types, collecting the errors.
+// It will only generate methods required for the given types.
+type MarshallingCodeGenerator struct {
+	structName string
+	used       bool
+
+	builtins     []methodDescription
+	seenBuiltins map[methodKey]methodDescription
+}
+
+type methodKey struct {
+	fromString bool
+	builtin    schema.Builtin
+	slice      bool
+}
+
+type methodDescription struct {
+	FromString bool
+	Method     string
+	Input      Code
+	Result     Code
+	IsList     bool
+	Block      []Code
+}
+
+// MarshallingCodeWrapper is returned by NewPossibleInstance and tracks usage within a block
+type MarshallingCodeWrapper struct {
+	g            *MarshallingCodeGenerator
+	instanceName string
+	used         bool
+
+	code []Code
+}
+
+func NewMarshallingCodeGenerator(structName string) *MarshallingCodeGenerator {
+	return &MarshallingCodeGenerator{
+		structName:   structName,
+		builtins:     nil,
+		seenBuiltins: make(map[methodKey]methodDescription),
+	}
+}
+
+// NewPossibleInstance Creates a statement to initialise a new encoding instance.
+//
+// Use the returned wrapper to convert FromStrings to the target types, adding any code you
+// are generating to the wrapper using Add. Once you've finished generating all the code which
+// may need type conversion with that _instance_ of the deserializer, call Finalize to generate the code full code
+// including error handling.
+//
+// Once you've finished writing the whole app with all the code which uses this generator call WriteToFile to write
+// the supporting struct and methods to the given file
+func (g *MarshallingCodeGenerator) NewPossibleInstance(instanceName string) *MarshallingCodeWrapper {
+	g.used = true
+	return &MarshallingCodeWrapper{g: g, instanceName: instanceName}
+}
+
+// WriteToFile writes the full encoder type into the given file.
+func (g *MarshallingCodeGenerator) WriteToFile(f *File) {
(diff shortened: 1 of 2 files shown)


internal/gocodegen/marshalling.go
```diff
@@ -0,0 +1,473 @@
...
+				Id("field"),
+				Id("msg"),
+				Id("err"),
+			),
+		),
+	)
+	f.Line()
+}
+
+func (b *MarshallingCodeGenerator) builtinFromString(t schema.Builtin, slice bool) (string, error) {
```
DomBlack, line 123: This is a straight copy/paste of `builtinDecoder` from codegen_main.go

internal/gocodegen/marshalling.go
```diff
@@ -0,0 +1,473 @@
...
+			return "", err
+		}
+	}
+
+	b.seenBuiltins[key] = fn
+	b.builtins = append(b.builtins, fn)
+	return fn.Method, nil
+}
+
+func (b *MarshallingCodeGenerator) builtinToString(t schema.Builtin, slice bool) (string, error) {
```
DomBlack, line 253: This method is brand new and does the reverse of the above - but I based it by copy/pasting the above method and then tweaking

ekerfelt (approved):

PR #219, opened 2022-05-18
Add Header support to Go Client Generator

DomBlack:
This commit adds header support to the Go Client Generator using the new
DescribeRPC and shared Type Marshaling generator.

The golden package is now being used for the Client Generator to make it easier
to update the expected test files.

It updates the input to the client generator test to include a bunch of new
cases to try and expose as many test edges as possible.

### DIFF
--- a/cli/internal/codegen/golang.go
+++ b/cli/internal/codegen/golang.go
@@ -8,18 +8,25 @@ import (
 
 	"github.com/cockroachdb/errors"
 	. "github.com/dave/jennifer/jen"
+	"github.com/fatih/structtag"
 
 	"encr.dev/cli/internal/version"
+	"encr.dev/internal/gocodegen"
+	"encr.dev/parser/encoding"
 	meta "encr.dev/proto/encore/parser/meta/v1"
 	schema "encr.dev/proto/encore/parser/schema/v1"
 )
 
 type golang struct {
-	md *meta.Data
+	md  *meta.Data
+	enc *gocodegen.MarshallingCodeGenerator
 }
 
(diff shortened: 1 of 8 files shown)


cli/internal/codegen/golang.go
```diff
@@ -438,8 +416,222 @@ func (g *golang) rpcCallSite(rpc *meta.RPC) (code []Code) {
...
+			code = append(code, Comment("Convert our params into the objects we need for the request"))
+		}
+
+		enc := g.enc.NewPossibleInstance("reqEncoder")
+
+		// Generate the headers
+		if len(headerFields) > 0 {
+			values := Dict{}
+
+			// Check the request schema for fields we can put in the query string
```
ekerfelt, line 444:
```suggestion
			// Check the request schema for fields we can put in the header
```

cli/internal/codegen/golang.go
```diff
@@ -8,18 +8,25 @@ import (
...
-	md *meta.Data
+	md  *meta.Data
+	enc *gocodegen.MarshallingCodeGenerator
 }
 
 func (g *golang) Generate(buf *bytes.Buffer, appSlug string, md *meta.Data) (err error) {
 	g.md = md
+	// note we use unicode characters to write "encore" to ensure we don't conflict with any existing names in the
+	// package the generated client is placed.
+	g.enc = gocodegen.NewMarshallingCodeGenerator("ℯ𝓃𝑐ℴ𝑟ℯMarshaller")
```
ekerfelt, line 29: Nit: I'm not a big fan of unicode chars in variable names, but I guess it's just a matter of preferences :)

ekerfelt (approved):

PR #116, opened 2021-12-18
cli: Improve TypeScript codegen

ValeriaVG:
- Allows Client methods overload
- Provides support for Raw endpoints

**Breaking changes:**
- Generated TypeScript client methods do not throw an error anymore and return `Result` instead, which either contains `{error: ErrorResponse}` or `data` field with the json from the response body
- Raw endpoint methods return underlying `Response`, without consuming response body
- Void endpoints no longer consume response body, unless they result in an error

Tested on JS & Deno; in order to support NodeJS one needs to either import fetch polyfill or extend `Client` class and override `doRaw` method.

Closes #114,#50

Issue #114, opened 2021-12-12
TypeScript Client improvements


ValeriaVG:
Client generation functionality is awesome, but I think there are some things that would make it better:

1. Add the ability to change token without re-creating a client instance. A simple solution would be to expose setter/getter for the said token instead of providing token as a constructor argument
2. Check that token is truthy before adding auth header. I've accidentally set it to `null` and requests to not-auth paths started to fail with 401
3. Return error response as parsed json instead of a stringified error. One solution could be to override the fetch implementation altogether, e.g. by providing it as one of the constructor options

Let me know what makes sense and I can implement it.

eandre:
All of these ideas make a lot of sense and we'd be happy to incorporate all of them! There's also #50 with another set of improvements that we should add. Perhaps it's time to create a proper npm package to hold things like the error type you described and so on?

ValeriaVG:
Given that the common types are quite small I'm not sure if it's worth enforcing a separate dependency. 
I'd imagine that a clean zero-dependency TypeScript would work without problems for node, deno and browsers (through a bundler).

I'd remove BaseClient completely and instead would pass Client's `exec` or `call` method to the service clients. This way the behaviour can be customized for almost any scenario by simple extending the Client class if needed and providing decent defaults otherwise.
 
 What do you think?

eandre:
That's true, I agree with you. If you have a vision for how you think this should look, go for that! I trust your judgement :)

Issue #50, opened 2021-08-05
Handling bodies for raw endpoints in the dashboard and TypeScript client 


vilhelmmelkstam:
This issue covers the response and request bodies of raw endpoint in the dashboard and in the TypeScript client.

### Dashboard
Currently, the even though Raw Endpoints may have request and response bodies, there are no way of editing the request body or viewing the response body. I understand that encore can't know if there will be a body and the format of it, but we should at least be able to edit it as if it were text. The same goes for viewing the response, all non-raw endpoints show their response, and the raw endpoints should probably show their reponse as well, probably just in plain-text.

 ### TypeScript Client
 A raw endpoint is currently generated into a function that looks like the following.
 ```typescript
public World(): Promise<void> {
    return this.baseClient.doVoid("POST", `/hello.World`)
}
```
However, this does not allow for neither a response nor request body. Therefore, I think it would be nice if the client would in some way handle bodies. Example solutions are:
1. Make the function take in a `ReadableStream` (The type of the fetch request body) and return the `Request` object.
2. Make the function take a `ReadableStream` and return it as a string. If a user wants it in JSON, they can parse it.
3. Assume the request and response body to still be JSON and use generics to specify the request and response schema. This would still allow for strings to be sent as they are JSON-strinfigy-able.

eandre:
Thanks for writing this up. I agree completely. I wonder if it would make sense to add a parameter that is effectively the `fetch` options object (https://developer.mozilla.org/en-US/docs/Web/API/Request/Request -> `init`), and return the `Response`? That would allow callers to customize any behavior that the underlying `fetch` API supports.

However at some point we should also start thinking about making the generated code support both client-side and server-side JavaScript (Node/Deno). I think the easiest approach for doing this would be to create an npm package that provides the base client (`@encoredev/client` or something). We don't need to do that as part of this change (we're already assuming the existence of `fetch` in the global scope), but something to think about.

vilhelmmelkstam:
That sounds reasonable. This would also allow developers to set other options (such as `credentials: include`, something that we currently add manually).

As for server-side fetching, I don't think it's a major problem as adding a polyfill such as [unfetch](https://github.com/developit/unfetch) is an easy fix. It's not the most elegant solution, but at least quick.

DomBlack:
@vilhelmmelkstam this has now been resolved in #223 and will be included with the next release of the Encore CLI tool.

PR #223, opened 2022-05-20
Update client generators for Header support.

DomBlack:
This commit updates both the Go and Typescript client generators
to support the upcoming header support.

This also includes a large scale refactoring of the Typescript client (in a backwards
compatible way) to address several open issues which the Go client already supports:
 - Ability to change token without re-creating the client via the `TokenGenerator` option (#114)
 - Raw endpoints are now fully supported and take in a `RequestInit` object and return the full Response object (#50)
 - Allow clients to override settings of the request by passing in a custom `fetch` implementation via the options (#116)


cli/internal/codegen/golang.go
```diff
@@ -17,16 +17,33 @@ import (
...
+// old code.
+type goGenVersion int
+
+const (
+	// GoInitial is the originally released typescript generator
+	GoInitial goGenVersion = iota
+
+	// GoExperimental can be used to lock experimental or uncompleted features in the generated code
+	// It should always be the last item in the enum
+	GoExperimental
```
eandre, line 31: Nit: these comments still refer to `typescript`

cli/internal/codegen/golang.go
```diff
@@ -17,16 +17,33 @@ import (
...
+	enc              *gocodegen.MarshallingCodeGenerator
+	generatorVersion goGenVersion
 }
 
 func (g *golang) Generate(buf *bytes.Buffer, appSlug string, md *meta.Data) (err error) {
 	g.md = md
 	// note we use unicode characters to write "encore" to ensure we don't conflict with any existing names in the
 	// package the generated client is placed.
-	g.enc = gocodegen.NewMarshallingCodeGenerator("ℯ𝓃𝑐ℴ𝑟ℯMarshaller")
+	g.enc = gocodegen.NewMarshallingCodeGenerator("ℯ𝓃𝑐ℴ𝑟ℯMarshaller", true)
```
eandre, line 46: How is the unicode name used? If it's in the generated code I think we should maybe go for something a bit more standard, even though it's technically allowed by the Go spec it might cause people to trip up. Maybe "encoreClientMarshaller" or something?

cli/internal/codegen/testdata/expected_golang.go
```diff
@@ -84,11 +89,87 @@ func WithAuthFunc(tokenGenerator func(ctx context.Context) (string, error)) Opti
...
+
+type productsClient struct {
+	base *baseClient
+}
+
+var _ ProductsClient = (*productsClient)(nil)
+
+func (c *productsClient) Create(ctx context.Context, params ProductsCreateProductRequest) (resp ProductsProduct, err error) {
+	// Convert our params into the objects we need for the request
+	headers := map[string][]string{"Idempotency-Key": {params.IdempotencyKey}}
```
eandre, line 138: Should we maybe use [http.Header](https://pkg.go.dev/net/http#Header) here? Would look a bit more standard.

cli/internal/codegen/golang.go
DomBlack: @ekerfelt also commented on this in the original PR. The reasoning was simply to avoid naming complications by putting an encore prefix on it.

Happy to change to just `marshaller` or `serde` if you'd prefer (and ignore the name conflict (as I think it's a minor risk)

cli/internal/codegen/testdata/expected_golang.go
DomBlack: I went with a basic map over a `http.Headers` because via the headers API you can't get a list of the keys - but you can cast it to a map. I guess I was thinking if the headers type gets changed in the future while keeping the API consistent we'd need to make a breaking change to swap to a type where we could read all the keys.

That said I'll change it, because again I think this is a minimal risk

eandre: `http.Headers` is exactly the same data structure under the hood (`map[string][]string`) so I figured it would behave identically?

DomBlack: It is, but it's a named type with it's own API. So I think it's safe to change - but my fear was what if they changed the named type to a struct with the map in it for some reason.

As I said, changing the code to reference `http.Header` 😄

cli/internal/codegen/golang.go
eandre: Yeah I think it might trip people up; maybe something longer and more specific would minimize the risk still. I guess we probably expect people to put it in a separate package regardless.

cli/internal/codegen/testdata/expected_golang.go
eandre: Ah I understand, yeah they cannot change that since iterating over the map is part of the public API.

eandre (approved):
Nice!

PR #224, opened 2022-05-20
Add structured errors to Typescript Client

### LINKED ISSUE #114: TypeScript Client improvements

DomBlack:
This commit adds structured errors to the typescript client in the
form of an `APIError` object which will be thrown incase of an
error. This makes a best attempt at recovering the structured error
as reported by an Encore API and parsing it to be easily usable from
withing a client.

This also causes the client to preemptively return throw an error if
no bearer token is present when trying to call an Authenticated API.

Fixes #114

```typescript
const client = new Client(Local)

try {
    const productListing = await client.products.List()
} catch (e: Error) {
    if (isAPIError(e)) {
        console.log(`Got an ${e.code} error with message ${e.message} and details ${e.details}`)
    } else {
       console.log(`Got error ${e.message}`)
    }
}
```


cli/internal/codegen/testdata/expected_typescript.ts
```diff
@@ -385,34 +399,324 @@ class BaseClient {
...
+        typeof(err.code) === "number" &&
+        typeof(err.message) === "string" &&
+        (err.details === undefined || err.details === null || typeof(err.details) === "object")
+    )
+}
+
+/**
+ * EncoreError represents a structured error as returned from an Encore application.
+ */
+export class EncoreError extends Error {
```
eandre, line 466: Maybe this should be `APIError`? `EncoreError` might suggest it's an error **with** Encore or **from Encore the product/platform**, as opposed to from the *Encore application*.

cli/internal/codegen/testdata/expected_typescript.ts
```diff
@@ -375,7 +389,7 @@ class BaseClient {
     }
 
     // callAPI is used by each generated API method to actually make the request
-    public async callAPI(method: string, path: string, body?: BodyInit, params?: CallParameters): Promise<Response> {
+    public async callAPI(method: string, path: string, requiresAuth: boolean, body?: BodyInit, params?: CallParameters): Promise<Response> {
```
eandre, line 392: Should this be public? If we're changing the signature it suggests it's not stable and should therefore not be public?

cli/internal/codegen/testdata/expected_typescript.ts
```diff
@@ -385,34 +399,324 @@ class BaseClient {
...
+            try {
+                let jsonBody = await response.json()
+                if (isAPIErrorResponse(jsonBody)) {
+                    body = jsonBody
+                } else {
+                    body.message += ": " + JSON.stringify(jsonBody)
+                }
+            } catch {
+                // otherwise we just append the text to the error message
+                body.message += ": " + await response.text()
```
eandre, line 435: If `response.json()` fails, I don't believe we can call `response.text()` as the body has already been consumed. Maybe better to call `response.text()` and then `JSON.parse()` on that?

cli/internal/codegen/testdata/expected_typescript.ts
DomBlack: The base client is private and can’t be accessed access of this file

ekerfelt (approved):

PR #225, opened 2022-05-23
Add structured errors to Go Client

DomBlack:
This commit adds the ability for a generated Go Client to return
the structured errors that are generated server side.


cli/internal/codegen/testdata/expected_golang.go
```diff
@@ -477,7 +478,19 @@ func callAPI(ctx context.Context, client *baseClient, method, path string, heade
 		_ = rawResponse.Body.Close()
 	}()
 	if rawResponse.StatusCode >= 400 {
-		return nil, fmt.Errorf("got error response: %s", rawResponse.Status)
+		// Read the full body sent back
+		body, err := ioutil.ReadAll(rawResponse.Body)
+		if err != nil {
+			return nil, fmt.Errorf("got error response with readable body: %s", rawResponse.Status)
```
ekerfelt, line 484:
```suggestion
			return nil, fmt.Errorf("got error response without readable body: %s", rawResponse.Status)
```

cli/internal/codegen/testdata/expected_golang.go
```diff
@@ -477,7 +478,19 @@ func callAPI(ctx context.Context, client *baseClient, method, path string, heade
...
+		}
+
+		// Attempt to decode the error response as a structured APIError
+		apiError := &APIError{}
+		if err := json.Unmarshal(body, apiError); err != nil {
+			// If the error is not a parsable as an APIError, then return an error with the raw body
+			return nil, fmt.Errorf("got error response: %s", string(body))
```
ekerfelt, line 491: AFAIK, the TS client always returns an APIError, even when the response isn't one. Should we here too for consistency?

ekerfelt (commented):
Two nits!

cli/internal/codegen/testdata/expected_golang.go
DomBlack: Fair, changed

ekerfelt (approved):

PR #230, opened 2022-05-23
parser/encoding/rpc: refactor encoding model

ekerfelt:
Refactor the encoding model in preparation for api explorer changes


cli/internal/codegen/testdata/expected_golang.go
```diff
@@ -371,10 +371,10 @@ func (c *svcClient) RequestWithAllInputTypes(ctx context.Context, params SvcAllI
 	// Copy the unmarshalled response body into our response struct
 	respDecoder := &serde{}
 
-	resp.A = respDecoder.ToTime("A", respHeaders.Get("x-alice"), false)
 	resp.B = respBody.B
 	resp.C = respBody.C
 	resp.Dave = respBody.Dave
+	resp.A = respDecoder.ToTime("A", respHeaders.Get("x-alice"), false)
```
eandre, line 377: Why the ordering change? Is it due to the refactoring or is the output indeterministic?

eandre (commented):
Looks great to me but leaving for @DomBlack to approve. Just one minor question

cli/internal/codegen/testdata/expected_golang.go
ekerfelt: It's due to the refactoring. I can change it back to make it consistent

parser/encoding/rpc.go
```diff
@@ -13,6 +13,14 @@ import (
...
 )
 
+type Lang int
+
+const (
+	_ Lang = iota
+	GO
+	TypeScript
+)
+
```
DomBlack, line 23: I'm not sure I agree with moving language-specific logic in here; I'd rather we kept `SrcField` as it appears in the Go code, and then let any language generator make any changes it deems are needed.

For instance, if we had a Swift client, that will have a different naming convention to TypeScript and Go - does it mean that the `encoding/rpc.go` file now also needs to understand that as well?

Instead if we keep it `SrcField` as the Go struct field name; then the various client generators can update as needed

cli/internal/codegen/typescript.go
```diff
@@ -356,16 +353,10 @@ func (ts *typescript) rpcCallSite(ns string, w *indentWriter, rpc *meta.RPC, rpc
 
 	w.WriteStringf("// Now make the actual call to the API\nconst resp = await %s\n", callAPI)
 
-	headerFields, queryFields, _, err := toFieldLists(rpcEncoding.ResponseEncoding.Fields)
-	if err != nil {
-		return err
-	}
-	if len(queryFields) > 0 {
```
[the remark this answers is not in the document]
ekerfelt: Fair, we should move it into the DescribeRespose method

parser/encoding/rpc.go
ekerfelt: Agreed with Dom on a solution

DomBlack (dismissed):

DomBlack (approved):

PR #233, opened 2022-05-24
parser: support named struct types as auth handler params

eandre:


parser/svc.go
```diff
@@ -442,15 +444,20 @@ func (p *parser) validateAuthHandler(h *est.AuthHandler) {
...
-	// Second param must be string
-	tok, _ := getField(params, 1)
-	if id, ok := tok.Type.(*ast.Ident); !ok || id.Name != "string" {
-		p.err(tok.Type.Pos(), "second parameter must be of type string"+sigHint)
-		return
-	} else if pkgNames.Decls["string"] != nil {
-		p.err(tok.Type.Pos(), "second parameter must be of type string (local name shadows builtin)"+sigHint)
-		return
+	// Second param must be string or named type pointing to a struct
+	authInfo, _ := getField(params, 1)
```
ekerfelt, line 448: Maybe we should require a pointer to a struct?

eandre: Haha yes, except we don't yet support pointers :)

DomBlack (approved):

parser/svc.go
DomBlack: 🧂

ekerfelt: But this is the ast! You can just check that the authInfo.Type is an ast.StarExpr (if not string)

eandre: Oh true :) I wasn't sure if you were trolling!

ekerfelt: Me, troll, on a Tuesday?

PR #236, opened 2022-05-24
Add 'encore eject' command

eandre:
- cli: add 'encore export docker' command
- cli: rename export -> eject
- runtime: disable tracing when ejected
- cli/daemon/export: simplify keychain setup


cli/cmd/encore/eject.go
```diff
@@ -0,0 +1,86 @@
...
+
+	daemon := setupDaemon(ctx)
+	params := &daemonpb.DockerExportParams{}
+	if push {
+		params.PushDestinationTag = imageTag
+	} else {
+		params.LocalDaemonTag = imageTag
+	}
+
+	goos, goarch := "linux", runtime.GOARCH
```
DomBlack, line 57: Should we support a buildx style docker image which has native support for a matrix of targets?

We do this for emissary and the placeholder service so they can be deployed onto either ARM64 or AMD64 systems: https://github.com/encoredev/placeholder-service/blob/main/.github/workflows/cd.yml#L64-L73

cli/daemon/export/export.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "add file to tar")
+		}
+		if _, err := io.Copy(tw, appFile); err != nil {
+			return nil, errors.Wrap(err, "copy file to tar")
+		}
+	}
+
+	// Download ca certs
+	const certsDest = "/etc/ssl/certs/ca-certificates.crt" // from https://go.dev/src/crypto/x509/root_linux.go
```
DomBlack, line 205: This comment says it's from `go.dev` - but then downloads them Mozilla - I think this comment is a little redundant - as that's the standard ca cert location

DomBlack (approved):
Small question; but otherwise it's good

cli/daemon/export/export.go
eandre: The reason we chose that destination folder is sourced from that link; did you confuse it with where we download the certs from?

cli/cmd/encore/eject.go
DomBlack: I thought the Go compiler can target any GOOS / GOARCH from any other GOOS / GOARCH version - so the encore-go installed should be able to compile both versions without issue. (Unless I'm missing something)

cli/daemon/export/export.go
DomBlack: At first yes, then I realised what you meant in this comment - but I thought it was redundant and potentially confusing - as `/etc/ssl/certs/ca-certificates.crt` isn't Go specific but more general for linux

DomBlack (approved):

PR #239, opened 2022-05-25
Generated Clients: Query Strings in Raw Endpoints

DomBlack:
This commit adds support for query strings fields to be used in
requests made to raw endpoints.

### DIFF
--- a/cli/daemon/run/testdata/echo_client/client.ts
+++ b/cli/daemon/run/testdata/echo_client/client.ts
@@ -231,12 +231,12 @@ export namespace echo {
          */
         public async MuteEcho(params: Data<string, string>): Promise<void> {
             // Convert our params into the objects we need for the request
-            const query: Record<string, string | string[]> = {
+            const queryFields: Record<string, string | string[]> = {
                 key:   params.Key,
                 value: params.Value,
             }
 
-            await this.baseClient.callAPI("GET", `/echo.MuteEcho?${encodeQuery(query)}`, false)
+            await this.baseClient.callAPI("GET", `/echo.MuteEcho`, false, undefined, {queryFields})
         }
 
         /**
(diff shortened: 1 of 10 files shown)


cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -516,7 +516,13 @@ function encodeQuery(parts: Record<string, string | string[]>): string {
     return pairs.join("&")
 }
 // CallParameters is the type of the parameters to a method call, but require headers to be a Record type
-type CallParameters = Omit<RequestInit, "method" | "body"> & { headers?: Record<string, string> }
+type CallParameters = Omit<RequestInit, "method" | "body"> & {
+    /** Any headers to be sent with the request */
+    headers?: Record<string, string>;
+
+    /** Any query parameters to be sent with the request */
+    queryFields?: Record<string, string | string[]>
```
eandre, line 524: queryFields is quite verbose. Go calls it query, and it aligns with our struct tag.

DomBlack: > Go calls it query,

However `RawQuery` (which is only 3 characters shorter 😉) in the URL object is the string of the entire query string, whereas I wanted to be clear that this was not a query string, but rather the fields which make up the query string.

Verboseness is good when it lowers the cognitive overhead needed to understand code at a glance.

> and it aligns with our struct tag.

That's fair, I'm still somewhat in favour of the verboseness to remove confusion maybe for frontend engineers who don't work with Encore directly.

What do you think?

eandre: I think given that the type is a `Record` and we should document it, I think `query` is clear enough. If I had suggested `qry` I would be inclined to agree that it’s too terse :)

eandre (approved):

PR #246, opened 2022-05-26
New Auth Support for Generated Clients

DomBlack:
This commit adds support for the new style authentication
handlers to the generated clents and allows us to generate
clients which _either_ support Bearer tokens or a complex `Params`
object which can contain Header and Query String fields.


cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -26,32 +26,14 @@ export default class Client {
     public readonly test: test.ServiceClient
 
 
-    /**
-     * @deprecated This constructor is deprecated, and you should move to using BaseURL with an Options object
-     */
-    constructor(target?: string, token?: string)
-
```
eandre, line 33: Do we no longer support the old-style constructor?

DomBlack: It's still added - but only when token based authentication is in use

eandre (approved):

PR #252, opened 2022-05-30
Case Insensitive Language on Client Generation

DomBlack:
This commit modifies the client generation to allow case insensitive
comparisions for the language when generating a client.

### DIFF
--- a/cli/internal/codegen/client.go
+++ b/cli/internal/codegen/client.go
@@ -52,7 +52,7 @@ func Client(lang Lang, appSlug string, md *meta.Data) (code []byte, err error) {
 	}()
 
 	var gen generator
-	switch lang {
+	switch Lang(strings.ToLower(string(lang))) {
 	case LangTypeScript:
 		gen = &typescript{generatorVersion: typescriptGenLatestVersion}
 	case LangGo:


cli/internal/codegen/client.go
```diff
@@ -52,7 +52,7 @@ func Client(lang Lang, appSlug string, md *meta.Data) (code []byte, err error) {
 	}()
 
 	var gen generator
-	switch lang {
```
eandre, line 55: Since `lang` is effectively an enumeration it’s unexpected to change its value like this. Can we move this logic to the place where the value is parsed instead?

DomBlack: The issue is where we parse it, we simply convert to the `Lang` type from what ever raw input was given. This is the first time we do any validation on the data - hence why I kept it with the validation logic.

Happy to move, but then I think we should fully validate the enum there, rather than just convert the data type

eandre: Yeah, I think we should validate it where we do the conversion so we don't end up with `Lang` values that are invalid in the first place

eandre (approved):
Nice!

PR #253, opened 2022-05-30
all: keep track of known apps persistently 

eandre:
To support dependency injection Encore needs to generate code that is
written into the app's file tree so that editors can pick it up.

It would be frustrating if the code generation was only enabled after
running 'encore run', but the daemon had no way of knowing the location
of Encore apps on the local filesystem.

Fix this by introducing a local sqlite db that tracks known Encore apps
on the local filesystem. We can then use this to watch the filesystem for
changes to the known Encore apps and trigger code generation even without
having to run 'encore run'.

### DIFF
--- a/cli/cmd/encore/daemon/schema.sql
+++ b/cli/cmd/encore/daemon/schema.sql
@@ -1,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS schema_migrations (
+    dummy BOOLEAN PRIMARY KEY,
+    version INT NOT NULL
+);
+
+INSERT OR REPLACE INTO schema_migrations(dummy, version) VALUES (true, 1);
+
+CREATE TABLE IF NOT EXISTS app (
+    root TEXT PRIMARY KEY,
+    local_id TEXT NOT NULL,
+    platform_id TEXT NULL, -- NULL if not linked
+    updated_at TEXT NOT NULL
+);
--- a/cli/daemon/apps/apps.go
+++ b/cli/daemon/apps/apps.go
@@ -1,0 +1,140 @@
+package apps
+
+import (
+	"database/sql"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/rs/zerolog/log"
+
+	"encr.dev/cli/daemon/internal/manifest"
+	"encr.dev/cli/internal/appfile"
+)
+
+var ErrNotFound = errors.New("app not found")
+
+// Instance describes an app instance known by the Encore daemon.
+type Instance struct {
+	root       string
+	localID    string
+	platformID string
+}
+
+func NewInstance(root, localID, platformID string) *Instance {
+	return &Instance{root, localID, platformID}
+}
+
+// Root returns the filesystem path for the app root.
+// It always returns a non-empty string.
+func (i *Instance) Root() string { return i.root }
+
+// LocalID reports a local, random id unique for this app,
+// as persisted in the .encore/manifest.json file.
+// It always returns a non-empty string.
+func (i *Instance) LocalID() string { return i.localID }
+
+// PlatformID reports the Encore Platform's ID for this app.
+// If the app is not linked it reports the empty string.
+func (i *Instance) PlatformID() string { return i.platformID }
+
+// PlatformOrLocalID reports PlatformID() if set and otherwise LocalID().
+func (i *Instance) PlatformOrLocalID() string {
+	if i.platformID != "" {
+		return i.platformID
+	}
+	return i.localID
+}
+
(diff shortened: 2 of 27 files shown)


cli/daemon/apps/apps.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Track begins tracking an app, and marks it as updated
+// if the app is already tracked.
+func (mgr *Manager) Track(appRoot string) (*Instance, error) {
+	app, err := mgr.resolve(appRoot)
+	if err != nil {
+		return nil, err
+	}
+	_, err = mgr.db.Exec(`
```
DomBlack, line 70: What if there are multiple checkout's of the same application?

cli/cmd/encore/daemon/schema.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS app (
+    root TEXT PRIMARY KEY,
+    local_id TEXT NOT NULL,
+    platform_id TEXT NULL, -- NULL if not linked
+    updated_at TEXT NOT NULL
+);
```
DomBlack, line 6: Can we please also add a schema version table, just so when we do need to make a breaking change to the schema, it's easier for us to write the migration code

cli/daemon/apps/apps.go
```diff
@@ -0,0 +1,140 @@
...
+func (i *Instance) Root() string { return i.root }
+
+// LocalID reports a local, random id unique for this app,
+// as persisted in the .encore/manifest.json file.
+// It always returns a non-empty string.
+func (i *Instance) LocalID() string { return i.localID }
+
+// PlatformID reports the Encore Platform's ID for this app.
+// If the app is not linked it reports the empty string.
+func (i *Instance) PlatformID() string { return i.platformID }
```
DomBlack, line 41: What do you think about having a `type AppID string` type? I'm wondering if it might be nice to add a port of the EID package here down the line - but for now strongly typing the ID to a string at least gives us a easy migration path down the line to make that change.

cli/daemon/apps/apps.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return i.localID
+}
+
+func NewManager(db *sql.DB) *Manager {
+	return &Manager{db: db, instances: make(map[string]*Instance)}
+}
+
+// Manager keeps track of known apps and watches them for changes.
+type Manager struct {
```
DomBlack, line 56: Could we add a list function to list all App's in the database?

I'm thinking the new autocomplete could then merge both the locally known instances with the platform's instances.

eandre: Yeah that's a good idea, but I think we should hold off until we need it

cli/daemon/apps/apps.go
eandre: I started with that actually, but we ended up needing `PlatformOrLocalID()` in a lot of places because we have to handle the case where the app is not linked. It felt wrong to consider the LocalID as returning an AppID because it's not really the same thing. Not sure what's best

cli/cmd/encore/daemon/schema.sql
eandre: Good idea

cli/daemon/apps/apps.go
eandre: The primary key is the `appRoot` so they'll be stored multiple times. I went back and forth on this, not sure what's best exactly, but this approach works well for our current use case

cli/daemon/apps/apps.go
DomBlack: Well I already merged the autocomplete code, so I'd argue we kinda need it now ;)  - But fair can be a future PR

eandre: Well what would it return beyond what you already have? The "LocalID" is not meaningful to users, it's just a randomly generated string, so not sure it should be an autocomplete option

cli/daemon/apps/apps.go
DomBlack: One of the reasones we picked XID was to allow ID's to be generated in a distributed manor without communication - so I think LocalID's and AppID's could be interchangeable - I guess the question is if we need different ID prefixes or not (I'd say yes maybe).

Not sure, hence why I wondered if just having a strong type here is good enough.

Maybe if we want to strongly be able to tell the difference, we should use a stronger `AppOrLocaID` type which then can act basically like concrete `Either[A, B]` type?

cli/daemon/apps/apps.go
DomBlack: The slug for be able to generate a client for a running app using `--env=local` when it might not be linked

DomBlack (approved):

DomBlack (approved):

PR #254, opened 2022-05-30
compiler/internal/codegen: Handle nil pointer responses

ekerfelt:
This PR adds logic to handle nil pointers when serializing the response in the generated request handlers.


compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
```diff
@@ -335,6 +336,103 @@ func __encore_svc_Four(w http.ResponseWriter, req *http.Request, ps httprouter.P
...
+	resp, respErr := svc.Nine(req.Context(), p0, p1)
+	if respErr != nil {
+		respErr = errs.Convert(respErr)
+		runtime.FinishRequest(nil, respErr)
+		errs.HTTPError(w, respErr)
+		return
+	}
+
+	// Serialize the response
+	respData := []byte("null")
```
eandre, line 402: Do we want this to be `null\n`?

ekerfelt: yup!

PR #259, opened 2022-06-02
compiler/internal/codegen: Skip auth validation for empty auth params

ekerfelt:
This PR implements changes to only call the encore authhandler if at least one auth field is set. If an auth field is set, the authhandler will be called for *all* endpoints (including public ones). Unauthenticated errors will be ignored for all but 'auth' endpoints, whereas other errors (such as parsing error, other authhandler errors) will be returned to the caller.

### DIFF
--- a/cli/daemon/run/testdata/echo/echo/echo.go
+++ b/cli/daemon/run/testdata/echo/echo/echo.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"log"
 	"os"
+	"reflect"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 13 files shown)


cli/daemon/run/testdata/echo/echo/echo.go
```diff
@@ -173,6 +174,9 @@ type AuthParams struct {
 
 //encore:authhandler
 func AuthHandler(ctx context.Context, params *AuthParams) (auth.UID, *AuthParams, error) {
+	if reflect.ValueOf(params).IsZero() {
```
eandre, line 177: Pretty sure this just checks that it’s not a nil pointer. You need `.Elem().IsZero()`

ekerfelt: of course!

eandre (approved):

PR #260, opened 2022-06-07
TypeScript Client: Allow `AuthDataGenerator` to return undefined

melkstam:
This allows the `AuthDataGenerator` function that is an optional value to the encore TS client to allow the return type `undefined`. As field itself allows for undefined, I see no reason why the `AuthDataGenerator` shouldn't be able to return undefined.

The current workaround is to return an empty string as it is also falsey. However, it's not very elegant.

### DIFF
--- a/cli/daemon/run/testdata/echo_client/client.ts
+++ b/cli/daemon/run/testdata/echo_client/client.ts
@@ -582,7 +582,7 @@ type CallParameters = Omit<RequestInit, "method" | "body"> & {
 }
 
 // AuthDataGenerator is a function that returns a new instance of the authentication data required by this API
-export type AuthDataGenerator = () => echo.AuthParams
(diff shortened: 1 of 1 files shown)


vilhelmmelkstam:
Our use case:
We currently fetch the token from a cookie, which may or may not be there. We currently use the workaround, but using `''` feels like it would be undefined behaviour (even though it works).

```typescript
const getAuth: AuthDataGenerator = () => {
  const token = Cookie.get('jwt');
  return token ?? '';
};
```

to
```typescript
const getAuth: AuthDataGenerator = () => {
  return Cookie.get('jwt');
};
```

eandre:
Thanks, that makes sense. @DomBlack can you review this?

DomBlack:
This looks great; thanks @vilhelmmelkstam for the contribution.

PR #273, opened 2022-06-14
Pubsub MVP

ekerfelt:


runtime/pubsub/local.go
```diff
@@ -0,0 +1,112 @@
+//go:build !gcp
```
DomBlack, line 1: This could be 
```suggestion
//go:build encore_local
```

runtime/pubsub/local.go
```diff
@@ -0,0 +1,163 @@
...
+		panic(fmt.Sprintf("Error: %v", err))
+	}
+	// drop log messages for now since we cannot format them in the common daemon format
+	consumer.SetLoggerLevel(nsq.LogLevelMax)
+	// create a dedicated handler which forwards messages to the encore subscription
+	consumer.AddHandler(&handler[T]{sub: sub, cfg: cfg})
+	// connect the consumer to the NSQD
+	err = consumer.ConnectToNSQD(l.addr)
+	if err != nil {
+		panic(fmt.Sprintf("Error: %v", err))
```
DomBlack, line 100: Should we put a better error in this panic?

```suggestion
		panic(fmt.Sprintf("Unable to setup subscription %s for topic %s: %v", name, l.name, err))
```

runtime/pubsub/local.go
```diff
@@ -0,0 +1,163 @@
...
+		return "", err
+	}
+	// create and publish the message wrapper
+	data, err := json.Marshal(&message[T]{ID: idx, Data: msg, Attributes: attrs})
+	if err != nil {
+		return "", err
+	}
+	err = l.producer.Publish(l.name, data)
+	if err != nil {
+		return "", err
```
DomBlack, line 133: We should return an `encore.dev/errs` here which wraps the NSQ error

runtime/pubsub/local.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		l.producer = producer
+	}
+	// generate a new message ID
+	idx := fmt.Sprint(atomic.AddUint32(&l.idSeq, 1))
+
+	// turn the attributes into an attribute map
+	attrs, err := utils.MarshalFields(msg, "pubsub-attr")
+	if err != nil {
+		return "", err
```
DomBlack, line 124: Same here; I think all errors being returned to any application code should already be wrapped as our error type

runtime/pubsub/utils/utils.go
```diff
@@ -0,0 +1,187 @@
...
+	"reflect"
+	"strconv"
+	"time"
+)
+
+// MarshalFields creates a map[string]string of fields in `msg` tagged with `tag`. The name of the tag
+// will be used as map key, and values are converted to strings using fmt.Sprintf. Pointers will be dereferenced
+// and ignored if nil. Only basic types (bool, numeric, string) and pointers to those types are supported fields.
+// Fields without a tag will not be marshalled. `msg` must be a struct or pointer to a struct
+func MarshalFields[T any](msg T, tag string) (map[string]string, error) {
```
DomBlack, line 16: As a short-term future todo; we should look to add caching to these two marshal functions for types (as json.Marshal does)

As a long-term todo; we should add code generation to types to remove all need for reflection and looping here

runtime/pubsub/local.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}()
+	// create a message to unmarshal the raw nsq body into
+	msg := &message[T]{}
+	err = json.Unmarshal(m.Body, msg)
+	if err != nil {
+		return err
+	}
+	// unmarshal the attributes and write them to the message type
+	err = utils.UnmarshalFields(msg.Attributes, &msg.Data, "pubsub-attr")
```
DomBlack, line 68: Should we unmarshal the attributes into the message?

When we marshal the original message we don't remove them - so those attributes are already inside the `msg.Data` object at this point as `json.Unmarshal` has done this already.

So I guess my question here is - are we either:
1. Missing the code to _remove the attribute fields_ during the original marshalling?
2. Doing a redundant extra unmarshal here?

runtime/pubsub/utils/utils.go
```diff
@@ -0,0 +1,187 @@
+package utils
```
DomBlack, line 1: Should this be in `runtime/pubsub/internal/utils` - such that only code within the `runtime/pubsub` or children packages can access it (enforced by the compiler)

runtime/pubsub/local.go
ekerfelt: Hmm, yeah. I discussed this with Andre, and we concluded that we don't want to constrain the user too much. The user can choose to exclude the field from the body with `json:"-"`, but the default would be to allow either or. You're correct that we will do a redundant unmarshal, but at this point I wanted to avoid the complexity of implementing custom json serialization. 

I think the next step is indeed to add generated code, but wanted to avoid colliding with andre's changes

DomBlack (dismissed):

runtime/pubsub/utils/utils.go
ekerfelt: Added caching for the unmarshalling for now. The marshalling stuff is pretty much using the same encoder for everything.

DomBlack (approved):

PR #274, opened 2022-06-15
Parse `NewSubscription` and `Publish` pubsub calls

DomBlack:
This commit adds the ability for Encore to parse calls to
`NewSubscription` and `Publish` on created `pubsub.Topic` references,
recording those calls into the EST.

Other pieces of this commit:
 - internal `locations` package which allows us to specify the location of a `ast.Node`,
   as well as filters for if a node is allowed in that position or not.
 - internal `walker` package which combines the best bits of `ast.Walk`, `astutils.Apply`
   and `ast.Inspect`. This massively simplified the code needed to track the referenced
   resources and if the calls where allowed in those locations.


parser/internal/locations/location.go
```diff
@@ -0,0 +1,139 @@
...
+
+//go:generate stringer -type=Location
+
+// This list is ordered from most outer possible location, to most inner location
+const (
+	File         Location = 1 << iota // Inside a file, outside a function body
+	InitFunction                      // Inside an init function declaration
+	Function                          // Inside any function declaration (including the init function)
+	Variable                          // Inside a variable
+	endOfList
```
DomBlack, line 19: This isn't exported, but exists to allow the Describe method to loop over all possible flags.

parser/parser_references.go
```diff
@@ -0,0 +1,116 @@
...
+	// For all resources defined, store them in a map per package for faster lookup
+	resourceMap := make(map[string]map[string]est.Resource, len(p.pkgs)) // path -> name -> resource
+	for _, pkg := range p.pkgs {
+		resources := make(map[string]est.Resource, len(pkg.Resources))
+		resourceMap[pkg.ImportPath] = resources
+		for _, res := range pkg.Resources {
+			id := res.Ident()
+			resources[id.Name] = res
+		}
+	}
```
DomBlack, line 22: Copied from the code which currently tracks `sql.Named` calls - I plan to migrate that to this new path in a future PR (as well as cronjobs)

ekerfelt (approved):

PR #277, opened 2022-06-16
Resource Parser Tidy Up

DomBlack:
This PR refactors the parser resources to use the new `walker` package and brings that code inline with the resource reference parser and generally just tidies up the new parser code.


parser/parser_resources.go
```diff
@@ -2,101 +2,19 @@ package parser
...
-	}
-
-	resourceUsageRegistry[resourceType][name] = &resourceUsageParser{
-		Resource:         res,
-		Name:             name,
-		AllowedLocations: allowedLocations,
-		Parse:            parse,
-	}
-}
-
```
DomBlack, line 92: I moved this into `resources.go` and then tidied it up there.

parser/est/est.go
```diff
@@ -198,7 +193,8 @@ type AuthHandler struct {
 type Resource interface {
 	Type() ResourceType
 	File() *File
-	Ident() *ast.Ident
+	Ident() *ast.Ident  // the ident
```
eandre, line 196: Nit: redundant comment

parser/est/est.go
```diff
@@ -198,7 +193,8 @@ type AuthHandler struct {
...
+	NodeType() NodeType // The NodeType of the ast.Node that the resource is bound to
```
eandre, line 197: Almost a redundant comment

parser/internal/walker/cursor.go
```diff
@@ -42,13 +43,87 @@ func (c *Cursor) Location() (loc locations.Location) {
 	return
 }
 
+// DocComment will return the doc comment which is associated with the given node.
+// it works by walking through the parent nodes until it finds a node with a Comment field
+// or fields.
+//
+// It will stop walking when it comes across a node which represents a block where a previous
+// comment would no longer be valid (such as a FuncType, StructType, InterfaceType or BlockStmt)
```
eandre, line 51: Nit: use present tense: "DocComment returns" rather than "DocComment will return"

```suggestion
// DocComment returns the doc comment associated with the given node.
// It walks through the parent nodes until it finds a node with a Comment field or fields,
// and stops when it comes across a node which represents a block where a previous
// comment would no longer be valid (such as a FuncType, StructType, InterfaceType or BlockStmt).
```

parser/parser_references.go
```diff
@@ -43,13 +43,14 @@ func (r *resourceUsageVisitor) Visit(cursor *walker.Cursor) (w walker.Visitor) {
 		resource, parser := r.resourceAndFuncFor(node)
 		if parser != nil {
 			if parser.AllowedLocations.Allowed(cursor.Location()) {
-				parser.Parse(r.p, r.file, resource, node)
+				parser.Parse(r.p, r.file, resource, cursor, node)
 			} else {
 				call := fmt.Sprintf("`%s.%s`", resource.Ident().Name, parser.Name)
 				r.p.errf(node.Pos(),
-					"You can not call %s here, %s️.",
+					"You can not call %s here, %s️. For more information please see %s",
```
eandre, line 50: "please" sounds a bit strange in a compiler error. Also "cannot" instead of "can not". How about:
```suggestion
					"You cannot call %s here, %s️. For more information see %s",
```

parser/parser_references.go
```diff
@@ -58,8 +59,14 @@ func (r *resourceUsageVisitor) Visit(cursor *walker.Cursor) (w walker.Visitor) {
 
 	case *ast.SelectorExpr:
 		if resource := r.resourceFor(node); resource != nil {
-			// TODO(domblack)
-			r.p.errf(node.Pos(), "A %s can not be used in this location. 👷‍♂️\n", resource.Type())
+			// If the resource type isn't registered, for now this is Ok as we have SQLDB resources that are not tracked
+			if res, found := resourceTypes[resource.Type()]; found {
+				r.p.errf(node.Pos(),
+					"A %s can not be referenced, apart from when calling a method on it. For more information please see %s",
```
eandre, line 65:
```suggestion
					"A %s can not be referenced, apart from when calling a method on it. For more information see %s",
```

parser/parser_resources.go
```diff
@@ -111,63 +29,64 @@ type resourceCreationVisitor struct {
...
-				}
 
-				if walkSpec {
-					ast.Walk(walkerFunc(f.VisitAndReportInvalidCreationCalls), spec)
+					f.file.Pkg.Resources = append(f.file.Pkg.Resources, resource)
 				}
+			} else {
+				f.p.errf(
+					node.Pos(),
+					"A %s can not be declared here, %s. For more information please see %s",
```
eandre, line 69:
```suggestion
					"A %s can not be declared here, %s. For more information see %s",
```

parser/parser_resources.go
```diff
@@ -111,63 +29,64 @@ type resourceCreationVisitor struct {
...
-		}
-
 	case *ast.SelectorExpr, *ast.IndexExpr, *ast.IndexListExpr:
+		// If we find a selector (`a.foo`) an index (`a.foo[bar]`) or an index list (`a.foo[bar, baz]`)
+		// then we want to check if that references a resource creation function and if so
+		// report an error, as all valid usages should already have been parsed and returned
 		if parser := f.parserFor(node); parser != nil {
-			f.p.errf(node.Pos(), "%s.%s can only be used to declare package level variables. For more information please see %s\n", parser.Resource.PkgName, parser.Name, parser.Resource.Docs)
-			return false
+			f.p.errf(node.Pos(), "%s.%s can only be called as a function to create a new instance and not referenced otherwise. For more information please see %s", parser.Resource.PkgName, parser.Name, parser.Resource.Docs)
```
eandre, line 84:
```suggestion
			f.p.errf(node.Pos(), "%s.%s can only be called as a function to create a new instance and not referenced otherwise. For more information see %s", parser.Resource.PkgName, parser.Name, parser.Resource.Docs)
```### ###

File: parser/internal/names/types.go [hunk 9]
```diff
@@ -15,6 +15,61 @@ import (
...
+// If the given node is:
+// - an ident, then it is resolved to a package level deceleration in the same file as the ident.
+// - a selector, then it is resolved to an exported ident from the selected package.
+// - any other node will result in nothing being returned
+//
+// Note:
+// - This function accounts for scoping and variable shadowing. (i.e. it will not mistake a variable "foo" in a function
+//   for the package imported as "foo").
+// - This function does not validate the existence of the objName for objects referenced in other packages. As such it
+//   can return references for objects in Go Modules.
```
eandre, line 35: can return? Should be cannot return, I believe?

parser/internal/names/types.go
```diff
@@ -15,6 +15,61 @@ import (
...
+	if !ok {
+		return "", "", nil
+	}
+
+	// Resolve the type arguments
+	var typArgs []ast.Expr
+	switch typNode := node.(type) {
+	case *ast.IndexExpr:
+		node = typNode.X
+		typArgs = []ast.Expr{typNode.Index}
```
eandre, line 47: I think we should not blindly unwrap index list expressions here; it becomes imprecise when we don't know whether something accepts type parameters. It would accept code like `var blah = cron.NewJob[int]()` even though `cron.NewJob` isn't generic.

It would also say that `otherpkg.Blah[0]` is a package object expression, but it's not. `otherpkg.Blah` is a package object.

If we want to do unwrapping of type parameters we should require the caller to decide whether type parameters are permissible in a particular location, and then we should arguably parse those type arguments as well. That feels like it belongs in the `parser` package rather than the `names` package though so I'd prefer if we keep the "unwrap index expressions" code outside of the package object resolution

parser/internal/names/types.go
DomBlack: No it can.

For instance, you could import `encore.dev/cron` and use `cron.NewJob`. This function will resolve that and understand that there is an exported object called `NewJob` inside the `encore.dev/cron` package - however, it will not validate if `NewJob` actually exists or not.

eandre: Ah, but it doesn't work for arbitrary packages, only for imports we've explicitly decided to track, correct?

parser/internal/names/types.go
DomBlack: I've updated the docs and return names to say `indexArgs` here.

For name resolution, I think this is still valid to have, as it's operating over all names within the application.

> It would also say that otherpkg.Blah[0] is a package object expression

This function say it's a reference to a package object - which it is with index args being type args _or_ an actual index into a map or slice of a package object.

parser/internal/names/types.go
DomBlack: Yeah; true

eandre (approved):

PR #278, opened 2022-06-16
Improve cors handling for local development

eandre:


cli/daemon/run/run.go
```diff
@@ -481,7 +481,9 @@ func (r *Run) generateConfig(p *Proc, params *startProcParams) *config.Runtime {
 		SQLServers:    sqlServers,
 		AuthKeys:      []config.EncoreAuthKey{p.authKey},
 		CORS: &config.CORS{
-			AllowOriginsWithCredentials:    []string{"localhost"},
+			// Allow all origins with credentials for local development
+			AllowOriginsWithCredentials: []string{"UNSAFE_ALL_ORIGINS_WITH_CREDENTIALS"},
```
DomBlack, line 485: Could we not reference the const exported from the config package - rather than having two separate copies of the magic string?

DomBlack (changes requested):

DomBlack (approved):

PR #284, opened 2022-06-21
Refactor resource service startup

DomBlack:
This commit refactors the resource service startup, such
that if you add code for a SQL database or pubsub topic when
`encore run` is already running, the watch will pick up those
changes and start the required services on your system.

This involved:
- Moving the resource startup code into a new `ResourceServices` struct
- Moving the `OpTracker` into it's own package from `cli/daemon/run` so it could be passed around
- Making `OpTracker` reciever functions work will `nil` pointers (i.e. no-op) so the calling code doesn't need to check

While refactoring this code, I added a `asyncBuildJobs` which allows jobs which can be run in paralel to be run in paralel.

I also added an NSQ log adapter and limited the logs to warning's and above.

This means the outputs from the daemon and CLI run now look like this:
image

### DIFF
--- a/cli/cmd/encore/main.go
+++ b/cli/cmd/encore/main.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/rs/zerolog"
@@ -156,7 +157,7 @@ func streamCommandOutput(stream commandOutputStream, convertJSON bool) int {
 			case st.Code() == codes.FailedPrecondition:
 				fmt.Fprintln(os.Stderr, st.Message())
 				return 1
-			case err == io.EOF || st.Code() == codes.Canceled:
+			case err == io.EOF || st.Code() == codes.Canceled || strings.HasSuffix(err.Error(), "error reading from server: EOF"):
 				return 0
 			default:
 				log.Fatal().Err(err).Msg("connection failure")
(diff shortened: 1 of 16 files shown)


cli/cmd/encore/main.go
```diff
@@ -156,7 +157,7 @@ func streamCommandOutput(stream commandOutputStream, convertJSON bool) int {
 			case st.Code() == codes.FailedPrecondition:
 				fmt.Fprintln(os.Stderr, st.Message())
 				return 1
-			case err == io.EOF || st.Code() == codes.Canceled:
+			case err == io.EOF || st.Code() == codes.Canceled || strings.HasSuffix(err.Error(), "error reading from server: EOF"):
```
DomBlack, line 160: For some reason the clean `EOF` error wasn't being picked up (I think it was because it was wrapped inside a GRPC error)

compiler/internal/codegen/codegen_main.go
```diff
@@ -142,7 +142,10 @@ func (b *Builder) Main() (f *File, err error) {
 	f.Line()
 
 	f.Func().Id("main").Params().Block(
-		If(Err().Op(":=").Qual("encore.dev/runtime", "ListenAndServe").Call(), Err().Op("!=").Nil()).Block(
+		If(
+			Err().Op(":=").Qual("encore.dev/runtime", "ListenAndServe").Call(),
+			Err().Op("!=").Nil().Op("&&").Err().Op("!=").Qual("io", "EOF"),
```
DomBlack, line 147: Ths addition of the `&& err != io.EOF` stops the Encore app from reporting a fatal error when it's shutdown cleanly

ekerfelt (approved):

PR #285, opened 2022-06-22
Add inital GCP PubSub code & Tracing

DomBlack:
This commit adds the initial support for GCP PubSub which it
supports along side NSQ.

This also adds full tracing support for tracing pub sub publish
and subscriptions.

To achieve this combined work, I've refactored a little bit in
the runtime to make the pubsub work easier

# Screenshots

image
image
image


cli/daemon/engine/trace/trace.go
```diff
@@ -691,6 +707,42 @@ func (tp *traceParser) logField() (*tracepb.LogField, error) {
...
+	publishID := tp.UVarint()
+	spanID := tp.Uint64()
+	req, ok := tp.reqMap[spanID]
+	if !ok {
+		return fmt.Errorf("unknown request span: %v", spanID)
+	}
+
+	publish := &tracepb.PubsubMsgPublished{
+		Goid:      tp.UVarint(),
+		CallLoc:   0, // TODO(eandre) reintroduce (not sure why ;) )
```
eandre, line 720: I'm in the process of removing all of these so no need to reintroduce :)

runtime/pubsub/internal/nsq/logAdapter.go
```diff
@@ -0,0 +1,54 @@
+package nsq
```
eandre, line 1: Nit: no camelCase in file names: `log_adapter.go` or `logadapter.go`

runtime/pubsub/topic.go
```diff
@@ -24,68 +28,155 @@ type Topic[T any] interface {
...
+		}
+
+		err = panicCatchWrapper(ctx, msg)
+		runtime.FinishRequest(nil, err)
+		return err
+	})
+
+	log.Info().Msg("registered subscription")
+
+	return struct{}{}
```
eandre, line 147: How does `struct{}{}` implement `Subscription[T]`? Or am I misunderstanding something?

runtime/pubsub/internal/gcp/topic.go
```diff
@@ -0,0 +1,86 @@
...
+	// Attempt to publish the message
+	return t.gcpTopic.Publish(ctx, gcpMsg).Get(ctx)
+}
+
+func (t *topic) Subscribe(logger *zerolog.Logger, _ *types.SubscriptionConfig, cfg *config.PubsubSubscription, f types.RawSubscriptionCallback) {
+	// Create the subscription object (and then check it exists on GCP's side)
+	subscription := t.client.Subscription(cfg.CloudName)
+	exists, err := subscription.Exists(ctx.App)
+	if err != nil {
+		panic(fmt.Sprintf("pubsub subscription %s for topic %s status call failed: %s", cfg.EncoreName, t.cfg.EncoreName, err))
```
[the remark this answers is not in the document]
DomBlack: Hmm; I'm not sure I agree here.

If you're deploying a new instance, you don't want it to replace the healthy instance if some tweak or change means the new instance is unable to connect to your topics.

This panic prevents the new instance from coming up; thus keeping the old instance active. If there's a temporary GCP outage while an existing instance is running; then the loop on line 64 of this file will keep the subscription retrying until GCP is backup.

i.e. to hit this panic either GCP needs to be down _while you are deploying_ or something is wrong with your seutp.

runtime/pubsub/topic.go
```diff
@@ -24,68 +28,155 @@ type Topic[T any] interface {
...
+	// Subscribe to the topic
+	t.topic.Subscribe(&log, cfg, subscription, func(ctx context.Context, msgID string, publishTime time.Time, deliveryAttempt int, attrs map[string]string, data []byte) (err error) {
+		runtime.BeginOperation()
+		defer runtime.FinishOperation()
+
+		var msg T
+
+		if err = json.Unmarshal(data, &msg); err != nil {
+			err = errs.B().Cause(err).Code(errs.InvalidArgument).Msg("failed to unmarshal message").Err()
+			log.Err(err).Str("msg-id", msgID).Int("delivery-attempt", deliveryAttempt).Msg("failed to unmarshal message")
```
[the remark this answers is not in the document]
DomBlack: We have a mismash of `snake_case` and `camelCase` right now (i.e both `app_id` and `appId`)

I'll convert everything to `snake_case`

runtime/pubsub/internal/gcp/topic.go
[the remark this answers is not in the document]
eandre: If GCP has a PubSub outage that causes this call to fail, this will cause any instance to crash on startup. That means autoscaling and node migrations breaks, not to mention deployments. We shouldn't bring down the whole application just because PubSub is down.

runtime/pubsub/topic.go
DomBlack: `Subscription[T]` is an empty interface, so the struct matches that need. But it means that these functions are already returning a value, so if we add a method to `Subscription[T]` this will cause a nice compile error

eandre: Ah, nice :)

runtime/pubsub/internal/gcp/topic.go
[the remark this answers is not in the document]
DomBlack: I think this maybe is a more fundamental question about Encore deployments;

Should replica's fail to startup _if_ they are not healthy? Or should we let unhealthy replica's start assuming whatever is making them unhealthy is a transient error?

My default would be if a replica fails it's health checks, it should not be allowed to start serving traffic.

So the next question is; do we assume that being unable to subscribe to your queues is unhealthy? And if not; do we believe risking zero consumers of your queue is healthy?

eandre (approved):

cli/daemon/run/run.go
```diff
@@ -517,17 +517,32 @@ func (r *Run) generateConfig(p *Proc, params *startProcParams) *config.Runtime {
...
-				EncoreName: t.Name,
+			topicCfg := &config.PubsubTopic{
+				ServerID:      0,
+				EncoreName:    t.Name,
+				CloudName:     t.Name,
+				Subscriptions: make(map[string]*config.PubsubSubscription),
 			}
+
+			if t.Ordered && t.GroupedBy != nil {
+				topicCfg.OrderingKey = *t.GroupedBy
```
ekerfelt, line 535: Shoudn't we be consistent in naming and call this GroupedBy?

runtime/pubsub/topic.go
```diff
@@ -2,15 +2,19 @@ package pubsub
...
+	"encore.dev/pubsub/internal/gcp"
+	"encore.dev/pubsub/internal/nsq"
+	"encore.dev/pubsub/internal/types"
+	"encore.dev/pubsub/internal/utils"
+	"encore.dev/runtime"
+	"encore.dev/runtime/config"
+)
 
 // Topic is the entry point for published messages
 type Topic[T any] interface {
```
ekerfelt, line 20: Should we move this to types?

ekerfelt (approved):

PR #288, opened 2022-06-23
Add Push handler for GCP PubSub Push

DomBlack:
This commit adds a new Encore HTTP handler
allowing subscriptions to register their own
HTTP handler to enable messages to be pushed to the
application, rather than pulled.

This is to allow apps to scale down to zero, but
still be able to respond to messages on the topic.


runtime/runtime/encore_routes.go
```diff
@@ -0,0 +1,68 @@
...
+			EncoreCompiler string `json:"encore_compiler"`
+			DeployId       string `json:"deploy_id"`
+		}{
+			AppRevision:    config.Cfg.Static.AppCommit.AsRevisionString(),
+			EncoreCompiler: config.Cfg.Static.EncoreCompiler,
+			DeployId:       config.Cfg.Runtime.DeployID,
+		},
+	})
+	_, _ = w.Write(bytes)
+}
```
DomBlack, line 49: Moved from `setup.go`

runtime/runtime/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	keyID := binary.BigEndian.Uint32(macBytes[:keyIDLen])
+	mac := macBytes[keyIDLen:]
+	for _, k := range config.Cfg.Runtime.AuthKeys {
+		if k.KeyID == keyID {
+			return checkAuth(k, req, mac), nil
+		}
+	}
+
+	return false, nil
+}
```
DomBlack, line 53: Moved from `setup.go`

runtime/beta/errs/error.go
```diff
@@ -141,6 +141,8 @@ func (e *Error) Unwrap() error {
 
 func HTTPError(w http.ResponseWriter, err error) {
 	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
```
eandre, line 144: What's this? Probably deserves a comment

runtime/pubsub/internal/gcp/push_handler.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func registerPushEndpoint(cfg *config.PubsubSubscription, f types.RawSubscriptionCallback) {
+	runtime.RegisterPubSubSubscriptionHandler(
+		cfg.ResourceID,
+		func(req *http.Request) error {
+			// If the request has not come from the Encore platform it must have
+			// a valid JWT set by Google.
+			if !runtime.IsEncoreAuthenticatedRequest(req.Context()) {
+				if err := validateGoogleJWT(req); err != nil {
```
eandre, line 36: We need to validate the audience and service account email, otherwise this is easy to circumvent by just getting another JWT from Google for some other purpose

runtime/runtime/setup.go
```diff
@@ -107,40 +101,38 @@ func (srv *Server) handler(w http.ResponseWriter, req *http.Request) {
...
-	// We use EscapedPath rather than `req.URL.Path` because if the path contains an encoded
-	// forward slash as %2F we don't want the router to treat that as a segment split.
-	//
-	// i.e. `/foo%2Fbar/baz` should be routed to `/:a/*b` as a = "foo/bar", b = "baz"
-	// where as if we use req.URL.Path we would get a = "foo", b = "bar/baz` which is incorrect.
-	path := req.URL.EscapedPath()
+	// Switch to the Encore internal router if we are on a Encore internal path
+	// Historically we used a prefix of `/__encore` but this is now deprecated in favour of a well-known `encore` path
+	// But for now we support both
+	if strings.HasPrefix(path, "/__encore/") || strings.HasPrefix(path, "/.well-known/encore/") {
```
eandre, line 118: Not sold on `.well-known/encore` over `__encore`. It seems primarily used for informational things and not for APIs. What's the reason for moving it?

runtime/beta/errs/error.go
DomBlack: From: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
> The X-Content-Type-Options response HTTP header is a marker used by the server to indicate that the [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) advertised in the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) headers should be followed and not be changed. The header allows you to avoid [MIME type sniffing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#mime_sniffing) by saying that the MIME types are deliberately configured.

Basically, I noticed that `http.Error` had this in it, but our `errs.HTTPError` did not. Seemed like a good idea to bring it over from the standard library.

eandre: Ah nice!

runtime/pubsub/internal/gcp/push_handler.go
DomBlack: @ekerfelt this means we need to add the service account to the Topic's config, so I can validate against it. ℹ️

runtime/runtime/setup.go
DomBlack: They're not just for informational use, but also "well-known services". As such I've seen a couple pieces of software put their health endpoint inside the well-known directory ([for example, Apollo](https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/#http-level-health-checks)).

The reasoning is this feels more inline with not stomping over an application URL space, and putting it in a clearly well known location with a clear identifier for what we're talking about.

runtime/pubsub/internal/gcp/push_handler.go
ekerfelt: 👍

runtime/runtime/encore_routes.go
```diff
@@ -0,0 +1,68 @@
...
+	"encore.dev/beta/errs"
+	"encore.dev/runtime/config"
+)
+
+var pubSubSubscriptions = make(map[string]func(r *http.Request) error)
+
+// RegisterPubSubSubscriptionHandler registers a handler for the given PubSub subscription
+//
+// This is an internal Encore API and should not be used.
+func RegisterPubSubSubscriptionHandler(subscriptionID string, handler func(r *http.Request) error) {
```
ekerfelt, line 17: is there some way we could avoid exporting this?

ekerfelt (approved):

PR #289, opened 2022-06-24
Support for Next.JS projects in generated clients

DomBlack:
This commit adds support for generated Typescript clients to be
compitable with [Next.JS](https://nextjs.org/) applications with the new `--preset nextjs` argument.

The existing client was not compiatble as we where using namespaces
to mimic the packages within the Encore application, however namespaces
are somewhat depcreated and not all transpliers support them; the key one
here being Babel (which Next.JS uses under the hood).

When turned on the new `--preset nextjs` flag, therefore prevents the code generator
from emitting namespaces, and instead it uses the package name as a prefix for
generated types; such as `$package_$struct`.

To make NextJS development easier, this flag also tells the TypeScript code generator
to generate [SWR](https://swr.vercel.app/) wrappers for all API's, while maintaining the
promise based functions if direct calls are still required (i.e. to perform actions).

If the original api call was:
```
const api = new Client()

const rsp: Promise<Data> = api.MySvc.MyAPI("segment", { name: "bob" })
```

The SWR wrapper will be:
```
const api = new Client() // this needs to be a singleton for the application instance

const rsp: SWRResponse<Data> = api.MySvc.useMyAPI("segment", { name: "bob" }, { refreshInternval: 1000 })
```


eandre:
This looks great! My only concern is that Next.JS is a moving target, and I can imagine lots of other scenarios where the same code generation options (no namespaces, `swr` support) are desirable by others who don't necessarily use Next. That would lead to the regrettable situation where if Next.JS changes, we'd be torn between updating the `--nextjs` flag to match the new behavior (and breaking other users who may still want the old behavior), and introducing a new option to select the new behavior (causing `--nextjs` to lose its intended meaning).

I think a better solution is to introduce code generation options that map to specific functionality rather than moving targets. So in this case we could have `--ts.namespaces=false` and `--ts.swr=true` or something to that effect.

DomBlack:
@eandre; I've updated the CLI options and per your suggestion an introduced a `--preset nextjs`

cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -214,12 +235,16 @@ export namespace echo {
 
             // Now make the actual call to the API
             const resp = await this.baseClient.callAPI("POST", `/echo.HeadersEcho`, undefined, {headers})
+            try {
```
eandre, line 239: With the addition of `try {` this newline seems out of place. Remove?

cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -190,7 +203,11 @@ export namespace echo {
         public async EmptyEcho(params: EmptyData): Promise<EmptyData> {
             // Now make the actual call to the API
             const resp = await this.baseClient.callAPI("POST", `/echo.EmptyEcho`, JSON.stringify(params))
-            return await resp.json() as EmptyData
+            try {
+                return await resp.json() as EmptyData
+            } catch (err) {
+                throw new APIError(500, { code: ErrCode.DataLoss, message: "unable to unmarshal the response", details: err })
```
eandre, line 209: `DataLoss` is not meant for this but rather fatal, permanent data loss on the server side. Change to `Internal`

cli/internal/codegen/client.go
```diff
@@ -44,7 +45,7 @@ func Detect(path string) (lang Lang, ok bool) {
 }
 
 // Client generates an API client based on the given app metadata.
-func Client(lang Lang, appSlug string, md *meta.Data) (code []byte, err error) {
+func Client(lang Lang, appSlug string, md *meta.Data, tsOptions *daemonpb.GenClientRequest_TypeScriptOptions) (code []byte, err error) {
```
eandre, line 48: I guess making the typescript options part of this signature is intended as a short-term thing? Ideally a nicer interface would be nicer to represent `lang+options` as a single protobuf message with a `oneof`

cli/internal/codegen/client.go
```diff
@@ -54,7 +55,11 @@ func Client(lang Lang, appSlug string, md *meta.Data) (code []byte, err error) {
 	var gen generator
 	switch lang {
 	case LangTypeScript:
-		gen = &typescript{generatorVersion: typescriptGenLatestVersion}
+		if tsOptions == nil {
+			tsOptions = &daemonpb.GenClientRequest_TypeScriptOptions{Namespaces: true}
```
eandre, line 59: Instead of doing this here we could make the client always pass in options as it's the thing that's aware of the presets, and the base case of "no preset" is essentially just the default preset

cli/internal/codegen/testdata/nextjs_expected_noauth_typescript.ts
```diff
@@ -0,0 +1,447 @@
...
+        return useSWR(
+            [`/svc.DummyAPI`, params],
+            () => this.DummyAPI(params),
+            swrConfig
+        )
+    }
+}
+
+
+
```
eandre, line 84: Lots of empty newlines here, is that intended?

cli/internal/codegen/testdata/nextjs_expected_noauth_typescript.ts
```diff
@@ -0,0 +1,447 @@
...
+    const pairs = []
+    for (const key in parts) {
+        const val = (Array.isArray(parts[key]) ?  parts[key] : [parts[key]]) as string[]
+        for (const v of val) {
+            pairs.push(`${key}=${encodeURIComponent(v)}`)
+        }
+    }
+    return pairs.join("&")
+}
+// CallParameters is the type of the parameters to a method call, but require headers to be a Record type
```
eandre, line 95: Should be a newline here

cli/internal/codegen/testdata/nextjs_expected_noauth_typescript.ts
```diff
@@ -0,0 +1,447 @@
...
+type CallParameters = Omit<RequestInit, "method" | "body"> & {
+    /** Any headers to be sent with the request */
+    headers?: Record<string, string>;
+
+    /** Any query parameters to be sent with the request */
+    query?: Record<string, string | string[]>
+}
+
+
```
eandre, line 104: Should be single newline?

cli/cmd/encore/gen.go
```diff
@@ -37,15 +44,29 @@ Use '--env=local' to generate it based on your local development version of the
...
 			appID := args[0]
 
-			if lang == "" {
+			if preset != "" {
+				switch strings.ToLower(preset) {
+				case "nextjs", "next.js":
+					lang = string(codegen.LangTypeScript)
+					tsOptions = &daemonpb.GenClientRequest_TypeScriptOptions{
+						Namespaces: false,
+						Swr:        true,
```
eandre, line 64: I'm not sold on coupling swr with next.js. Yes, swr is also developed by the same authors but I believe React Query is increasingly becoming the library of choice for this sort of thing.

What's more, the `useBlah` helpers are only valid for endpoints that fetch data, not for mutations, since they regularly fire multiple times in the background. Unintentionally calling a stateful, mutating endpoint when not intending to is recipe for disaster, so unless we have a good way of representing which endpoints actually do that and which don't we shouldn't generate such wrappers I don't think.

cli/cmd/encore/gen.go
```diff
@@ -37,15 +44,29 @@ Use '--env=local' to generate it based on your local development version of the
 Supported language codes are:
   typescript: A TypeScript-client using the in-browser Fetch API
   go: A Go client using net/http"
+
+Support presets are:
+	nextjs: Generates a TypScript client for use within a Next.js application
```
eandre, line 49:
```suggestion
Support presetsed are:
  nextjs: Generates a TypScript client for use within a Next.js application
```

cli/cmd/encore/gen.go
```diff
@@ -37,15 +44,29 @@ Use '--env=local' to generate it based on your local development version of the
...
 `,
 		Args: cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
-			if output == "" && lang == "" {
-				fatal("specify at least one of --output or --lang.")
+			if output == "" && lang == "" && preset == "" {
+				fatal("specify at least one of --output, --lang or --preset.")
```
eandre, line 54: I am concerned that this makes for quite the analysis paralysis when trying to decide how to use this command. Could we combine `lang` and `preset` somehow to just a single option? We could have `--lang=ts` choose the default preset and `--lang=ts.nextjs` (or `--lang=ts/nextjs`) choose the nextjs preset, or whatever.

DomBlack:
I'm going to close this PR for now as I'm not sure we need it anymore.

We can re-open it if we needed it agian

PR #291, opened 2022-06-27
Add tool for generating Public API: encore.dev

DomBlack:
This commit adds a tool which allows us to easily
generate the [encore.dev](https://github.com/encoredev/encore.dev)
public API from the runtime implementation.

This auto generation means we can be sure to easily keep the API in-step
with the actual implementation, without missing docs, comments or constants.


runtime/cron/cron.go
```diff
@@ -1,19 +1,74 @@
...
+//
+// 		// SendWelcomeEmail emails everyone who signed up recently.
+// 		// It's idempotent: it only sends a welcome email to each person once.
+// 		//encore:api private
+// 		func SendWelcomeEmail(ctx context.Context) error {
+// 			// ...
+// 			return nil
+// 		}
+func NewJob(id string, jobConfig JobConfig) *Job {
+	panic("encore apps must be run using the encore command")
```
eandre, line 37: This shouldn't be here I assume?

tools/publicapigen/main.go
```diff
@@ -0,0 +1,463 @@
...
+var (
+	fset      = token.NewFileSet()
+	files     = map[string]*ast.File{}
+	constants = map[string]map[string]ast.BasicLit{}
+	types     = map[string]map[string]ast.Expr{}
+)
+
+func main() {
+	log.Logger = zerolog.New(zerolog.NewConsoleWriter()).With().Timestamp().Caller().Logger()
+	log.Info().Msg("Generating Public API")
```
eandre, line 32: Lots of uppercase log messages here :) I guess you can keep them this time!

tools/publicapigen/main.go
```diff
@@ -0,0 +1,463 @@
...
+
+		if err := walkDir(dir2, rel2, f); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func repoDir() string {
+	_, file, _, ok := runtime.Caller(0)
```
eandre, line 384: The output of `runtime.Caller(0)` will involve file paths based on where the binary was built, which means this could cause surprising results if run on someone else's machine. I guess that's quite unlikely and it doesn't look like we're doing anything bad with the return value (if we called `os.RemoveAll` or something that could be bad)

tools/publicapigen/main.go
```diff
@@ -0,0 +1,463 @@
...
+	if !ok {
+		panic("Unable to get caller location")
+	}
+
+	publicapigenDir := filepath.Dir(file)
+	toolsDir := filepath.Dir(publicapigenDir)
+	encoreDir := filepath.Dir(toolsDir)
```
eandre, line 391: Instead of hard-coding these paths, a more robust way is to run:

```
cmd := exec.Command("git", "rev-parse", "--show-toplevel")
out, err := cmd.CombinedOutput()
// handle err
return string(bytes.TrimSpace(out))
```

tools/publicapigen/main.go
```diff
@@ -0,0 +1,463 @@
...
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func isPrivateFile(fileName string) bool {
+	return strings.HasPrefix(fileName, "runtime/") ||
```
eandre, line 460: Are we sure these are always prefixes (and relative paths as a result) and not absolute paths?

eandre (commented):
This is awesome, really nice work! A few minor comments but overall this is great.

tools/publicapigen/main.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	node.List = []*ast.Comment{node.List[0]}
+	node.List[0].Text = "  " // double space to prevent a panic when printing the file back out
+}
+
+func mustKeep(nodes ...*ast.CommentGroup) bool {
+	for _, node := range nodes {
+		if node != nil && node.List != nil {
+			for i, comment := range node.List {
+				if comment.Text == "//publicapigen:keep" {
```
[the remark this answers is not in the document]
DomBlack: I don't think it's super needed, but added it as it's a nice enhancement if an IDE isn't setup to trim trailing spaces.

eandre (approved):

PR #294, opened 2022-06-29
Enable Tests to be written against PubSub topics

DomBlack:
This commit adds a new implementation of a pubsub topic; the test implementation. This Implementation is automatically aware of the test currently running, and isolates it's behaviour to that test.

This means you can run two separate tests in parallel, both of which interact with the same `pubsub.Topic`, yet they will only observe the side-effects of their own published messages.

To test a pubsub topic;

```go
import (
    "context"
    "testing"

    "github.com/stretchr/testify/require"

    "encore.dev/pubsub"
    "encore.dev/test"
)

type Msg struct { Str string }

var myTopic = pubsub.NewTopic[Msg]("my-topic")

func TestOne(t *testing.T) {
    t.Parallel()

    err := myTopic.Publish(context.Background(), Msg{"Hello World"})
    require.NoError(t, err)

    // even though these two tests run in parallel, the message from TestTwo can not appear here.
    messages := test.Topic(myTopic).PublishedMessages()
    require.Len(t, messages, 1)
    require.Equal(t, "Hello World", messages[0].Str)
}

func TestTwo(t *testing.T) {
    t.Parallel()

    err := myTopic.Publish(context.Background(), Msg{"Goodbye World"})
    require.NoError(t, err)
}
```

To achieve this we've added four new linked functions between the Go runtime and the Encore runtime in encoredev/go#8 . These functions inform the Encore runtime when tests are started, paused, resumed and finished. The Encore runtime then tags the running go routine with the test information. (See the new `test_interface.go` file)

This then allows the new `pubsub/test.Topic` implementation to access the current test, and assign it a specific implementation instance of a topic.

The `test.Topic()` function then uses a private API to extract that specific instance for a test and casts it into a public interface, allowing user code to perform assertions upon the messages published on the topic.

Future expansions to this implementation will allow the user to forcible test failure scenarios.


runtime/pubsub/internal/utils/utils.go
```diff
@@ -1,17 +1,35 @@
...
 	"sync"
 	"time"
+
+	"encore.dev/beta/errs"
 )
 
 const AttrTag = "pubsub-attr"
 
+// UnmarshalMessage unmarshals a message into a struct. The message must be a JSON object.
+func UnmarshalMessage[T any](attrs map[string]string, data []byte) (msg T, err error) {
```
DomBlack, line 19: I moved this from the `NewSubscription` function, so it could be reused by the `testImplementation` as well, without duplicating the code.

That way we _know_ that in tests the real marshal and unmarshal code is being fully executed.

runtime/pubsub/test_internal.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func GetTestTopicInstance[T any](topic *Topic[T]) any {
+	testTopic, ok := topic.topic.(*test.TestTopic[T])
+	if !ok {
+		panic("testTopic not called with a test topic")
+	}
+
+	return testTopic.TestInstance(runtime.CurrentTest())
+}
```
DomBlack, line 15: This isn't a public API (hence the `_internal` suffix on the file).

This is the linking piece between the `pubsub` unexported stuff, and the new `test` package

runtime/runtime/request.go
```diff
@@ -66,8 +68,10 @@ type Request struct {
 	PathSegments httprouter.Params
 	MsgData      PubSubMsgData
 	Start        time.Time
-	Logger       zerolog.Logger
+	Logger       *zerolog.Logger
```
DomBlack, line 71: Made this a pointer, as we had usecases where the zero value was being used by code to attempt to log - and it logged nowhere. Now the `Logger()` function can see if its' been set, and if not default to the RootLogger instead

runtime/test/package.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"encore.dev/runtime/config"
+)
+
+func init() {
+	if !config.Cfg.Static.Testing {
+		panic("You can only import the test package from within a test. None of this packages functionality will work outside of a call to `encore test`")
+	}
+}
```
DomBlack, line 13: Given the intended use for this package is only for testing; and it's going to be deeply tied into the runtime's test support - I think this is a good panic, as it prevents people trying to use functionality provided in this package in their running apps.

eandre: It's really frustrating to debug these issues, from experience. I think if we want to prevent this we should do it at compile time by inspecting the imports in the parser step, so we can point at where the problem is

runtime/test/package.go
```diff
@@ -0,0 +1,13 @@
+// Package test provides a number of functions and tools for writing fully integrated test suites for Encore
+// applications.
+package test
```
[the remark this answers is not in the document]
DomBlack: I think test is fine personally. 

It doesn’t overlap with go’s testing package and if it conflicts with a customers test helpers package; then it can be aliased. 

When we make the expectation and matcher api’s would we call the packages ematching & eexpect. 

I think that impacts readability too much. Given go allows for aliasing of imports I’m not overly concerned by this

runtime/test/package.go
DomBlack: Fair. Will change to a compiler error if we see an import for this outside test packages

ekerfelt (approved):

PR #296, opened 2022-06-30
Update trace parsing

DomBlack:
This commit updates the trace parsing to bring it inline with the
protobuf we use for traces in the long term storage engine and allows
the Parse function to be called from our engine directly.

This removes the need for us to maintain two copies of this proto and
parsing logic.

### DIFF
--- a/cli/daemon/engine/trace/trace.go
+++ b/cli/daemon/engine/trace/trace.go
@@ -80,7 +80,7 @@ func (st *Store) List(appID string) []*TraceMeta {
 	return tr
 }
 
-func Parse(log *zerolog.Logger, traceID ID, data []byte, version trace.Version, symTable SymTable) ([]*tracepb.Request, error) {
+func Parse(log *zerolog.Logger, traceID ID, data []byte, version trace.Version, symTable SymTabler) ([]*tracepb.Request, error) {
 	id := &tracepb.TraceID{
 		Low:  bin.Uint64(traceID[:8]),
 		High: bin.Uint64(traceID[8:]),
@@ -110,7 +110,7 @@ type goKey struct {
 	goid   uint32
 }
 
-type SymTable interface {
+type SymTabler interface {
 	SymTable(ctx context.Context) (*sym.Table, error)
 }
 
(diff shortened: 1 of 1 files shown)


cli/daemon/dash/trace.go
```diff
@@ -323,9 +323,7 @@ func (tp *traceParser) parseReq(req *tracepb.Request) (*Request, error) {
 		MessageID:      req.MessageId,
 		Attempt:        req.Attempt,
 		Published:      nil,
-
-		CallLoc: nullInt32(req.CallLoc),
-		DefLoc:  req.DefLoc,
+		DefLoc:         req.DefLoc,
```
DomBlack, line 326: Even though we don't use in on the hosted version, the local dashboard still replies on this for now.

cli/daemon/engine/trace/trace.go
```diff
@@ -78,14 +80,16 @@ func (st *Store) List(appID string) []*TraceMeta {
 	return tr
 }
 
-func Parse(traceID ID, data []byte, proc *run.Proc) ([]*tracepb.Request, error) {
+func Parse(log *zerolog.Logger, traceID ID, data []byte, version trace.Version, symTable SymTable) ([]*tracepb.Request, error) {
```
DomBlack, line 83: This is a merge of the two copies of this code.

Basically keeping the extra stuff the local dev does, while allowing the platform to use it and have backwards compatibility with older stored traces

proto/encore/engine/trace/trace.proto
```diff
@@ -36,9 +36,11 @@ message Request {
   string endpoint_name = 18;
 
   // Fields set if Type == PUBSUB_MSG
-  string message_id = 19;
-  uint32 attempt = 20;
-  uint64 publish_time = 21;
+  string topic_name = 19;
+  string subscription_name = 20;
```
DomBlack, line 40: These are needed for the platform's trace view, as we don't have access to the defloc's to extra this in the UI like the local dev does

eandre (approved):
Nice!

PR #298, opened 2022-07-01
runtime/pubsub: Add multi project support

ekerfelt:
Configure serverId on both topic and sub to be able to use separate projects for each


runtime/pubsub/internal/gcp/clients.go
```diff
@@ -5,39 +5,29 @@ import (
...
 var (
-	clients     = make(map[string]*pubsub.Client)
+	client      *pubsub.Client
 	clientMutex sync.Mutex
 )
 
 // getClient returns a singleton pubsub client for the given project or panics if it cannot be created.
-func getClient(project *config.GCPPubSubServer) *pubsub.Client {
+func getClient() *pubsub.Client {
 	clientMutex.Lock()
```
eandre, line 19: This would be a bit nicer with a `sync.Once`

runtime/runtime/config/config.go
```diff
@@ -162,9 +162,10 @@ type PubsubTopic struct {
 }
 
 type PubsubSubscription struct {
+	ServerID   int    `json:"server_id"`   // the index into (*Runtime).PubsubServers
```
eandre, line 165: Does this indirection really make sense now? Or would it make more sense to just put the project id and service account in here? (Happy to keep it as-is if we don't want to make changes at this point)

runtime/pubsub/internal/gcp/topic.go
```diff
@@ -56,8 +56,9 @@ func (t *topic) Subscribe(logger *zerolog.Logger, _ *types.SubscriptionConfig, s
 
 	// If we have a resource ID, let's register a push endpoint for it
 	if subscriptionCfg.ResourceID != "" {
-		if t.serverCfg.PushServiceAccount != "" {
-			registerPushEndpoint(t.serverCfg, subscriptionCfg, f)
+		serverCfg := t.serverCfg[subscriptionCfg.ServerID].GCP
```
DomBlack, line 59: Can we put an explicit check to see if `t.serverCfg[subscriptionCfg.ServerID]` exists or not.

Right now  this will panic with a nil pointer if we get this wrong - where as it would be nicer to panic with a error we'd understand

runtime/pubsub/topic.go
```diff
@@ -46,7 +46,7 @@ func NewTopic[T any](name string, cfg *TopicConfig) *Topic[T] {
 	case server.NSQServer != nil:
 		return &Topic[T]{topicCfg: topic, topic: nsq.NewTopic(server.NSQServer, topic)}
 	case server.GCP != nil:
-		return &Topic[T]{topicCfg: topic, topic: gcp.NewTopic(server.GCP, topic)}
+		return &Topic[T]{topicCfg: topic, topic: gcp.NewTopic(config.Cfg.Runtime.PubsubServers, topic)}
```
DomBlack, line 49: Given the `config.Cfg.Runtime.PubsubServers` is global, we shouldn't pass this in as a parameter,

eandre (approved):

runtime/pubsub/internal/gcp/topic.go
eandre: I don't think we should validate the internal consistency of our configuration in business logic, but rather in `config.ParseRuntime` where we do some lightweight validation already today.

PR #302, opened 2022-07-04
Refactor PubSub Subscription API design

DomBlack:
This commit refactors the the PubSub subscription API to make
it a package level function, which aligns more with the concept
of resource creation.

This refactor required modifying the parser to introduced phased based
passes into the resource parser, such that Topics get parsed before
subscriptions as subscriptions need to reference the already parsed topics.
Without this passes approach, a subscription in `a.go` might be parsed before
the topic it references in `b.go`.

In a future PR more information will be validated on phase of topics and subscriptions
- such as forcing the `DeliveryGarauntees` option to be explicility set.

This commit also fixes the `publicapigen` to work for the PubSub package.


compiler/internal/codegen/codegen_main.go
```diff
@@ -131,7 +131,7 @@ func (b *Builder) Main() (f *File, err error) {
 			subscriptions := Dict{}
 
 			for _, sub := range topic.Subscribers {
-				traceID := int(b.res.Nodes[sub.DeclFile.Pkg][sub.CallSite].Id)
+				traceID := int(b.res.Nodes[sub.DeclFile.Pkg][sub.IdentAST].Id)
```
DomBlack, line 134: Due to now treating subscriptions as a resource, rather than resource usage - this changed what node we register the reference from the call site to the ident that the subscriber is bound to.

parser/est/est.go
```diff
@@ -115,13 +115,21 @@ const (
...
+	FuncFile *File        // The file the subscriber function is declared in
+	DeclFile *File        // The file that the subscriber is defined in
+	IdentAST *ast.Ident   // The AST node representing the value this topic is bound against
 }
 
+func (p *PubSubSubscriber) Type() ResourceType         { return PubSubTopicResource }
+func (p *PubSubSubscriber) File() *File                { return p.DeclFile }
+func (p *PubSubSubscriber) Ident() *ast.Ident          { return p.IdentAST }
+func (p *PubSubSubscriber) NodeType() NodeType         { return PubSubSubscriberNode }
+func (p *PubSubSubscriber) AllowOnlyParsedUsage() bool { return true }
```
DomBlack, line 131: We now have to implement the Resource interface, which requires an ident to be tracked.

parser/feature_pubsub.go
```diff
@@ -30,20 +29,32 @@ func init() {
...
 		locations.AllowedIn(locations.Function).ButNotIn(locations.InitFunction),
 	)
 
+	registerResource(
+		est.PubSubSubscriptionResource,
+		"pubsub subscription",
+		"https://encore.dev/docs/develop/pubsub",
+		"pubsub",
+		"encore.dev/pubsub",
+		est.PubSubTopicResource,
```
DomBlack, line 45: this last field means that the Subscription Resource is dependant on the Topic resources being parsed first

parser/feature_pubsub.go
```diff
@@ -160,40 +172,35 @@ func (p *parser) parsePubSubSubscription(file *est.File, resource est.Resource,
...
+		IdentAST: ident,
 	}
 	topic.Subscribers = append(topic.Subscribers, subscription)
-
-	// Record the reference to the topic declaration
-	funcFile.References[subscription.CallSite] = &est.Node{
-		Type:       est.PubSubSubscriberNode,
-		Topic:      topic,
-		Subscriber: subscription,
-	}
```
DomBlack, line 196: Resources have their references tracked automatically

runtime/pubsub/types.go
```diff
@@ -1,38 +1,17 @@
 package pubsub
 
 import (
-	"context"
-
-	internal "encore.dev/pubsub/internal/types"
+	"encore.dev/pubsub/internal/types"
 )
 
-// SubscriptionConfig is used when creating a subscription
```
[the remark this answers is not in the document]
eandre: Maybe would be nicer if `publicapigen` uses this documentation instead? I could imagine we want the `internal/types` package to contain rather internal documentation than external-facing documentation.

parser/testdata/pubsub_err_subscriber_not_function.txt
```diff
@@ -18,7 +18,7 @@ type MessageType struct {
 
 var (
     BasicTopic = pubsub.NewTopic[*MessageType]("basic-topic")
-    _ = BasicTopic.NewSubscription("basic-subscription", shared.Subscriber)
+    _ = pubsub.NewSubscription(BasicTopic, "basic-subscription", shared.Subscriber)
```
eandre, line 21: This is out of date now I believe? We agreed on making the config required and moving the func to the config as `Handler`, I believe

runtime/pubsub/types.go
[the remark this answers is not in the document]
DomBlack: I wanted them copied from the internal docs, because we'd keep the docs more up to date in the place the thing is actually defined.

That said - I've just made it so we only copy the doc's over if doc's are missing from the target location (i.e. we've not written explicit docs in the public location)

PR #312, opened 2022-07-17
Add a how-to guide for using ent and atlas

Minivera:
This is based on the discussion here: https://github.com/encoredev/encore/discussions/307, and my own usage of ent within my application. This is fairly early and there might be a better way of doing this with features of Encore I don't know yet.

It's very easy to integrate a heavy ORM like ent with Encore if you're okay with tinkering a bit with scripts and commands. This guide shows one way of doing it, which could be expanded, but this seems to work really well on my end. We could also add a section for using only Atlas if it could be useful.

Any feedback is very much appreciated!

### DIFF
--- a/docs/docs.cue
+++ b/docs/docs.cue
@@ -117,6 +117,7 @@ sections: [
 			{title: "Use Firebase Authentication", segment: "firebase-auth"},
 			{title: "Integrate with GitHub", segment: "github"},
 			{title: "Migrate away from Encore", segment: "migrate-away"},
+			{title: "Use the ent ORM for migrations", segment: "entgo-orm"},
 		]
 	},
 	{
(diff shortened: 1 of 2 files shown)


eandre:
This looks great, thank you @Minivera!

Minivera:
@eandre just letting you know that I made some quick changes to some of the code to make sure it really works, and so the build will need to be manually approved again. It should be good on my end now 🙂

eandre:
Thanks @Minivera! For this to show up in the Encore website docs it needs to be added to the `docs.cue` file, at the end of the `how-to` section. If you can add that then I think this is ready to merge.

Minivera:
Ha! I hadn't realized I needed to do that, my bad. Should be good now.

eandre (approved):
Thank you!

PR #321, opened 2022-08-02
Add search functionality to endpoint dropdown

simon-johansson:
First commit 🎉🍾

## What
* Convert the endpoint dropdown into a Combobox with fuzzy search possibility
* Bonus: Add prettier(https://prettier.io/) for automatic code formatting. This also allows for sorting the tailwind css classes so that they automagically gets displayed in the same order everywhere. 

## Why
This was something that Skrym requested as it can be cumbersome to find the endpoint you are looking for if you have a lot of them.

## How
* Use Combobox component from HeadlessUI (https://headlessui.com/react/combobox#combobox)
* Add fuzzysort (https://www.npmjs.com/package/fuzzysort) npm dependency

## What is left
* Have not formatted all files according to the prettier rules.
* Would like to add basic unit tests to the Combobox component, but we need to add a unit testing library in order to do that.

## Screenshot
fuzzy search


eandre (dismissed):

cli/daemon/dash/dashapp/src/components/Combobox.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [wrapperRef]);
+
+  const filteredItems = fuzzysort.go(query, items, {
+    key: "name",
+    all: true,
+    threshold: -200,
+  });
+
+  return (
+    <HeadlessCombobox
```
DomBlack, line 43: Does the HeadlessCombobox handle aria labels etc? We use them on the webapp, would be nice to start using them everywhere to improve the accessibility

DomBlack (commented):
🎉 Awesome first PR!

cli/daemon/dash/dashapp/src/components/Combobox.tsx
simon-johansson: Thanks! 
Yes, they add some aria stuff:
image

They also add the functionality to select via keyboard and tab so it's not mouse/trackpad dependent. 
https://headlessui.com/react/combobox#accessibility-notes

DomBlack: Awesome! 🙏

DomBlack (approved):

eandre (approved):

PR #332, opened 2022-08-09
Fix top offset for trace tooltip

simon-johansson:
## What
Fix bug were the trace tooltip does not follows along with the scrolling

## Screenshot
**Before** - Tooltip does not follow along with the scroll
before

**After** - Tooltip is always inline with the trace beam
after

### DIFF
--- a/cli/daemon/dash/dashapp/src/components/trace/SpanDetail.tsx
+++ b/cli/daemon/dash/dashapp/src/components/trace/SpanDetail.tsx
@@ -271,7 +271,9 @@ const GoroutineDetail: FunctionComponent<{g: gdata, req: Request, trace: Trace, 
     setBarOver(true)
     setHoverObj(obj)
     const spanEl = (ev.target as HTMLElement)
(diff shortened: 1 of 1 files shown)


cli/daemon/dash/dashapp/src/components/trace/SpanDetail.tsx
```diff
@@ -271,7 +271,9 @@ const GoroutineDetail: FunctionComponent<{g: gdata, req: Request, trace: Trace,
     setBarOver(true)
     setHoverObj(obj)
     const spanEl = (ev.target as HTMLElement)
-    el.style.marginTop = `calc(${spanEl.offsetTop}px - 40px)`;
+    const offset = spanEl.getBoundingClientRect();
```
simon-johansson, line 274: `getBoundingClientRect` gives you offset relative to your client viewport, `offsetTop` did never change in this case.

simon-johansson:
We do not have this problem in the admin dashboard

PR #333, opened 2022-08-09
Add front-end unit test possibilities

simon-johansson:
## What
Adding `Jest` and `react-testing-library` as dependencies to the project (the recommended unit testing libs for React). Also needed to add some `babel` dependencies which allow us to use ES6 Modules, JSX, and TypeScript in our tests.

## Why
Really want to have some unit tests for the front-end. Not having anything is bad

## What's next
We should run the FE unit tests during CI for pull requests. What do you think, should we add this to the existing GitHub action that we have or should we create a new one for testing & linting the front-end code?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,17 @@ npm install
 npm run build
 ```
 
+### Testing the Development Dashboard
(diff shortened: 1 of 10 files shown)


eandre (approved):
Awesome, this is much needed! Thanks for doing this!

PR #341, opened 2022-08-11
Add `config.Load` parsing capability

DomBlack:
This commit adds the ability for Encore to parse a call to `config.Load` and register the type requested in the config load call.

To achieve this we've added a new type into the schema to represent any wrapped types using the `config.Value` wrapper.


parser/parser.go
```diff
@@ -581,6 +598,18 @@ func (p *parser) validateApp() {
...
 
+func (p *parser) validateTypeDoesntUseConfigTypes(pos token.Pos, param *est.Param) {
+	err := schema.Walk(p.decls, param.Type, func(n any) error {
+		if _, ok := n.(*schema.ConfigValue); ok {
+			return errors.New("config.Value")
+		}
+		return nil
+	})
+	if err != nil {
+		p.errf(pos, "type %s can only be used in data types used by config.Load and cannot be used for API response/response parameters or PubSub message types", err.Error())
```
eandre, line 609: Maybe accept `API response/response parameters or PubSub message` as a `context string` argument so we can be more specific in the error and not have to list all possible cases where it's not allowed

runtime/config/types.go
```diff
@@ -0,0 +1,37 @@
...
+	"encore.dev/types/uuid"
+)
+
+// Value represents a value in the configuration for this application
+// which can be any value represented in the configuration files.
+//
+// It is a function because the underlying value could change while
+// the application is still running due to unit tests providing
+// overrides to test different behaviours.
+type Value[T any] func() T
```
eandre, line 15: I'm still leaning towards `Dynamic[T]` (or maybe `Dyn[T]` to save on typing as it will be quite commonly used) but don't need to block on that to merge to a feature branch obviously

eandre (approved):

PR #346, opened 2022-08-22
runtime/appruntime/api: fix panic on request decode error

eandre:
If the API receives a request that cannot be decoded it
would sometimes cause a runtime panic. Fix this, and
improve test coverage of this area while we're at it.

Thanks @melkstam and Max D for the bug report

### DIFF
--- a/runtime/appruntime/api/handler.go
+++ b/runtime/appruntime/api/handler.go
@@ -122,7 +122,16 @@ func (d *Desc[Req, Resp]) begin(c IncomingContext) (reqData Req, beginErr error)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ekerfelt (commented):
Change looks fine, but not sure I follow the test changes. I guess you are extending the Unauthenticated test to a complete end-to-end test? Why not add an additional test instead?

eandre:
Ah good catch, thanks. I was merging the two tests into a more general table-driven test but forgot to migrate the old test over. PTAL again

ekerfelt (approved):

PR #357, opened 2022-09-02
Compute concrete CUE values for services.

DomBlack:
This commits adds to the compiler a step to compute a concrete value
for each service which uses config. If a services configuration is not
concrete, then a compile error is given to the user.

The compiler first reads all configuration files from the application root
into a virtual file system (package `vfs`) which implements the `io/fs.FS`
interface. This pickup is done by finding all files with the `.CUE` extension.

The result could be the following file list:
```
svc/encore.gen.cue
svc/foo.cue
baz/encore.gen.cue
another/encore.gen.cue
another/config.cue
```

During the `checkApp` phase, the compiler then asks the CUE libraries to compute
a concrete value for each service. So for example the `another` service would
have a single concrete `cue.Value` created from the unification of:

```
another/encore.gen.cue
another/config.cue
another/config/currencies.cue
another/config/users.cue
```


compiler/internal/cueutil/build.go
```diff
@@ -0,0 +1,268 @@
...
+		return cue.Value{}, eerror.Wrap(err, "config", "unable to list all config files for service", map[string]any{"path": serviceRelPath})
+	}
+
+	// Tell CUE to load all the files
+	loaderCfg := &load.Config{
+		Dir:   tmpPath,
+		Tools: true,
+	}
+	pkgs := load.Instances(configFilesForService, loaderCfg)
+	for _, pkg := range pkgs {
```
eandre, line 52: Should we support multiple packages in a given service? Why? What does it mean?

compiler/internal/cueutil/build.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return cue.Value{}, toDescriptiveError(err)
+	}
+	if len(values) == 0 {
+		return cue.Value{}, eerror.New("config", "no values generated from config", nil)
+	}
+
+	// Unify all returned values into a single value
+	rtnValue := values[0]
+	for _, value := range values {
```
eandre, line 75: When do we have multiple values? Is it one per package? (See above re: multiple packages)

compiler/internal/cueutil/build.go
```diff
@@ -0,0 +1,268 @@
...
+	if err == nil {
+		return nil
+	}
+
+	errs := cueerrors.Errors(err)
+	if len(errs) == 1 {
+		return errs[0]
+	}
+
+	var str strings.Builder
```
eandre, line 126: We have the `pkg/errlist` package that it would be nice to reuse here instead of constructing a multiline error string.

compiler/internal/cueutil/build.go
```diff
@@ -0,0 +1,268 @@
...
+		} else {
+			if err := os.Mkdir(filepath.Join(tmpPath, path), 0755); err != nil && !errors.Is(err, os.ErrExist) {
+				return err
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		_ = os.RemoveAll(tmpPath) // clear the temporary directory incase of error
```
eandre, line 179: See comment earlier in the PR: better to move the temp dir creation up a level and remove this from here

pkg/vfs/file.go
```diff
@@ -0,0 +1,111 @@
...
+	_ fs.File     = (*File)(nil)
+	_ fs.FileInfo = (*File)(nil)
+	_ fs.DirEntry = (*File)(nil)
+)
+
+func (f *File) Name() string {
+	return f.node.name
+}
+
+// Reader is the interface that wraps the basic Read method.
```
eandre, line 30: Copying the doc string from `io.Reader` is a bit excessive :) I think just `Read implements io.Reader` would do.

pkg/vfs/utils_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"io/fs"
+	"path/filepath"
+	"testing"
+	"testing/fstest"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestFromDir(t *testing.T) {
```
eandre, line 12: Nit: I would expect this test in `vfs_test.go` instead of `utils_test.go` as it is really testing the whole implementation with `fstest.TestFS`

pkg/vfs/utils.go
```diff
@@ -0,0 +1,43 @@
+package vfs
+
+import (
+	"io/fs"
+	"os"
+
+	"encr.dev/pkg/eerror"
+)
+
+func FromDir(workingDir string, predicate func(fileName string) bool) (*VFS, error) {
```
eandre, line 10: Might be useful to document the predicate only gets called for files and not directories? Is there any reason why it doesn't accept `(path string, info fs.DirEntry)`?

pkg/vfs/utils.go
```diff
@@ -0,0 +1,43 @@
...
+	dirFS := os.DirFS(workingDir)
+	rtn := New()
+
+	if predicate == nil {
+		predicate = func(fileName string) bool { return true }
+	}
+
+	err := fs.WalkDir(dirFS, ".", func(path string, info fs.DirEntry, err error) error {
```
eandre, line 18: This doesn't appear to check the `err` value so if something fails it will silently drop files from the resulting FS

eandre (commented):
This is really great work. I really appreciate the degree to which you're documenting all the functions and code sections.

Adding some comments that I think should be improved, but overall it's shaping up really nicely. Thank you!

compiler/build.go
```diff
@@ -184,9 +193,20 @@ func (b *builder) Build() (res *Result, err error) {
 	}
 
 	res.Parse = b.res
+	res.AppConfigFiles = b.configFiles
 	return res, nil
 }
 
+func (b *builder) startAppCheck() error {
+	b.lastOpID = b.cfg.OpTracker.Add("Verifying application configuration", time.Now())
```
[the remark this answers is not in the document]
DomBlack: I've refactored, so all three value sets to this variable, now track their own OpID as well as a separate variable.

`lastOpID` is used for the recovery paths to stop the tracker correctly, so we still need a single reference point for what is the "last Op ID" started

compiler/config.go
```diff
@@ -0,0 +1,70 @@
...
+func (b *builder) pickupConfigFiles() error {
+	// Create a virtual filesystem for the config files
+	configFiles, err := vfs.FromDir(b.appRoot, func(path string) bool {
+		// any CUE files
+		if filepath.Ext(path) == ".cue" {
+			return true
+		}
+
+		// Pickup any files within a CUE module folder (either at the root of the app or in a subfolder)
+		if strings.Contains(path, "/cue.mod/") || strings.HasPrefix(path, "cue.mod/") {
```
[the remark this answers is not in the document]
DomBlack: The ability to import config from other packages (like a shared library package).

Even though Encore itself doesn't use `cue.mod` yet, users _might_ and we need to pick up the files otherwise stuff which works locally might not work when deployed to the cloud.

compiler/config.go
```diff
@@ -0,0 +1,70 @@
...
+			return true
+		}
+
+		// Pickup any files called "config.ext" where the file type is supported by CUE
```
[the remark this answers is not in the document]
DomBlack: I disagree, once of the reasons we picked CUE was because of it's interop with config languages that people are maybe more familiar with.

compiler/internal/cueutil/build.go
DomBlack: `errlist` works with filesets and is based on `scanner.ErrorList` / `scanner.Error` where each error has a token position - we don't have that because these errors are not Go parse errors.

eandre (approved):

PR #369, opened 2022-09-16
Redis cache support

eandre:
This adds support for Redis Caching as per [the proposal](https://github.com/encoredev/encore/discussions/355).


runtime/storage/cache/basic.go
```diff
@@ -0,0 +1,563 @@
...
+}
+
+// Get gets the value stored at key.
+// If the key does not exist, it returns an error matching Miss.
+func (s *StringKeyspace[K]) Get(ctx context.Context, key K) (string, error) {
+	return s.basicKeyspace.Get(ctx, key)
+}
+
+// Set updates the value stored at key to val.
+// If the key does not exist, it returns an error matching Miss.
```
ekerfelt, line 39: is this true? I assumes this would add a key/val

runtime/storage/cache/basic.go
```diff
@@ -0,0 +1,563 @@
...
+// See https://redis.io/commands/incrbyfloat/ for more information.
+func (s *FloatKeyspace[K]) Decr(ctx context.Context, key K, delta float64) (newVal float64, err error) {
+	const op = "decr"
+	k, err := s.key(key, op)
+	if err != nil {
+		return 0, err
+	}
+
+	res, err := do[K, float64, *redis.FloatCmd](s.client, ctx, k, func(c cmdable) *redis.FloatCmd {
+		return c.IncrByFloat(ctx, k, -delta)
```
ekerfelt, line 403: doesn't this support decr?

runtime/storage/cache/cache.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	pipe := cl.redis.TxPipeline()
+	res := fn(pipe)
+	_ = pipe.Process(ctx, exp)
+	_, _ = pipe.Exec(ctx)
+	return res
+}
+
+func do2[K, V, Res any](
```
ekerfelt, line 219: maybe a small comment :)

runtime/storage/cache/basic.go
eandre: No, there is no dedicated `DecrByFloat` in Redis

runtime/storage/cache/basic.go
eandre: Whoops :) good catch!

parser/feature_cache.go
```diff
@@ -0,0 +1,396 @@
...
+				p.errf(ks.Ident().Pos(), "cache.%s key type must be a basic type or a named struct type",
+					con.FuncName)
+				return
+			}
+
+			// Validate struct fields
+			for _, f := range st.Fields {
+				fieldName := f.Name
+				if _, exists := seenPathSegments[fieldName]; !exists {
+					p.errf(ks.Ident().Pos(), "invalid cache value type %s: field %s not used in KeyPattern",
```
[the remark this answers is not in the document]
eandre: I was thinking yes, similar to Go's unused variable being an error. It seems bug-prone to not use it in the key pattern. What do you think?

parser/feature_cache.go
```diff
@@ -0,0 +1,396 @@
...
+			est.CacheKeyspaceResource,
+			constructor.FuncName, numTypeArgs,
+			createKeyspaceParser(constructor),
+			locations.AllowedIn(locations.Variable).ButNotIn(locations.Function),
+		)
+	}
+}
+
+func (p *parser) parseCacheCluster(file *est.File, cursor *walker.Cursor, ident *ast.Ident, callExpr *ast.CallExpr) est.Resource {
+	if len(callExpr.Args) != 2 {
```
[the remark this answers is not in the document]
eandre: Yes, but this runs before the compiler

parser/testdata/cache_keyspace_invalid.txt
```diff
@@ -0,0 +1,26 @@
...
+
+var cluster = cache.NewCluster("cluster", cache.ClusterConfig{})
+
+type Key struct {
+    ID float64
+    Bar float64
+}
+
+var keyspace = cache.NewStringKeyspace[Key](cluster, cache.KeyspaceConfig{
+    Path: "/foo/:ID",
```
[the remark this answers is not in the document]
eandre: Yes

cli/daemon/redis/redis.go
```diff
@@ -0,0 +1,92 @@
...
+		acc += tickInterval
+		if acc > cleanupInterval {
+			acc -= cleanupInterval
+			s.clearKeys()
+		}
+	}
+}
+
+// clearKeys clears random keys to get the redis server
+// down to 100 persisted keys, as a simple way to bound
```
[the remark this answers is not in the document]
ekerfelt: Never mind. Given that it's a volatile cache even a tiny number is fine here

parser/testdata/cache_keyspace_invalid.txt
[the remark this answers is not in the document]
ekerfelt: ahh, because it would need to use Bar too?

runtime/storage/cache/basic.go
ekerfelt: 👍 curious that they've added support for int decr

cli/daemon/engine/trace/trace.go
```diff
@@ -868,6 +874,7 @@ func (tp *traceParser) publishEnd(ts uint64) error {
 	publish.EndTime = ts
 	publish.MessageId = tp.String()
 	publish.Err = tp.ByteString()
+	delete(tp.publishMap, publishID)
```
ekerfelt, line 877: are we really deleting this here? Don't we lose the reference to `publish`?

ekerfelt (dismissed):

runtime/storage/cache/basic.go
```diff
@@ -0,0 +1,636 @@
...
+// and returns the new value.
+//
+// If the key does not exist it is first created with a value of 0
+// before incrementing.
+//
+// Negative values can be used to decrease the value,
+// but typically you want to use the Decr method for that.
+//
+// See https://redis.io/commands/incrby/ for more information.
+func (s *IntKeyspace[K]) Incr(ctx context.Context, key K, delta int64) (newVal int64, err error) {
```
DomBlack, line 287: I really don't like `Incr` as a method name.

Either we should copy the underlying API name directly `IncrBy` to make it easier for people used to using Redis; or we should use a more standard shortened  version (`Inc` / `Dec`). I think I'd prefer if we used `Increment` and `Decrement` though... Based on one of the ACR's we made a while back;

> // Decision; we use the suffix `Config` instead of an acronym like `Cfg` or `Args` to be friendly to non native English speakers

runtime/storage/cache/cache.go
```diff
@@ -0,0 +1,267 @@
...
+// An EvictionPolicy describes how the cache evicts keys to make room for new data
+// when the maximum memory limit is reached.
+//
+// See https://redis.io/docs/manual/eviction/#eviction-policies for more information.
+type EvictionPolicy string
+
+// NOTE: These values need to be added to the runtimeconstants package
+// and the parser package for the parser to be aware of them.
+
+const (
```
DomBlack, line 34: I think again I'd prefer `LeastRecentlyUsed`, `LeastFrequentlyUsed` and `Random` here. it's easier to understand what the behaviour is without having to read the documentation of the contants.

DomBlack, line 34: I'm not sure I understand fully the difference between `AllKeysX` and `VolatileX`. I've read the doc's and I think by setting Volatile it means that only keys with an expiration will be deleted and keys without an expiration will never be deleted.

> if there are no keys to evict with an expiration set.
This bit then confuses me, I think the wording would be better:

> If no keys within the cache have an expiration set, then this setting behaves like `NoEviction`.

DomBlack, line 34: I also don't like how it feels like we've squashed `AllKeys` / `Volatile` with `LRU` / `LFU` / `Random` / `NoEviction`.

These feel like two different settings - where the documentation for each setting can be read in isolation. The only special edge case is simply if `NoEvicition` is set it doesn't matter about `Volitle` vs `AllKeys`.

In this case then `Volatile` could be called `KeysWithExpiry`. So the option becomes

```go
cache.New(..., cache.Options{
  EvictionPolicy: cache.EvicationPolicy {
    Type: cache.Randomly, // cache.LeastFrequentlyUsed / cache.LeastRecentlyUsed / cache.NoEviction
    What:  cache.AllKeys // cache.KeysWithExpiries
  }
})
```

runtime/storage/cache/cache.go
```diff
@@ -0,0 +1,267 @@
...
+	// It behaves like NoEviction if there are no keys to evict with an expiration set.
+	VolatileRandom EvictionPolicy = "volatile-random"
+
+	// NoEviction does not evict any keys, and instead returns an error to the client
+	// when the max memory limit is reached.
+	NoEviction EvictionPolicy = "noeviction"
+)
+
+//publicapigen:keep
+type constStr string
```
DomBlack, line 67: Can we put a doc on here explaining what it is for people. It seems like a read type initially if you're not used to seeing this

runtime/storage/cache/cache.go
```diff
@@ -0,0 +1,267 @@
...
+	// to a string. It's of type any to avoid making KeyspaceConfig
+	// a generic type. It's an internal field set by Encore's compiler.
+	//
+	// The type must be func(K) string.
+	//publicapigen:drop
+	EncoreInternal_KeyMapper any
+}
+
+// An OpError describes the operation that failed.
+type OpError struct {
```
DomBlack, line 112: Why isn't this an `errs` type? It feels very weird that we have our own error type, and yet we have here another type of error again.

runtime/storage/cache/cache.go
```diff
@@ -0,0 +1,267 @@
...
+// KeyExists is the error reported when a key already exists
+// and the requested operation is specified to only apply to
+// keys that do not already exist.
+// It must be checked against with errors.Is.
+var KeyExists = errors.New("key already exists")
+
+// An WriteOption customizes the behavior of a single cache write operation.
+type WriteOption interface {
+	//publicapigen:keep
+	writeOption() // ensure only our package can implement
```
DomBlack, line 139: I believe you can still embed this interface and calling writeOption will result in an nil pointer (or at least this is what you told me when I did this in a previous PR)

So this doesn't prevent other packages from implementing

runtime/storage/cache/manager_internal.go
```diff
@@ -0,0 +1,431 @@
...
+			}
+			mgr.clients[clusterName] = cl
+			return cl
+		}
+	}
+
+	panic(fmt.Sprintf("cache: unknown cluster %q", clusterName))
+}
+
+func (mgr *Manager) runningInEncoreCloud() bool {
```
DomBlack, line 93: This feels like it should be controlled by config, rather than being hard coded like this.

runtime/storage/cache/manager_internal.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func (mgr *Manager) newMiniredisClient() (*redis.Client, error) {
+	err := mgr.initTestSrv.Do(func() error {
+		var err error
+		mgr.testSrv, err = miniredis.Run()
+
+		// Periodically clean up cache keys if running in Encore Cloud.
+		if err == nil && mgr.runningInEncoreCloud() {
+			go miniredisCleanup(mgr.testSrv, 15*time.Second, 100)
```
DomBlack, line 143: If everything scales back to zero after the request, will this actually run on Encore Cloud?

runtime/storage/cache/pkgfn.go
```diff
@@ -0,0 +1,15 @@
+//go:build encore_app
+
+package cache
+
+//publicapigen:drop
+var Singleton *Manager
+
+// NewCluster declares a new cache cluster.
```
DomBlack, line 8: I think this comment should link off to our docs

runtime/storage/cache/basic.go
eandre: It is a Redis command: https://redis.io/commands/incr/ — but while Redis has both `Incr` and `IncrBy` we combined both into one.

DomBlack: Oh I get that, but given our decision about avoiding acronyms what do you think about Increment and Decrement or Increase / Decrease?

eandre: Yeah that's fair. We can go with Increment/Decrement.

runtime/storage/cache/cache.go
eandre: Yeah I hear you, but it's a bit of a mouthful: `AllKeysLeastFrequentlyUsed`.

I don't think the idea of splitting it into two fields is right as the underlying configuration is a single concept (the eviction policy) and not two orthogonal concepts.

It's not clear future policies will map cleanly to those two dimensions you described (and NoEviction arguably doesn't already today).

eandre: That's correct. We can try to improve the docs

DomBlack: > it's a bit of a mouthful: AllKeysLeastFrequentlyUsed.

sure, but with auto complete in IDE's it's not actually more keys - what it is though is more readable. With a good fuzzy autocomplete typing AllLFU will almost certainly auto complete the same thing

runtime/storage/cache/cache.go
eandre: Yes, that's correct, Ibut it's purely meant as a signal to not create your own implementations so that's fine

runtime/storage/cache/cache.go
eandre: They serve very different use cases. `encore.dev/beta/errs` is for application-returned errors and API responses. This type is for caching-specific information like the key(s) and cache operation. Much like the standard library has lots of different error types.

I don't see why we necessarily need to have exactly one error type given that Encore supports lots of different use cases. (We should probably have a separate error type for `sqldb` errors, too.)

parser/literals.go
```diff
@@ -188,7 +188,7 @@ func (p *parser) parseConstantValue(file *est.File, value ast.Expr) (rtn constan
 				}
 			}
 		}
-		p.errf(value.Pos(), "You can not call a function here, only constant values are supported")
```
DomBlack, line 191: Did you re-add this error?

cli/daemon/run/run.go
```diff
@@ -554,6 +557,26 @@ func (r *Run) generateConfig(p *Proc, params *startProcParams) *config.Runtime {
...
+			Host: params.Redis.Addr(),
+		}
+		redisServers = append(redisServers, srv)
+
+		for _, cluster := range params.Meta.CacheClusters {
+			redisDBs = append(redisDBs, &config.RedisDatabase{
+				ServerID:   0,
+				Database:   0,
+				EncoreName: cluster.Name,
+				KeyPrefix:  cluster.Name + "/",
```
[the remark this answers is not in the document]
eandre: We create a new `ResourceServices` for every app when invoking `encore run`, so they're already isolated

DomBlack (dismissed):

DomBlack (approved):

PR #389, opened 2022-09-26
runtime: bump versions

eandre:
The GCP libraries have introduced some changes that need
to be upgraded simultaneously with the GCP protobuf
library.

This also removes the old WireGuard tunnel as the wireguard-go library has changed
in a bunch of backwards-incompatible ways, and since it's not currently being used
we can reintroduce it when we are ready to use it again.


DomBlack (dismissed):

DomBlack (approved):
Approved, but could you update the PR to say you've also removed the VPN tunnel - that was a little unexpected

PR #390, opened 2022-09-26
Initial support for reading Config at runtime 

DomBlack:
This PR adds initial support for generating the Go code needed to
unmarshal configuration at runtime. As well as adding a new end to end
test verifying the config is being loaded correctly into a running app.

At the suggestion of André, I've also added a new `config.Values[T]` type
which abstracts over a slice of values `T`.


compiler/rewrite.go
```diff
@@ -105,6 +105,16 @@ func (b *builder) rewritePkg(pkg *est.Package, targetDir string) error {
...
+				res := rewrite.Res.(*est.Config)
+
+				var buf bytes.Buffer
+				buf.WriteString(strconv.Quote(pkg.Service.Name))
+				buf.WriteString(", EncoreInternal_UnmarshalConfig_")
+				buf.WriteString(b.res.Meta.Decls[res.ConfigStruct.Type.GetNamed().Id].Name)
+				ep := fset.Position(res.FuncCall.Rparen)
+				_, _ = fmt.Fprintf(&buf, "/*line :%d:%d*/", ep.Line, ep.Column)
+				rw.Replace(res.FuncCall.Lparen+1, res.FuncCall.Rparen, buf.Bytes())
+
```
DomBlack, line 117: This effectiley rewrites:

```go
config.Load[Foo]()
```

to

```go
config.Load[Foo]("serviceName", EncoreInternal_UnmarshalConfig_Foo)
```

cli/daemon/run/testdata/echo/echo/config.go
```diff
@@ -0,0 +1,40 @@
...
+	SubConfig config.Value[struct {
+		SubKey SubCfgType
+	}]
+}
+
+type SubCfgType struct {
+	MaxCount uint
+}
+
+var cfg = config.Load[CfgType]()
```
ekerfelt, line 23: How does this magic work? Why isn't this causing a compile error in the IDE?

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
```diff
@@ -1505,10 +1517,145 @@ func DI(ctx context.Context) error {
...
+
+							if decoder.LastError != nil {
+								panic(fmt.Sprintf("unable to decode the config: %v", decoder.LastError))
+							}
+
+							return
+						}()
+					case "Foo":
+						obj.Foo = func() (rtn []byte) {
+							decoder := &etype.Marshaller{}
```
ekerfelt, line 1622: why func literal?

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
```diff
@@ -1505,10 +1517,145 @@ func DI(ctx context.Context) error {
...
+			})
+		case "FieldE":
+			obj.FieldE = func() (obj struct {
+				SubA string
+				SubB FooParams
+				Blah time.Time
+				Foo  []byte
+				Json json.RawMessage
+			}) {
+				itr.ReadObjectCB(func(itr *jsoniter.Iterator, field string) bool {
```
ekerfelt, line 1602: This gets a bit messy. Could we maybe extract a FieldE_Decoder func?

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
```diff
@@ -1505,10 +1517,145 @@ func DI(ctx context.Context) error {
...
+
+							if decoder.LastError != nil {
+								panic(fmt.Sprintf("unable to decode the config: %v", decoder.LastError))
+							}
+
+							return
+						}()
+					case "Json":
+						obj.Json = func() (rtn json.RawMessage) {
+							decoder := &etype.Marshaller{}
```
ekerfelt, line 1634: why not use a shared marshaller?

compiler/internal/codegen/testdata/TestCodeGenMain__token_auth.golden
```diff
@@ -246,6 +247,17 @@ var EncoreInternal_AuthHandlerAuthHandler = &api.AuthHandlerDesc[string]{
 		return [][]byte{v}, nil
 	},
 }
+
+
```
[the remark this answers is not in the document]
DomBlack: This is the golden tests only, real code won't write the unmarshallers if there are no calls to `config.Load`

That check is here; https://github.com/encoredev/encore/blob/333bbf478127d7bef4fd91e3caed431cef439e16/compiler/internal/cuegen/generator.go#L27-L30

cli/daemon/run/testdata/echo/echo/config.go
DomBlack: The public API will not have the two "hidden" arguments

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
DomBlack: Because error tracking needs to occur within the single read - and the `readType` code generator is not aware it's been called within another call to `readType` where a shared marshaller could be initialised.

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
DomBlack: It's not a named type, and I need to track the type arguments for generics, so for generic purposes this is actually cleaner (code no in this PR)

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
ekerfelt: ah, right. how about a call to a helper method? guess all variables can be parameterized and moved to a non-generated helper method?

compiler/internal/codegen/testdata/TestCodeGenMain__variants.golden
ekerfelt: Not part of this exercise, but I guess the decoder methods could easily be turned into static helpers. would reduce the generated code and overall complexity.

ekerfelt (approved):

PR #392, opened 2022-09-27
Print message when encore logs is waiting for logs

campesel:
`encore logs` doesn't print any message when started, so you can't tell whether it's waiting for logs. This PR adds the code to print a message to standard output when `encore logs` is started.

The message can be omitted by passing the `-q` flag to `encore logs`, which might be useful when using this command for scripting.

### DIFF
--- a/cli/cmd/encore/logs.go
+++ b/cli/cmd/encore/logs.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"os"
 	"os/signal"
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/logrusorgru/aurora/v3"
 	"github.com/rs/zerolog"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/encore/logs.go
```diff
@@ -92,6 +93,10 @@ func streamLogs(appRoot, envName string) {
 	// Use the same configuration as the runtime
 	zerolog.TimeFieldFormat = time.RFC3339Nano
 
+	if !logsQuiet {
+		os.Stdout.WriteString("Connected, waiting for logs...")
```
DomBlack, line 97: What do you think of maybe using `aurora` to gray scale this log line? (so it's not as visible as the logs when they do start coming in?)

See https://github.com/encoredev/encore/blob/main/cli/cmd/encore/run.go#L73 for an example

DomBlack (commented):
Looks great! One dicussion point maybe, what do you think?

cli/cmd/encore/logs.go
eandre: Alternatively we could make it a spinner that we delete when logs show up

DomBlack (approved):

cli/cmd/encore/logs.go
campesel: @eandre I went for a message to mirror what the web app does. Do you feel strongly about the spinner?

eandre: Don't feel strongly at all, ship it :)

PR #394, opened 2022-09-27
cli/daemon/run: track app compilation

### LINKED ISSUE #393: `encore run` does not print compile errors in all circumstances
Mrinal Wahal reported on [Slack](https://encore.dev/slack) an instance where `encore run` did not print the error when there was a compile-time error in the application. The error was discovered by running `encore daemon -f` in a separate terminal, which printed the error message. We should ensure compilation errors are always printed.

eandre:
This makes the compilation step tracked by Encore
to ensure compilation errors are always printed.

Fixes #393

### DIFF
--- a/cli/daemon/run/run.go
+++ b/cli/daemon/run/run.go
@@ -279,7 +279,7 @@ func (r *Run) buildAndStart(ctx context.Context, tracker *optracker.OpTracker) e
 	}
 
 	var build *compiler.Result
-	jobs.Go("Compiling application source code", false, 0, func(ctx context.Context) (err error) {
+	jobs.Go("Compiling Encore application", true, 0, func(ctx context.Context) (err error) {
 		cfg := &compiler.Config{
 			Revision:              parse.Meta.AppRevision,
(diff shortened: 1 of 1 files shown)


eandre:
@DomBlack if you can check that this makes sense that would be great

cli/daemon/run/run.go
```diff
@@ -279,7 +279,7 @@ func (r *Run) buildAndStart(ctx context.Context, tracker *optracker.OpTracker) e
 	}
 
 	var build *compiler.Result
-	jobs.Go("Compiling application source code", false, 0, func(ctx context.Context) (err error) {
+	jobs.Go("Compiling Encore application", true, 0, func(ctx context.Context) (err error) {
```
DomBlack, line 282: I don't this makes sense as the compiler.Build process within the job already fully tracks errors, this way we're going to end up with nested tracked jobs... maybe that's fine - but I'm not sure that's the root cause here.

I suspect it might be multiple rounds of panic/recovery hiding the error

PR #395, opened 2022-09-28
[ENC-801] Add generic support to config codegen.

DomBlack:
This commit adds support for config to be unmarshaled when generic types are used either at the top level config type or within the top level type.

For example:

```go
type Optional[T any] struct { Value: T }
type Foo struct { ReadOnly: Optional[bool] }

var (
    _ = config.Load[Optional[int]]()
    _ = config.Load[Foo]()
)
```

While implementing this I discovered a bug within our `GetConcreteType` code which did not support generic parameters being used within inline structs inside decls. Which has been fixed.

While debugging that, I also introduced a new set of helper functions in the `gocodegen` package which can return a string representing the formatted Go type of a `*schema.Type` which helps both debugging and also with generating unique names for concrete code referencing generic types.


compiler/internal/codegen/codegen_config_marshalers.go
```diff
@@ -0,0 +1,447 @@
+package codegen
```
DomBlack, line 1: I renamed this file to correct a typo in the filename (`marshallers` -> `marshalers`).

So I think diffing the two files would be best during this review - not sure why Github hasn't seen it's a rename

compiler/internal/codegen/codegen_config_marshalers.go
```diff
@@ -0,0 +1,447 @@
...
+			Return(unmarshaler),
+		), returnType
+
+	}
+}
+
+// typeUnmarshalerName returns a generated name for the unmarshaler function for the given type and the type that
+// the decl is.
+func (cb *configUnmarshalersBuilder) typeUnmarshalerName(id uint32) (reader string, rtnTyp *Statement) {
+	return fmt.Sprintf("encoreInternalTypeConfigUnmarshaler_%s", cb.md.Decls[id].Name), Id(cb.md.Decls[id].Name)
```
eandre, line 341: In the past we've defined these as `EncoreInternal_`. I think it's fine they're not exported, but might be good to have the underscore (`encoreInternal_`) to stay consistent

eandre (approved):
Nice!

PR #396, opened 2022-09-28
Default to primary environment when running encore logs

campesel:
Currently, `encore logs` returns an error when no environment is given. This PR makes `encore logs` default to the app primary environment.

### DIFF
--- a/cli/cmd/encore/logs.go
+++ b/cli/cmd/encore/logs.go
@@ -42,6 +42,9 @@ func streamLogs(appRoot, envName string) {
 		fatal("app is not linked with Encore Cloud")
 	}
 
+	if envName == "" {
+		envName = "@primary"
+	}
 	logs, err := platform.EnvLogs(ctx, appSlug, envName)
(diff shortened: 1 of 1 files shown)


cli/cmd/encore/logs.go
```diff
@@ -121,6 +121,6 @@ func streamLogs(appRoot, envName string) {
 
 func init() {
 	rootCmd.AddCommand(logsCmd)
-	logsCmd.Flags().StringVarP(&logsEnv, "env", "e", "", "Environment name to stream logs from (defaults to the production environment)")
+	logsCmd.Flags().StringVarP(&logsEnv, "env", "e", "@primary", "Environment name to stream logs from (defaults to the primary environment)")
```
eandre, line 124: This string is not meaningful to our users, so I think it would be better to keep the default as `""` here and convert it to `@primary` in code

DomBlack (approved):

eandre (approved):

PR #397, opened 2022-09-29
[ENC-795] Add support for pointers to parser

DomBlack:
This commit adds a new `Pointer` data type to the Encore schema representation of data types. This allows us to generate fully type awaire code for data structures where previously we've been able to work around not knowing where pointers where with reflection at runtime.

We needed this capability now as config unmarshalling uses generated code rather than runtime reflection.

The intended outcome of this commit is to only add pointers for config and to as much as possible leave all other generated code unchanged from the non-pointer awaire version.


cli/daemon/run/testdata/echo/echo/config.go
```diff
@@ -12,7 +12,7 @@ type CfgType[T any] struct {
 	AdminUsers   config.Values[string]
 
 	SubConfig config.Value[struct {
-		SubKey SubCfgType[T]
+		SubKey *SubCfgType[T]
```
DomBlack, line 15: This tests that pointers are processed OK by the config system

cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -184,7 +184,7 @@ export namespace echo {
 
     export interface EmptyData {
         OmitEmpty: Data<string, string>
-        NullPtr: string
+        NullPtr: (string | null)
```
DomBlack, line 187: This change in Typescript is more correct as we're now modelling that these things could be null technically if a `nil` is set within the Go data structure.

This is because we do not validate if `nil` is set or not automatically

cli/internal/codegen/testdata/expected_golang.go
```diff
@@ -76,28 +76,28 @@ func WithHTTPClient(client HTTPDoer) Option {
 	}
 }
 
-// WithAuthToken allows you to set an authentication token to be used for each request.
-//
-// This token will be sent as a Bearer token in the Authorization header.
-func WithAuthToken(bearerToken string) Option {
+// WithAuth allows you to set the authentication data to be used with each request
```
DomBlack, line 79: I swapped the Go client gen from a `token` auth to a data structure auth, so there's more changes in this file than you might expect. But this is fine

cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -66,7 +66,7 @@ export interface ClientOptions {
      * request either by passing in a static object or by passing in
      * a function which returns a new object for each request.
      */
-    auth?: echo.AuthParams | AuthDataGenerator
+    auth?: (echo.AuthParams | null) | AuthDataGenerator
```
eandre, line 69: This is a really complex type signature, combining optional `?`, parentheses, and `null` inside the parentheses. Is this correct? And if so, can we simplify it somehow? 

If `omitempty` is set `null` gets converted to `undefined` and it'd simplify to not being the parentheses or the `null`, is that implemented?

cli/daemon/run/testdata/echo_client/client.ts
```diff
@@ -207,16 +207,16 @@ export namespace echo {
...
+        Struct: Data<(Data<string, string> | null), number>
 
-        StructPtr: Data<number, number>
-        StructSlice: Data<string, string>[]
-        StructMap: { [key: string]: Data<string, number> }
-        StructMapPtr: { [key: string]: Data<string, string> }
+        StructPtr: (Data<number, number> | null)
+        StructSlice: (Data<string, string> | null)[]
+        StructMap: { [key: string]: (Data<string, number> | null) }
+        StructMapPtr: ({ [key: string]: (Data<string, string> | null) } | null)
```
eandre, line 215: Not sure about all of these. They're technically correct but very unlikely to match the user's intention. Can we do better somehow?

eandre: Looking through all these changes, I feel like we're more technically correct at the expense of pragmatism.

I think we should think through these changes and see if we can do a better job inferring user intent somehow.

One idea is to define an `Option[T]` type and ask users to use that to communicate that something is optional instead of `*T`.

Another idea is to default to assuming pointers are non-null in certain (or all?) cases and add a validation struct tag that communicates that it's nullable/optional.

cli/daemon/run/testdata/echo_client/client.ts
DomBlack: True, but our recursive nature of how our types signatures are generated makes it very hard to do this without a large scale refactor

cli/daemon/run/testdata/echo_client/client.ts
DomBlack: From my understanding on the `?` optional field, it only according to my reading of the spec means `undefined` - `null` is different to undefined. -> https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties

`omitempty` isn't possible on the AuthParams. so it will neer be undefined here.

cli/daemon/run/testdata/echo_client/client.ts
DomBlack: The user could _not_ use a pointer then...

cli/daemon/run/testdata/echo_client/client.ts
eandre: That's true, but taken as a whole this change seems to make a bunch of changes to generated code that are (1) sometimes more incorrect, (2) sometimes more correct but unlikely to match user intent, and (3) more difficult to understand. 

These changes will cause compiler errors all over frontend application code.

On balance this doesn't seem like a good trade off. What are our options?

cli/daemon/run/testdata/echo_client/client.ts
DomBlack: When we add the validate system, a `validate:"present"` flag could be used to signal intention here.

But the biggest usecase I see for pointers on API parameters and response types (in the real world) is this:

```go
type Params {
    Index *int `json:",omitempty"`
}
```

This allows the API to understand the difference between `0` and `undefined` / `null`

cli/daemon/run/testdata/echo_client/client.ts
DomBlack: > (1) sometimes more incorrect,
I disagree with this,

> (2) sometimes more correct but unlikely to match user intent, 
I disagree with this too, I think our existing behaviour is already unlikely to match user intent

> (3) more difficult to understand.
Only because I'm trying to keep this a minimal change set, I already have ideas on a better refactor to combine all type generators

cli/daemon/run/testdata/echo_client/client.ts
eandre: Exactly, so what can we do to more  accurately infer the user's intent?

DomBlack: For now we've decided to defer this so we can think through a better design. (ENC-827).

ekerfelt (approved):

PR #399, opened 2022-09-30
Reset clients on logout

ekerfelt:
We need to recreate the default token source and auth client to actually logout without restarting the daemon


eandre (approved):
Nice catch!

PR #403, opened 2022-10-03
cli/cmd/encore: throw error if db doesn't exist

### LINKED ISSUE #401: `encore db conn-uri <system>` does not throw a warning if the system/db does not eixst
### Problem

For example, I run `encore db conn-uri hello` but my application don't have a service called `hello`. We should throw an error saying "The hello service does not exist"

### Solution

The relevant code for this CLI command is here:
https://github.com/encoredev/encore/blob/main/cli/cmd/encore/db.go#L173-L215

And specifically here:
https://github.com/encoredev/encore/blob/main/cli/cmd/encore/db.go#L184

Something like this might work (psuedocode):

```diff
  if len(args) > 0 {
+  if _, err := os.Stat(filepath.Join(appRoot, args[0])); err == nil {
+    dbName = args[0]
+  }
  } else {
```

miles170:
Closes #401.

### DIFF
--- a/cli/cmd/encore/db.go
+++ b/cli/cmd/encore/db.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"encr.dev/cli/daemon/sqldb/docker"
 	daemonpb "encr.dev/proto/encore/daemon"
@@ -207,6 +209,12 @@ var dbConnURICmd = &cobra.Command{
 			EnvName: dbEnv,
 		})
 		if err != nil {
+			st, ok := status.FromError(err)
+			if ok {
+				if st.Code() == codes.NotFound {
+					fatalf("no such database found: %s", dbName)
+				}
+			}
 			fatalf("could not connect to the database for service %s: %v", dbName, err)
 		}
 
--- a/cli/daemon/daemon.go
+++ b/cli/daemon/daemon.go
@@ -204,6 +204,11 @@ func (s *Server) availableUpdate() string {
 	return ""
 }
 
+var errDatabaseNotFound = (func() error {
+	st := status.New(codes.NotFound, "database not found")
+	return st.Err()
+})()
+
 var errNotLinked = (func() error {
 	st, err := status.New(codes.FailedPrecondition, "app not linked").WithDetails(
 		&errdetails.PreconditionFailure{
--- a/cli/daemon/db.go
+++ b/cli/daemon/db.go
@@ -50,6 +50,17 @@ func (s *Server) dbConnectLocal(ctx context.Context, req *daemonpb.DBConnectRequ
 		return nil, err
 	}
 
+	databaseExists := false
+	for _, s := range parse.Meta.Svcs {
+		if s.Name == req.DbName {
+			databaseExists = len(s.Migrations) > 0
+			break
+		}
+	}
+	if !databaseExists {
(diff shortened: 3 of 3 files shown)


cli/cmd/encore/db.go
```diff
@@ -182,6 +182,9 @@ var dbConnURICmd = &cobra.Command{
 		dbName := ""
 		if len(args) > 0 {
 			dbName = args[0]
+			if _, err := os.Stat(filepath.Join(appRoot, dbName)); err != nil {
```
eandre, line 185: It's possible for a service directory to exist as a subdirectory and not necessarily live directly in the application root.   

But possibly it would be better if we had the daemon return a dedicated status code if the service does not exist, and handle this in the error check on line 210 instead.

eandre (commented):
Thanks @miles170 for the submission! I posted a comment that would be good to address. It makes the change a bit larger but is more correct. Do you want to take a stab?

cli/cmd/encore/db.go
eandre: Thanks @miles170. You're right to think we shouldn't modify generated code. However what you see is the generated code for the GRPC connection. The server-side implementation is here: https://github.com/encoredev/encore/blob/main/cli/daemon/db.go#L22 (which is not generated code)

cli/daemon/db.go
```diff
@@ -20,6 +23,18 @@ import (
...
+	filepath.WalkDir(req.AppRoot, func(path string, info fs.DirEntry, err error) error {
+		if info.IsDir() && filepath.Base(path) == req.DbName {
+			serviceExist = true
+			return io.EOF
+		}
+		return nil
+	})
+	if !serviceExist {
+		return nil, errServiceNotFound
+	}
```
eandre, line 36: As I mentioned before we should not look at just the directory names but inspect the Encore application metadata. 

Move this down into `dbConnectLocal` and instead of walking the filesystem we should inspect the `parse.Meta` object. It is a protobuf-generated data structure that you can see here: https://github.com/encoredev/encore/blob/main/proto/encore/parser/meta/v1/meta.proto#L10

What we need to do is to iterate over the `Svcs` list and see if one exists that matches the `req.DbName`.

If so, we know that the service exists, and then we need to check whether that service has a database.

The way to do that is by checking if the `Migrations` slice is non-empty, see: https://github.com/encoredev/encore/blob/main/proto/encore/parser/meta/v1/meta.proto#L46

cli/daemon/daemon.go
```diff
@@ -204,6 +204,11 @@ func (s *Server) availableUpdate() string {
 	return ""
 }
 
+var errServiceNotFound = (func() error {
```
eandre, line 207: Can we make this `database not found` instead? Both in the identifier here as well as the error message.

Finally, we need to check for this error on the client side, using https://pkg.go.dev/google.golang.org/grpc/status#Code to check for `codes.NotFound`, and print a helpful error message there.

eandre (commented):
This is starting to look great! A few minor things but almost there. Thanks so much!

cli/cmd/encore/db.go
```diff
@@ -207,6 +209,12 @@ var dbConnURICmd = &cobra.Command{
 			EnvName: dbEnv,
 		})
 		if err != nil {
+			st, ok := status.FromError(err)
+			if ok {
+				if st.Code() == codes.NotFound {
+					fatalf("The %s service does not exist", dbName)
```
eandre, line 215:
```suggestion
					fatalf("no such database found: %s", dbName)
```

eandre (approved):

PR #405, opened 2022-10-03
[ENC-796] Add support for Environmental selectors

DomBlack:
This commit adds support for environmental selectors to the CUE configuration system allowed in Encore applications.

We've decided to embed the configuration in a definition called `#Meta` which can easily be referenced but shouldn't interfere too much with the end users' own CUE definitions.

We've also moved the application's own config into a `#Config` definition which is closed. That `#Config` is then inlined, which right now has no behaviour difference from how we were doing it. However, in an upcoming version of CUE, an inlined definition will be considered closed and will cause an error _if_ a field is added to the config file which is not present in the closed definition. i.e. this will detect typos.

With this change, the follow configuration resulted in `encore run` logging `932` for `FieldC`, however `encore test` on the exact same code, resulted in `FieldC` logging `1337`.

```cue
FieldC: [
	if #Meta.Environment.Type  == "test"  { 1337 },
	if #Meta.Environment.Cloud == "local" { 932 },
	42
][0]
```

Screenshot 2022-10-03 at 15 59 33

**Breaking Changes**

This commit removes the `EnvironmentType` `EnvLocal` and adds a new `EnvTest` from the meta data API. This allows an Encore app to understand if it's running within a unit test or not. I believe this makes more sense as an app running under `encore run` will now have an environment of `EnvDevelopment` and a `CloudProvider` of `CloudLocal`. However running under `encore test`, it will be `EnvTest` && `CloudLocal`.

### DIFF
--- a/cli/daemon/run.go
+++ b/cli/daemon/run.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/logrusorgru/aurora/v3"
+	"github.com/rs/zerolog/log"
 	"golang.org/x/mod/modfile"
 
 	"encr.dev/cli/daemon/export"
@@ -26,6 +27,7 @@ import (
 	"encr.dev/cli/internal/version"
 	"encr.dev/internal/optracker"
 	"encr.dev/parser"
+	"encr.dev/pkg/errlist"
 	"encr.dev/pkg/vcs"
 	daemonpb "encr.dev/proto/encore/daemon"
 	meta "encr.dev/proto/encore/parser/meta/v1"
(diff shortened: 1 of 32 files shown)


cli/daemon/run.go
```diff
@@ -212,6 +214,20 @@ func (s *Server) Test(req *daemonpb.TestRequest, stream daemonpb.Daemon_TestServ
...
+				case error:
+					err = recovered
+				default:
+					err = fmt.Errorf("%v", recovered)
+				}
+				log.Err(err).Msg("panic during test run")
+				testResults <- fmt.Errorf("panic occured witihn Encore during test run: %w\n%s", recovered, errlist.GetStack())
+			}
+		}()
+
```
DomBlack, line 230: Previously a panic within Encore's codebase during a call to `encore test` would crash the daemon without logging anything to the CLI. This change enables that logging and keeps the daemon running.

compiler/internal/cuegen/service.go
```diff
@@ -178,9 +196,6 @@ func (s *service) generateCue() error {
 		s.file.Decls = append(s.file.Decls, field)
 	}
 
-	// Now write the top level fields required in the config
-	s.file.Decls = append(s.file.Decls, s.topLevelFields...)
```
ekerfelt, line 182: why aren't we adding the topLevelFields to the decls?

ekerfelt (approved):

PR #407, opened 2022-10-04
[ENC-832] Error if config is referenced outside the service

DomBlack:
This commit adds a new check that prevents a configuration variable from being referenced from outside the service. This behaviour is needed as each service loads it's own unique configuration, and services will not be awaire of each others configurations at runtime.

Configuration passing can be achieved via API calls, rather than direct variable refernces.

### DIFF
--- a/parser/testdata/config_err_use_from_other_service.txt
+++ b/parser/testdata/config_err_use_from_other_service.txt
@@ -1,0 +1,38 @@
+# Verify calls to config.Load are called
+! parse
+stderr 'A config instance can only be referenced from within the service that the call to'
+
+-- svc/svc.go --
+package svc
+
+import (
+    "context"
+
+    "encore.dev/config"
+)
+
+type Config struct {
+    FooEnabled bool
+}
+
+var Cfg = config.Load[Config]()
+
+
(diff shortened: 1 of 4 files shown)


bih (approved):

parser/testdata/config_err_use_from_other_service.txt
```diff
@@ -0,0 +1,38 @@
...
+
+-- libraries/foo/foo.go --
+package foo
+
+import (
+    "test/svc"
+)
+
+func init() {
+    if svc.Cfg.FooEnabled {
```
eandre, line 35: We need to support accepting the config objects as inputs to functions inside libraries. Is that still supported?

DomBlack: Yes, because the resource usage tracker only tracks the _original_ identifier the configuration instance is loaded onto.

As such, you can reference it and pass the whole instance to a library function and that will work fine - as long as that reference to it is done from within the original service that the config is loaded in

eandre: Ok great!

PR #408, opened 2022-10-04
[ENC-799] Error on Unexported Config Fields

DomBlack:
If a data type has unexported fields and is used by a call to `config.Load[T]()` then we will now report an error to the user. This error is to prevent undefined behaviour when a user expects to be able to load an unexported field and is not able to.

This error will only occur on unexported fields used by config, unexported fields used by any other Encore feature will be unaffected.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -59,7 +59,7 @@ type parser struct {
 	decls                       []*schema.Decl
 	paths                       paths.Set                          // RPC paths
 	resourceMap                 map[string]map[string]est.Resource // pkg/path -> name -> resource
-	structsWithUnexportedFields map[*schema.Struct]*ast.Field      // A struct will be in this map if it has unexported fields
+	hasUnexportedFields map[*schema.Struct]*ast.Field      // A struct will be in this map if it has unexported fields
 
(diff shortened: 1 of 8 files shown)


parser/parser.go
```diff
@@ -41,24 +41,25 @@ type parser struct {
...
+	pubSubTopics                []*est.PubSubTopic
+	cacheClusters               []*est.CacheCluster
+	names                       names.Application
+	authHandler                 *est.AuthHandler
+	middleware                  []*est.Middleware
+	declMap                     map[string]*schema.Decl // pkg/path.Name -> decl
+	decls                       []*schema.Decl
+	paths                       paths.Set                          // RPC paths
+	resourceMap                 map[string]map[string]est.Resource // pkg/path -> name -> resource
+	structsWithUnexportedFields map[*schema.Struct]*ast.Field      // A struct will be in this map if it has unexported fields
```
eandre, line 62: This is really long compared with all the other fields which makes it quite awkward. How about:
```suggestion
	hasUnexportedFields map[*schema.Struct]*ast.Field      // A struct will be in this map if it has unexported fields
```

eandre (approved):

PR #415, opened 2022-10-08
Allow `encore test` to support `-json`.

DomBlack:
This commit allows encore to obey the `-json` arg
which instructs the tests to output everything in
raw JSON. Currently we where always pretty printing the JSON output, which broke tooling intergration
(such as the GoLand plugin).

### DIFF
--- a/cli/cmd/encore/test.go
+++ b/cli/cmd/encore/test.go
@@ -2,10 +2,13 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"os/signal"
+	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 
 	daemonpb "encr.dev/proto/encore/daemon"
 )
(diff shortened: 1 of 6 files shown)


cli/cmd/encore/test.go
```diff
@@ -50,7 +56,7 @@ func runTests(appRoot, testDir string, args []string) {
 	if err != nil {
 		fatal(err)
 	}
-	os.Exit(streamCommandOutput(stream, true))
+	os.Exit(streamCommandOutput(stream, convertJSON))
```
DomBlack, line 59: Note before merging, I want to get this working so we still convert the JSON logs to into pretty printed logs within the JSON test format. Hence why this PR is draft still

eandre:
I believe this JSON output unwrapping was intended to be the JSON from the structured logging, not to unwrap "go test -json" output. With that in mind, is this the correct change?

I feel like we have lots of minor issues related to JSON parsing (like the test output on the platform) and I just want to make sure we're moving in the right direction.

eandre (approved):

PR #424, opened 2022-10-10
Add a JavaScript client to the code generation

### LINKED ISSUE #371: Client gen code support: JavaScript
We currently support TypeScript, but for older clients, they may want an untyped version if they don't use TypeScript at all in their codebase.

See code files here:
- <https://github.com/encoredev/encore/blob/main/cli/internal/codegen/typescript.go>
- <https://github.com/encoredev/encore/blob/main/cli/internal/codegen/testdata/expected_typescript.ts>
- <https://github.com/encoredev/encore/blob/main/cli/internal/codegen/testdata/expected_noauth_typescript.ts>

Minivera:
For older or client who do not use TypeScript, having a JavaScript client generated with code similar to the TypeScript client could be fairly useful. I have tried to implement that JavaScript client while staying as true to the coding style of the TypeScript client as possible. I have tested it locally and implemented the tests to make sure it generates the clients as expected. Indentation was a bit annoying in parts, but I think this works pretty great for what it is.

Some alternative solutions:

- For creating the TypeScript client, we could also use the TypeScript compiler to compile the TypeScript client to JavaScript instead.
- For formatting, we could write teh code non-formatted and format it with a binary tool like [Rome](https://rome.tools/), which might work better than trying to ident it manually.

Closes #371


eandre:
Thanks very much @Minivera! We do need the formatting to be handled correctly by the code generator directly as we don't want to depend on additional tools for this. How much work do you think that would be?

Minivera:
> How much work do you think that would be?

The formatting should be the same as the TypeScript client I think? I commented on possible alternative solutions since it was slightly annoying to get going, but I think it's correct 🙂

CONTRIBUTING.md
```diff
@@ -95,6 +95,11 @@ npm test #run all tests
 npm test -- --watch #re-run tests on file change
 ```
 
+### Testing the Daemon run logic
+The deamon tests in the `cli/daemon/run` use many auto generated files from the `e2e-tests/testdata` directory. To generate the
```
eandre, line 99: Is this true? I cannot find any reference to `e2e-tests` in the `cli/` folder or any subdirectory.

e2e-tests/testdata/echo_client/client.js
```diff
@@ -0,0 +1,912 @@
...
+ */
+export default class Client {
+    cache
+    di
+    echo
+    endtoend
+    flakey_di
+    middleware
+    test
+    validation
```
eandre, line 27: Is this normal JSON? I've never seen this pattern before. I guess it's not necessary and we should drop it?

e2e-tests/testdata/echo_client/client.js
```diff
@@ -0,0 +1,912 @@
...
+
+
+    /**
+     * Creates a Client for calling the public and authenticated APIs of your Encore application.
+     *
+     * @param target  The target which the client should be configured to use. See Local and Environment for options.
+     * @param options Options for the client
+     */
+    constructor(target = "prod", options) {
```
eandre, line 36: Having a default value followed by something without defaults looks quite strange. How about:
```suggestion
    constructor(target = "prod", options = undefined) {
```

e2e-tests/testdata/echo_client/client.js
```diff
@@ -0,0 +1,912 @@
...
+        this.endtoend = new endtoend.ServiceClient(base)
+        this.flakey_di = new flakey_di.ServiceClient(base)
+        this.middleware = new middleware.ServiceClient(base)
+        this.test = new test.ServiceClient(base)
+        this.validation = new validation.ServiceClient(base)
+    }
+}
+
+class CacheServiceClient {
+    baseClient
```
eandre, line 50: Same here as above

e2e-tests/testdata/echo_client/client.js
```diff
@@ -0,0 +1,912 @@
...
+        )
+    }
+    return value
+}
+
+class BaseClient {
+    baseURL
+    fetcher
+    headers
+    authGenerator
```
eandre, line 558: Same here

e2e-tests/testdata/echo_client/.eslintrc.cjs
```diff
@@ -5,6 +5,12 @@ module.exports = {
     plugins: [
         '@typescript-eslint',
     ],
+    env: {
+        browser: true,
+        es6: true,
+        node: true,
+        commonjs: true
+    },
```
eandre, line 13: What does this do?

eandre (commented):
Thanks, this is great overall. A few minor points in the output.

On a final note, we'd like to get this merged for Hacktoberfest. One issue that remains is
we have some outstanding fixes for the TypeScript code generation that will require some
changes to the JS client as well to stay in sync.

I wonder if it would make sense to merge this but keep it disabled for a few days until we have a chance to make those fixes? What do you think @Minivera?

Minivera:
> I wonder if it would make sense to merge this but keep it disabled for a few days until we have a chance to make those fixes? What do you think @Minivera?

I think that would make sense, since the TypeScript client changed a lot (files moved around as well, which makes the docs I wrote outdated) since I started this PR. I think it would work best to be iterative like you suggest, maybe even take the opportunity to merge some things together.

e2e-tests/testdata/echo_client/.eslintrc.cjs
Minivera: This tells ESLint to not throw any errors based on which APIs are available in which environment. By telling ESLint this is a library that will work with ES6, in the browser, and in node, then it understands to not throw for things like the `fetch` API being missing in older versions of node.

eandre (approved):

eandre (approved):

eandre:
Thank you @Minivera!

PR #425, opened 2022-10-10
Small improvements to the ent ORM documentation page

Minivera:
The documentation page had a few issues that were highlighted on Slack and while building a good example app, namely:

- The URL for installing ent changed, which send the reader on the wrong page and wasn't clear on how to install ent in an existing codebase.
- If setting up ent in a clean database, the database wouldn't exist, which would make the migrations fail. I added a section to clear this up.
- The review on the example app PR had a great suggestion to clean up the migration generation code. I've added it here too.
- There was a possible confusion where readers may try to use ent types as params or return types in Encore. That won't work, I added an example and a disclaimer.

This should make the docs a bit easier to follow and unblock any reader who tried to set it up in a new codebase.

### DIFF
--- a/docs/how-to/entgo-orm.md
+++ b/docs/how-to/entgo-orm.md
@@ -20,7 +20,7 @@ it uses plain SQL migration files for its migrations.
 Let's take a look at how you may integrate ent with encore.
 
 ## Add ent schemas to a service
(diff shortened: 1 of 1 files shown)


eandre (approved):
Looks great, thanks!

PR #430, opened 2022-10-12
Use 'encore' to keep track of the upstream repository

campesel:
`encore clone` uses `origin` to keep track of the upstream repository. Instead, it should use `encore` for `git push encore` to work.

This PR adds a flag `--origin` to the `git clone` command (which is called by `encore clone`) to use the name `encore` to keep track of the upstream repository.


eandre (approved):
Nice catch!

PR #431, opened 2022-10-13
Move our End-to-End tests to the top level.

DomBlack:
This commit moves our end to end tests to the top level of Encore's project in a `e2e-tests` folder.

To achieve this I needed to move some internal packages from `/cli/internal/...` to `/internal/...`

This also fixes an issue with our build process to do with the font's folder not being embedable


e2e-tests/README.md
```diff
@@ -0,0 +1,11 @@
+# End to End Tests
```
DomBlack, line 1: This is the only new file, everything else has just been moved around on the file system.

eandre (approved):

PR #435, opened 2022-10-13
[ENC-888] Improve NG-Error Rendering on Client.

DomBlack:
This commit improves the next generation errors rendering on the CLI. It achieves this by introducing a new Command between the daemon and the front end called display error. Whenever we want to display an error list, this command is used to send the full `errlist.List` object to the front end, which then renders it to `stderr`.

The reason we render the error list on the front end is:
1. We can calculate the size of the terminal window and then format the error to fix perfectly within the terminal (this is done by changing `errinsrc.ErrorWidth` before render).
2. If the command is run with `--verbose`, then only the client knows that, by rendering client side, `errlist` can obey the verbose command.

In the future:
1. Using bubbletea, we can expose a terminal UI for paging between multiple errors (when it shows `(and x more errors)`)
2. We can render errors nice the local dashboard within the browser.

### Live Reload Errors on a wide terminal
image

#### Initial compile failures on a narrow terminal
image


cli/daemon/run/watch.go
```diff
@@ -37,12 +39,17 @@ func (mgr *Manager) watch(run *Run) error {
...
+					} else {
+						errStr := err.Error()
+						if !strings.HasSuffix(errStr, "\n") {
+							errStr += "\n"
+						}
+						mgr.RunStderr(run, []byte(errStr))
 					}
-					mgr.RunStderr(run, []byte(errStr))
 				} else {
+					mgr.RunError(run, nil)
```
eandre, line 52: Is this needed? That's very unintuitive, if so

pkg/errinsrc/errinsrc.go
```diff
@@ -88,16 +88,21 @@ func (e *ErrInSrc) Bailout() {
...
 
 	// Write the header
 	const headerGrayLevel = 12
 	const spacing = 4 + 2 + 7 // (4 = "--" on both sides, 2 = " " on the sides of the title, 7 = "[E0000]")
+	b.WriteRune('\n')         // Always start with a new line as these errors are expected to be full screen
 	b.WriteString(aurora.Gray(headerGrayLevel, fmt.Sprintf("%c%c ", set.HorizontalBar, set.HorizontalBar)).String())
 	b.WriteString(aurora.Red(e.Params.Title).String())
 	b.WriteByte(' ')
-	b.WriteString(aurora.Gray(headerGrayLevel, strings.Repeat(string(set.HorizontalBar), errorWidth-len(e.Params.Title)-spacing)).String())
+	b.WriteString(aurora.Gray(headerGrayLevel, strings.Repeat(string(set.HorizontalBar), ErrorWidth-len(e.Params.Title)-spacing)).String())
```
eandre, line 105: With very small terminals these can become negative. Do we handle that? Should we enforce a minimum value?

cli/daemon/run/watch.go
DomBlack: For the bubbletea UI I want to add yes, I'll remove it for now though

pkg/errinsrc/srcrender.go
```diff
@@ -275,7 +275,10 @@ func renderErrorText(builder *strings.Builder, startCol int, numDigitsInLineNumb
 
 func splitTextOnWords(text string, startingCol int) (rtn []string) {
 	text = strings.TrimSpace(text)
-	maxLineLength := errorWidth - startingCol
+	maxLineLength := TerminalWidth - startingCol
+	if maxLineLength <= 0 {
```
eandre, line 279: Should this be
```suggestion
	if maxLineLength < 20 {
```

eandre (approved):

PR #447, opened 2022-10-17
Add config in service to metadata

DomBlack:


pkg/errlist/errlist.go
```diff
@@ -91,6 +91,7 @@ func (l *List) Report(err error) {
 				l.Report(e)
 			}
 		}
+		return
```
DomBlack, line 94: Fixed a panic if a list was reported; as this was missing :(

proto/encore/parser/meta/v1/meta.proto
```diff
@@ -45,6 +45,7 @@ message Service {
   repeated RPC         rpcs       = 3;
   repeated DBMigration migrations = 4;
   repeated string      databases  = 5; // databases this service connects to
+  bool                 has_config = 6; // true if the service has uses config
```
ekerfelt, line 48: are you going to use this in the deploy step?

ekerfelt (approved):

proto/encore/parser/meta/v1/meta.proto
DomBlack: Yeah it's being used to detect which services need config computed for them

PR #450, opened 2022-10-19
[ENC-921] Update `publicapigen` for Config

DomBlack:
This commit modifies publicapigen in a couple of ways:

- Any arguments prefixed with `__` are now removed from functions. This was to support the `config.Load[T]()` wanted output, given we actually take multiple arguments in the final code.
- Comments are added above each of the `panic()` calls, to explain why the panic is there.
- Those comments also point to a permalink of the original function, if people want to look at the original code while trying to debug an issue they're seeing as we've heard feedback it's hard to do right now.

These changes can be seen in encoredev/encore.dev#11

### DIFF
--- a/tools/publicapigen/main.go
+++ b/tools/publicapigen/main.go
@@ -12,6 +12,8 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"sort"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


eandre (dismissed):
Nice!

eandre (approved):

PR #453, opened 2022-10-20
Publish predefined Encore logs-based metrics

campesel:
This PR adds the code to automatically publish a set of predefined metrics for all Encore apps running in Encore Cloud, GCP or AWS.

The metrics we're looking to add initially are:

- Request count by service/endpoint
- Error count by service/endpoint/error code
- Response time by service/endpoint/status code

The idea is that we use metric filters that count and extract data from JSON log lines, and then publish metrics to Google Cloud Monitoring/AWS CloudWatch.

Metrics can be visualized through cloud-native dashboards or via Grafana integrations.


runtime/appruntime/api/reqtrack.go
```diff
@@ -97,6 +97,11 @@ func (s *Server) beginRequest(ctx context.Context, p *beginRequestParams) (*mode
 		curr.Trace.BeginRequest(req, curr.Goctr)
 	}
 
+	s.customMetrics.Counter("e-requests-total", map[string]string{
```
campesel, line 100: I prefixed the counter name with `e` to make it less likely for a user to pick the same metric name (when we support user-defined metrics).

runtime/appruntime/app/app.go
```diff
@@ -61,6 +62,7 @@ func New(p *NewParams) *App {
 		})
 	}
 	rootLogger := zerolog.New(logOutput).With().Timestamp().Logger()
+	customMetrics := custommetrics.NewManager(cfg.Runtime.AppSlug, cfg.Runtime.EnvCloud, rootLogger)
```
campesel, line 65: I was thinking, we could read a `metrics_enabled` flag from config and pass it to `NewManager` to enable/disable metrics?

runtime/custommetrics/custom_metrics.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"strings"
+
+	"github.com/rs/zerolog"
+
+	encore "encore.dev"
+)
+
+type Manager interface {
```
campesel, line 11: I'm not sure `Manager` is a good name for this. I'm open to suggestions!

runtime/custommetrics/custom_metrics.go
```diff
@@ -0,0 +1,43 @@
...
+			metricPrefix: metricPrefix,
+			logger:       logger,
+		}
+	case encore.CloudAzure:
+		// Custom metrics are in still in preview, so we won't be using them for now.
+	case encore.EncoreCloud:
+		impl = &gcpMetricsManager{
+			metricPrefix: metricPrefix,
+			logger:       logger,
+		}
```
campesel, line 36: All metrics managers have `metricPrefix` and `logger` fields. Should I add a new type `common` with these fields and embed it into the metrics managers?

runtime/appruntime/api/handler_test.go
```diff
@@ -86,10 +87,11 @@ func TestDesc_EndToEnd(t *testing.T) {
 		Runtime: &config.Runtime{},
 	}
 	logger := zerolog.New(os.Stdout)
+	customMetrics := custommetrics.NewManager("", "", logger)
```
campesel, line 90: Since we don't pass in a cloud environment, the custom metrics manager won't be initialized. This makes a test fail.

Do we use `local` for testing? Or should the functions that emit metrics check whether the manager is `nil`?

runtime/appruntime/app/app.go
campesel: We've decided to have metrics always enabled. We'll revisit this in the future if use cases where metrics should be disabled come up.

runtime/appruntime/api/handler_test.go
campesel: We've decided to have a metrics exporter for tests, so that we can test that metrics get emitted when an endpoint is called, for example.

runtime/appruntime/app/app.go
```diff
@@ -136,3 +138,22 @@ func jsonAPI(cfg *config.Config) jsoniter.API {
...
+		return metrics.NewAWSMetricsExporter(cfg.Runtime.AppSlug, logger)
+	case encore.CloudGCP:
+		return metrics.NewGCPMetricsExporter(cfg.Runtime.AppSlug, logger)
+	case encore.CloudAzure:
+		// Custom metrics are in still in preview, so we won't be using them for now.
+		return metrics.NewNullMetricsExporter()
+	case encore.EncoreCloud:
+		return metrics.NewGCPMetricsExporter(cfg.Runtime.AppSlug, logger)
+	case encore.CloudLocal:
+		// TODO
```
campesel, line 154: Should we support metrics locally straight off the bat?

runtime/appruntime/api/handler_test.go
```diff
@@ -139,4 +144,40 @@ func TestDesc_EndToEnd(t *testing.T) {
 			}
 		})
 	}
+
+	testMetricsExporter.AssertCounter(t, "e_requests_total", 2, map[string]string{
+		"service":  "service",
+		"endpoint": "endpoint",
+	})
```
campesel, line 151: The total number of requests is 2 because the auth handler (which would increase this counter) isn't called in this test. I think this is expected.

runtime/appruntime/metrics/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Predefined Encore metrics
+
+These are the metrics all Encore apps export to cloud providers:
+
+- `e_requests_total` measures the number of requests and has two tags `service` and `endpoint`.
+- `e_errors_total` measures the number of errors and has three tags `service`, `endpoint` and `code`. `code` is a
+  human-readable error code (e.g. `not_found`).
+- `e_request_durations_milliseconds` measures the response time and has three tags `service`, `endpoint`
+  and `status_code`. `status_code` is an HTTP status code (e.g. `201`).
```
campesel, line 16: Open to suggestions on metric and tag names!

runtime/appruntime/metrics/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if httpStatus == 0 {
+		httpStatus = http.StatusOK
+	}
+	m.exp.Observe(
+		"e_request_durations_milliseconds",
+		"duration", float64(durMillis),
+		"service", service,
+		"endpoint", endpoint,
+		"status_code", strconv.Itoa(httpStatus),
```
campesel, line 45: `status_code` differs from `code` (used above) in that it's not supposed to be human-readable. This is why they're not named the same.

For consistency, we could name them both `code` and convert the HTTP status code into a string such as `http_201`, for example. Alternatively, we could use the `StatusText` function (see [docs](https://pkg.go.dev/net/http#StatusText)) to convert an HTTP status code into a human-readable code.

runtime/appruntime/metrics/README.md
```diff
@@ -0,0 +1,16 @@
...
+This package allows an Encore app to export custom metrics to cloud providers.
+
+We have two types of custom metrics: a predefined set of custom metrics which are enabled for all Encore apps and
+user-defined custom metrics. User-defined custom metrics aren't supported yet.
+
+## Predefined Encore metrics
+
+These are the metrics all Encore apps export to cloud providers:
+
+- `e_requests_total` measures the number of requests and has two tags `service` and `endpoint`.
```
DomBlack, line 12: I'd change the prefix to `encore_` and then reserve that prefix

runtime/appruntime/app/app.go
```diff
@@ -136,3 +138,22 @@ func jsonAPI(cfg *config.Config) jsoniter.API {
...
+	case encore.CloudAWS:
+		return metrics.NewAWSMetricsExporter(logger)
+	case encore.CloudGCP:
+		return metrics.NewGCPMetricsExporter(logger)
+	case encore.CloudAzure:
+		// Custom metrics are in still in preview, so we won't be using them for now.
+		return metrics.NewNullMetricsExporter()
+	case encore.EncoreCloud:
+		return metrics.NewGCPMetricsExporter(logger)
+	case encore.CloudLocal:
```
DomBlack, line 153:
```suggestion
	switch cfg.Runtime.Metrics.ExporterType {
	case "LogBased":
		return metrics.NewLogBasedExporter(logger)
	case "datadog":
		return metrics.NewDataDogExporter(logger, cfg.Runtime.Metrics.DataDogCfg)
	case "": // no metrics
		return metrics.NewNullMetricsExporter()
	default:
	panic("unknown metrics type")
	}
```

runtime/appruntime/metrics/README.md
eandre: I rather like that the prefix is short and sweet. Seems unlikely to collide with anything?

runtime/appruntime/metrics/README.md
DomBlack: - `encore_pubsub_publish_counter` (`topic`, `service`, `error_code`)
- `encore_pubsub_receive_counter` (`topic`, `subscription`, `error_code`)
- `encore_pubsub_receive_duration_ms` (`topic`, `subscription`, `error_code`)
- `encore_cache_read_counter` ( `cluster`, `keyspace`, `hit` )
- `encore_cache_write_counter` (`cluster`, `keyspace` )
- `encore_sqldb_query_counter` (`service`, `database`, `error_code`)
- `encore_sqldb_query_duration_ms` (`service`, `database`, `error_code`)

runtime/appruntime/metrics/README.md
campesel: We've gone with the short prefix.

runtime/appruntime/metrics/README.md
```diff
@@ -0,0 +1,16 @@
...
+- `e_errors_total` measures the number of errors and has three tags `service`, `endpoint` and `code`. `code` is a
+  human-readable error code (e.g. `not_found`).
+- `e_request_durations_milliseconds` measures the response time and has three tags `service`, `endpoint`
```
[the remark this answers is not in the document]
campesel: We've decided to follow [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/), so that we don't have to change our metric names when we support Prometheus.

runtime/appruntime/metrics/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	)
+}
+
+func code(err error, httpStatus int) string {
+	if err != nil {
+		e := errs.Convert(err).(*errs.Error)
+		return e.Code.String()
+	}
+
+	code := http.StatusText(httpStatus)
```
eandre, line 52: I don't think the output of `http.StatusText` is suitable for time series values; they're intended as human-readable text. What's more, most people will not know off the top of their head what HTTP status code `non_authoritative_information` refers to, which makes it difficult to reason about and search their code base for.

I think we should either do:

(a) only HTTP status codes (`400` etc), regardless of error code
(b) use our codes (`internal` etc) when the code matches one of those and `http_123` otherwise
(c) output both when we can (but might not work with AWS tag limits)

What do you think @DomBlack?

DomBlack: I think using our own codes makes the most sense here; it aligns with how people think about their programs when writing errors in Encore

eandre: So you're advocating for (b)? Or something else?

DomBlack: (b)

runtime/appruntime/metrics/README.md
DomBlack: Are you sure that no user ever will want to use `e_` as a prefix in their structured logs?

I can easily imagine people wanting to use `enc_` as a prefix, what's the issue with using `encore_`? Given that it's not actually visible to users at the end?

runtime/appruntime/metrics/metrics.go
campesel: Sounds good to me too, I've pushed a commit that implements (b).

runtime/appruntime/config/config.go
```diff
@@ -262,3 +263,13 @@ type RedisDatabase struct {
 	// without having to coordinate and persist database index ids.
 	KeyPrefix string `json:"key_prefix"`
 }
+
+type MetricsExporterType string
+
+const (
+	MetricsExporterTypeLogsBased MetricsExporterType = "logs-based"
```
eandre, line 270: Nit: Can we make this `logs_based` maybe? Or was this format inspired by something else?

runtime/appruntime/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+func code(err error, httpStatus int) string {
+	if err != nil {
+		e := errs.Convert(err).(*errs.Error)
+		return e.Code.String()
+	}
+
+	if httpStatus == 0 || httpStatus == http.StatusOK {
+		return errs.OK.String()
+	}
+	return "http_" + strconv.Itoa(httpStatus)
```
eandre, line 49: I think if the status code is one of the ones we recognize we should probably output that instead. You can use `errs.HTTPStatusToCode(httpStatus)` to get that, and fall back to `http_N` if that function reports `errs.Unknown`

runtime/appruntime/metrics/test_exporter.go
```diff
@@ -0,0 +1,136 @@
+package metrics
```
eandre, line 1: I think it would be nice to move this file to a separate package so as to not pollute the metrics API with the test functionality. Move to a `metricstest` subdirectory?

runtime/rlog/rlog.go
```diff
@@ -242,6 +252,10 @@ func addContext(ctx zerolog.Context, key string, val interface{}) zerolog.Contex
 	}
 }
 
+func reserved(key string) bool {
+	return strings.HasPrefix(key, metrics.EncoreMetricKeyPrefix)
```
eandre, line 256: This constant is logging-related, not metrics related. It would make more sense to define it in this package (but it needs to be tagged with `//publicapigen:drop` so it doesn't get added to the public API).

eandre (commented):
Looks great! A few more minor points and then this should be ready to merge

runtime/appruntime/config/config.go
campesel: I've seen other constants having `-` instead of `_`, but I don't think all constants have `-`. I prefer `_` too!

runtime/appruntime/metrics/metrics.go
campesel: I think that the reason why it's implemented this way is to cover the case where an error happens (so `err` isn't `nil`), but `httpStatus` code is `0`. As far as I understand, this is possible?

runtime/appruntime/metrics/test_exporter.go
campesel: Great point, will do!

runtime/rlog/rlog.go
```diff
@@ -27,6 +26,9 @@ const (
 	levelInfo  logLevel = 2
 	levelWarn  logLevel = 3
 	levelError logLevel = 4
+
+	//publicapigen:drop
+	InternalKeyPrefix = "encore_"
```
eandre, line 31: Move this to a separate `const` declaration; this one is used for log levels and it makes more sense to keep the  `InternalKeyPrefix ` constant separate. Also, it would be nice to have a doc comment for it.

runtime/appruntime/metrics/metrics.go
```diff
@@ -43,8 +42,14 @@ func code(err error, httpStatus int) string {
...
 		return errs.OK.String()
 	}
-	return "http_" + strconv.Itoa(httpStatus)
+
+	errCode := errs.HTTPStatusToCode(httpStatus)
+	if errCode == errs.Unknown {
+		return "http_" + strconv.Itoa(httpStatus)
+	} else {
+		return errCode.String()
+	}
```
eandre, line 54: Nit: I would write this as:

```suggestion
	if code := errs.HTTPStatusToCode(httpStatus); code != errs.Unknown {
		return code.String()
	}
	return "http_" + strconv.Itoa(httpStatus)
```

runtime/rlog/rlog.go
```diff
@@ -26,7 +26,16 @@ const (
...
+	// InternalKeyPrefix is the prefix of log field keys that are reserved for
+	// internal use only. For example, logs-based metrics rely on a log field named
+	// 'encore_metric_name' to determine which metric is being emitted.
+	//
+	// If Encore apps add log fields that start with this prefix, we'll prefix them
+	// with 'x_' to make sure they don't interfere with internal log fields.
+	//
 	//publicapigen:drop
 	InternalKeyPrefix = "encore_"
 )
```
eandre, line 41: Nit: No need to use a `const ( ... )` block with a single item. Also, don't use language like "we'll prefix them"
in documentation because it's not clear who "we" refers to.

How about:

```suggestion
// InternalKeyPrefix is the prefix of log field keys that are reserved for
// internal use only. Log fields beginning with this value have an additional
// "x_" prefix prepended to avoid interference with reserved names.
//
//publicapigen:drop
const InternalKeyPrefix = "encore_"
```

eandre (approved):

PR #454, opened 2022-10-20
[ENC-794] Support config override in tests.

DomBlack:
This commit adds the ability for unit tests to provide overrides to configuration values, which code running within that unit test or sub tests will see, but other running tests will see the original value, even if that test is running in parallel.

* svc.go *
```go
package svc

import "encore.dev/config"

type Config stuct {
    ReadOnlyMode config.Bool
}

var cfg = config.Load[*Config]()

func ReadOnly() bool {
    return cfg.ReadOnlyMode()
}
```

* svc.cue *
```cue
ReadOnlyMode: false
```

* svc_test.go *
```go
package svc

import (
    "testing"

    "encore.dev/et"
)

func Test_Enabled(t *testing.T) {
    // In this test we override the configuration
    et.SetCfg(cfg.ReadOnlyMode, true)

    if value := ReadOnly(); value != true {
        t.Fatalf("expected true, got %b", value)
    }
}

func Test_Disabled(t *testing.T) {
    // In this test we rely on the values in the CUE files
    // even though this runs afterwards, it doesn't see the override
    if value := ReadOnly(); value != false {
        t.Fatalf("expected false, got %b", value)
    }
}
```


runtime/config/helpers_internal.go
```diff
@@ -18,18 +18,67 @@ func envName(serviceName string) string {
...
 	}
 }
 
+// GetMetaForValue returns the ValueID and ValuePath for the given Value
+func GetMetaForValue[T any](value func() T) (ValueID, ValuePath) {
+	// Get the current request
+	req := Singleton.rt.Current()
+
+	// Lock so we're the only running extraction
+	Singleton.extraction.mutex.Lock()
```
eandre, line 45: Defer unlock right after this might be clearer; I had to scan a bunch to make sure we were unlocking it correctly and it made me a bit nervous

runtime/config/manager_internal.go
```diff
@@ -0,0 +1,99 @@
...
+	"testing"
+
+	jsoniter "github.com/json-iterator/go"
+
+	"encore.dev/appruntime/model"
+	"encore.dev/appruntime/reqtrack"
+)
+
+type ValueID uint64
+type ValuePath = []string
```
eandre, line 18: Do we need this to be an alias for some reason? In general making things different types is clearer if it's warranted to have a named type in the first place

.github/workflows/go.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         echo "::set-output name=go-mod::$(go env GOMODCACHE)"
 
     - name: Install encore-go
-      run: curl --fail -L -o encore-go.tar.gz https://github.com/encoredev/go/releases/download/encore-go1.18.4/linux_x86-64.tar.gz && tar -C . -xzf ./encore-go.tar.gz
+      run: curl --fail -L -o encore-go.tar.gz https://github.com/encoredev/go/releases/download/encore-go1.19.2/linux_x86-64.tar.gz && tar -C . -xzf ./encore-go.tar.gz
```
DomBlack, line 32: Encore now uses go 1.19 features itself (`atomic.Uint64` / `atomic.Bool`)

Even though our `go.mod` file said 1.19, and we've told github at the tp of this file to use Go 1.19 - we ran tests using our older 1.18 fork

eandre (approved):

PR #458, opened 2022-10-21
[ENC-931] Allow middleware to publish to pubsub

DomBlack:
This commit modifies Pubsub such that you can now publish
to pubsub topics from global middleware. To do this we track
if the publish is within a service _or_ if it is directly called
by middleware, if it's called by middleware we then track
which middleware is doing the call, and what that middleware
is targeting.

When we convert the EST into metadata, we expand the middleware
targets out into the services which have the middleware installed.

I've also moved all pubsub errors over to the new error format

### DIFF
--- a/parser/feature_pubsub.go
+++ b/parser/feature_pubsub.go
@@ -302,6 +302,14 @@ func (p *parser) parsePubSubPublish(file *est.File, resource est.Resource, _ *wa
 		return
 	}
 
+	// FIXME: this is a temporary error, while we fix it at a deeper level
+	if file.Pkg.Service == nil {
+		p.err(
+			callExpr.Args[1].Pos(),
+			"pubsub.Publish can only be used within a service. Currently the call is not declared within a service.",
+		)
+	}
+
 	// Record the publisher
 	topic.Publishers = append(topic.Publishers, &est.PubSubPublisher{
 		DeclFile: file,
--- a/parser/feature_pubsub.go
+++ b/parser/feature_pubsub.go
@@ -296,7 +296,6 @@ func (p *parser) parsePubSubPublish(file *est.File, resource est.Resource, c *wa
 		middleware := p.findContainingMiddlewareDefinition(c)
 		if middleware == nil || !middleware.Global {
 			p.errInSrc(srcerrors.PubSubPublishInvalidLocation(p.fset, callExpr))
-			panic("needs to be in a service or global middleware")
 		}
 
 		publisher.GlobalMiddleware = middleware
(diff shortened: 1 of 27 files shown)


parser/meta.go
```diff
@@ -573,3 +589,48 @@ func sortedRefs(refs map[ast.Node]*est.Node) []refPair {
 	})
 	return p
 }
+
+type selectorMap struct {
+	services map[selector.Selector]map[*est.Service]struct{}
+	rpcs     map[selector.Selector]map[*est.RPC]struct{}
+}
+
+func newSelectorMap() *selectorMap {
```
eandre, line 598: We have something similar to this in `parser/est/selectors.go` but in the opposite direction: given an RPC, which middleware(s) apply to it. Should we move this piece of code there and generalize it to support both use cases?

parser/parser.go
```diff
@@ -589,7 +591,8 @@ func (p *parser) validateApp() {
 			}
 
 			for node, ref := range f.References {
-				if res := ref.Res; ref.Res != nil {
+				if res := ref.Res; ref.Res != nil &&
+					res.Type() != est.PubSubTopicResource { // PubSub topics are allow to be published to in global middleware, so it's ok to reference a topic outside a service
```
eandre, line 595: Nit: seems overkill to break this into two lines

runtime/appruntime/api/handler.go
```diff
@@ -208,7 +208,7 @@ func (d *Desc[Req, Resp]) invokeHandlerRaw(mwReq middleware.Request, c IncomingC
 		panic("invokeHandlerRaw called on non-Raw endpoint")
 	}
 
-	// Middleware can override the context, so check if the context is different
+	// GlobalMiddleware can override the context, so check if the context is different
```
eandre, line 211: Refactoring gone awry?

parser/feature_pubsub.go
```diff
@@ -295,17 +280,32 @@ func (p *parser) parsePubSubSubscription(file *est.File, cursor *walker.Cursor,
...
+	publisher := &est.PubSubPublisher{
 		DeclFile: file,
-	})
+	}
+
+	if file.Pkg.Service == nil {
+		middleware := p.findContainingMiddlewareDefinition(c)
+		if middleware == nil || !middleware.Global {
+			p.errInSrc(srcerrors.PubSubPublishInvalidLocation(p.fset, callExpr))
+			panic("needs to be in a service or global middleware")
```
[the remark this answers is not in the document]
DomBlack: I missed this, my bad

parser/meta.go
DomBlack: I don't think so for now, one is a generalised lookup cache, the other is a one time search

eandre (approved):

PR #460, opened 2022-10-21
Fix parsing of cache keyspace keys with type parameters

eandre:


DomBlack (approved):
Does it work if you have a generic type within the top level type? I think it should..

DomBlack (approved):

PR #474, opened 2022-10-26
[ENC-966] Always report HTTP Status in Middleware

DomBlack:
This commit changes middleware to always return an HTTP status code in the middleware response.

### DIFF
--- a/runtime/appruntime/api/handler.go
+++ b/runtime/appruntime/api/handler.go
@@ -171,6 +171,7 @@ func (d *Desc[Req, Resp]) begin(c IncomingContext) (reqData Req, beginErr error)
 func (d *Desc[Req, Resp]) handleIncoming(c IncomingContext, reqData Req) (resp Resp, httpStatus int, respErr error) {
 	if err := d.validate(reqData); err != nil {
 		respErr = err
+		httpStatus = errs.HTTPStatus(err)
 		return
 	}
 
@@ -192,12 +193,14 @@ func (d *Desc[Req, Resp]) invokeHandlerNonRaw(mwReq middleware.Request, reqData 
 	handlerResp, handlerErr := d.AppHandler(mwReq.Context(), reqData)
 	if handlerErr != nil {
 		mwResp.Err = errs.Convert(handlerErr)
+		mwResp.HTTPStatus = errs.HTTPStatus(mwResp.Err)
 	} else {
 		// Only assign the payload if we're not dealing with *Void,
 		// otherwise we would end up making Payload a "typed nil".
 		if !isVoid[Resp]() {
 			mwResp.Payload = handlerResp
 		}
+		mwResp.HTTPStatus = 200
 	}
 	return mwResp
 }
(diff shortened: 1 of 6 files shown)


runtime/appruntime/api/handler.go
```diff
@@ -358,7 +376,10 @@ func (d *Desc[Req, Resp]) Call(c CallContext, req Req) (resp Resp, respErr error
 
 		if rpcErr == nil {
 			r, rpcErr = d.CloneResp(r)
-			httpStatus = errs.HTTPStatus(errs.Convert(rpcErr))
+			if rpcErr != nil {
+				// only override the http status if an error occurred trying to clone the response
+				httpStatus = errs.HTTPStatus(errs.Convert(rpcErr))
+			}
```
DomBlack, line 382: Interestingly we always overrode the httpStatus after cloning the response, which means if middleware had set `201` or something, we'd reset it to `200`

runtime/appruntime/api/reqtrack.go
```diff
@@ -133,7 +133,7 @@ func (s *Server) finishRequest(output [][]byte, err error, httpStatus int) {
 	case model.AuthHandler:
 		req.Logger.Info().Dur("duration", dur).Msg("auth handler completed")
 	default:
-		if httpStatus != 0 {
+		if httpStatus != errs.HTTPStatus(err) {
```
eandre, line 136: Now that it's always set, does it make sense to always log the http status?

runtime/middleware/middleware.go
```diff
@@ -83,9 +83,9 @@ type Response struct {
 
 	// HTTPStatus is the HTTP status code the response is written with.
 	//
-	// For non-raw handlers it is zero by default and Encore chooses an appropriate
+	// For non-raw handlers it is initially Encore chooses an appropriate
```
eandre, line 86: I think this comment needs rewording but not sure offhand how to best word it

runtime/appruntime/api/handler.go
```diff
@@ -233,6 +236,15 @@ func (d *Desc[Req, Resp]) executeEndpoint(c execContext, invokeHandler func(midd
 	var nextFn middleware.Next
 	numMiddleware := len(d.middleware)
 	nextFn = func(req middleware.Request) (resp middleware.Response) {
+		// Ensure the HTTP status code is correctly set in the response
+		defer func() {
+			// If no explicit HTTP status has been set, then we use the default for the type of error
+			// or if Err is nil, we'll set 200
+			if resp.HTTPStatus == 0 {
```
eandre, line 243: One challenge with this is: What if a middleware returns a non-nil `Err` but keeps `HTTPStatus` as a `2xx` or `3xx` code? That would be quite confusing.

runtime/appruntime/api/reqtrack.go
DomBlack: I believed it might be better to only log the status if it's different to the expected status (i.e how we currently do it)

runtime/appruntime/api/handler.go
DomBlack: I think this is fine;

- `return &MiddlewareResp { Err: err }` would set the status to 5xx or 4xx
- `return &MiddlewareResp { Err: err, Status: 201 }` would return 2xx

So maybe we just have an explicit `|| (resp.Err != nil && resp.Status == 200)` as the override?

eandre (approved):

PR #477, opened 2022-10-26
compiler: only use external linking with cgo

eandre:
External linking only makes sense with cgo,
this was an unintended interaction.


DomBlack (commented):
Lgtm

PR #488, opened 2022-10-27
rlog: add warn log level

eandre:


proto/encore/engine/trace/trace.proto
```diff
@@ -250,10 +250,14 @@ message HTTPWroteRequestData {
...
 
 message LogMessage {
+  // Note: these levels are not in severity/verbosity order
+  // for backwards compatibility reasons.
   enum Level {
     DEBUG = 0;
     INFO = 1;
     ERROR = 2;
+    WARN = 3;
+    TRACE = 4;
```
DomBlack, line 260: These don't align with the values here: https://github.com/encoredev/encore/pull/488/files#diff-b72206509fc4d8826ad43ce29161b64ba6519a99dd35b6c03686ff7cd10ec3bcR779-R788

or 
https://github.com/encoredev/encore/pull/488/files#diff-a0ed25f6333985d85f1536565cca801ffb885f078e5a2206cb3b581963682a4bR22-R28

DomBlack: 👍 Could we add a comment to that effect please? The existing comment wasn't super clear as to the why

DomBlack (approved):

PR #492, opened 2022-10-27
runtime/appruntime: start services immediately

eandre:
Instead of waiting for incoming requests, run all
service init functions immediately on startup.


compiler/internal/codegen/codegen_main.go
```diff
@@ -142,6 +144,21 @@ func (b *Builder) computeHandlerRegistrationConfig(mwNames map[*est.Middleware]*
 	})
 }
 
+func (b *Builder) computeServiceInitConfig() *Statement {
+	return Index().Qual("encore.dev/appruntime/service", "Initializer").CustomFunc(Options{
+		Open:      "{",
+		Close:     "}",
+		Separator: ",",
+		Multi:     true,
+	}, func(g *Group) {
```
DomBlack, line 153: I'd prefer we use `Values` rather `custom` for these types of blocks, and then construct the `[]Code` for the values

DomBlack (commented):
BTW this leads to an interesting problem, that apps with services which startup immediately will require 1 min instance and always allocated CPU.

Because services with no API endpoints and no pubsub subscribers means that service is polling external systems for events

eandre:
This is only really changing that we invoke service initializers immediately on container startup. It's definitely true if you have an application that needs to run all the time that it needs Always On CPU and Min Instances configured in Cloud Run, but that's a separate question I think

compiler/internal/codegen/codegen_main.go
eandre: This is ensuring we have newlines for each entry rather than putting all on the same line. Or is there a better way to accomplish that?

Willyham:
> Because services with no API endpoints and no pubsub subscribers means that service is polling external systems for events

This might not be useful but there are some other use cases here as well. There might be a service which starts up, does some computation and stores the result in some store/external system. For example, I might be writing a service which continually searches a hash space for collisions or something. My point is that I don't think you can use the inputs/outputs to determine which is "running" as a service or not.

I think in the world where everything is built into Encore you _could_ do that, because a service is probably useless if it didn't have _any_ inputs or outputs. But while Encore lets you call out to other things that it doesn't support natively (like different store types, external apis, etc) and doesn't know about those through static analysis, then you have to assume that the service is doing something useful and should probably run.

Willyham:
There's also the question here about lifecycle management. I'm speculating without knowing how this works under the hood, but I assume an error in the initialiser will now cause the whole app to not start? Do we need to run any clean up stuff or not have retries in that case?

Also, are all the encore primitives like logging/metrics available at the time this function is invoked? Will logs definitely be captured and flushed? Even if I return an error from init?

DomBlack (approved):

PR #494, opened 2022-10-28
pkg/ansi: fix cursor save/restore

eandre:
It turns out the terminal escape codes we used were not
supported by all terminals, notably macOS's Terminal.app.

Fix this by using a different terminal escape code that does work.

### DIFF
--- a/e2e-tests/testdata/echo/go.mod
+++ b/e2e-tests/testdata/echo/go.mod
@@ -2,7 +2,6 @@ module encore.app
 
 go 1.18
 
-require "encore.dev" v1.1.0
+require "encore.dev" v1.9.3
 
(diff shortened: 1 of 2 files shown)


e2e-tests/testdata/echo/go.mod
```diff
@@ -2,7 +2,6 @@ module encore.app
 
 go 1.18
 
-require "encore.dev" v1.1.0
-
-replace encore.dev => ../../../../../runtime
+require "encore.dev" v1.9.3
 
+replace encore.dev => ../../../runtime
```
eandre, line 7: This is just unrelated cleanup, the replace directive needed to be updated since we moved the `e2e-tests` directory

DomBlack (approved):

PR #496, opened 2022-10-28
Add support for experiments

DomBlack:
This commit adds the ability to Encore to have features behind experiment flags which can be opted into by either Environmental Variables _or_ by the `encore.app` file.

All parser features can either define `experiment.None`, which enables those features for all users or create a new experiment name to lock the behaviour behind the feature flag.

Both the parser and compiler have access to the set of `experiments` enabled, so can also switch behaviour based on them.

By using these, we can enable trunk based development for Encore without worrying brand new features will cause existing apps to break during the development cycle.


eandre:
This was continued in #509 so closing this

PR #509, opened 2022-11-03
Introduce experiments API

eandre:
This adds the ability to define experiments that can be conditionally enabled
either on a per-app level (by adding an "experiments" field to encore.app) or by setting an environment variable.

This functionality will be used to simplify trunk-based development of new features.

Thanks to @DomBlack for doing this work, this is a continuation of #496.


cli/daemon/run/tests.go
```diff
@@ -29,6 +31,15 @@ import (
 func (mgr *Manager) Check(ctx context.Context, appRoot, relwd string, codegenDebug bool) (buildDir string, err error) {
 	vcsRevision := vcs.GetRevision(appRoot)
 
+	exp, err := appfile.Experiments(appRoot)
+	if err != nil {
+		return "", err
+	}
+	expSet, err := experiments.NewSet(exp, nil)
```
ekerfelt, line 38: Seems like we're repeating these two everywhere. Would it make sense to move appfile.Experiments into experiments.NewSet instead?

eandre: Not sure, it's not *everywhere*, so I don't think they should be unconditionally combined. Maybe makes sense to add a helper function `experiments.FromAppfile()` or something?

internal/experiments/experiment.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (s *Set) add(keys ...Name) error {
+	for _, key := range keys {
+		if key == "" {
+			continue
+		}
+
+		if !key.Valid() {
+			return &UnknownExperimentError{key}
```
ekerfelt, line 90: hmm, maybe we should log rather than err? What if we deprecate a test, i.e. remove it from the experiments list. Will all apps having had that test fail until they've removed the flag.

ekerfelt (approved):

cli/daemon/run/tests.go
ekerfelt: figured since the first parameter of newSet is called fromAppFile it was only used in combination. But yeah, maybe helper would make sense

ekerfelt (approved):

internal/experiments/experiment.go
eandre: I think that's safer? If you rely on the functionality it would be potentially dangerous to suddenly start behaving differently

ekerfelt: true. Probably best to play it safe

cli/daemon/run/tests.go
eandre: Yeah, since I have a stacked PR on top I'll be lazy and defer changing this for now :)

PR #512, opened 2022-11-04
cli/daemon/engine/trace: fix error handling

eandre:
Trace version 9 emits the error message twice.
Handle this on the parser. We'll fix the duplication
in the next trace version.

### DIFF
--- a/cli/daemon/engine/trace/trace.go
+++ b/cli/daemon/engine/trace/trace.go
@@ -431,9 +431,12 @@ func (tp *traceParser) requestEnd(ts uint64) error {
 		errMsg := tp.ByteString()
 		if len(errMsg) > 0 {
 			req.Err = errMsg
-			if tp.version >= 5 {
-				req.ErrStack = tp.stack(filterNone)
-			}
+
+			// Version 9 has a spurious duplicated error message.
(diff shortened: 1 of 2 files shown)


cli/daemon/engine/trace/trace.go
```diff
@@ -431,9 +431,12 @@ func (tp *traceParser) requestEnd(ts uint64) error {
 		errMsg := tp.ByteString()
 		if len(errMsg) > 0 {
 			req.Err = errMsg
-			if tp.version >= 5 {
-				req.ErrStack = tp.stack(filterNone)
-			}
+
+			// Version 9 has a spurious duplicated error message.
+			// Ignore it.
+			_ = tp.String()
```
ekerfelt, line 437: should we restrict this to v9?

ekerfelt (approved):

cli/daemon/engine/trace/trace.go
eandre: v9 is the latest version and it didn't seem like something worth creating a new version over. I don't want to make it conditional on `== 9` before we actually add the fix though, in case we forget about it or something.

PR #514, opened 2022-11-06
Disable JavaScript client generation for now

Minivera:
When I was investigating to make sure the JavaScript client was the same as the TYpeScript client, I noticed that the JS PR was merged before the JS client generation was properly disabled. According to the comment on my initial PR, I've created this follow up to make sure it's disabled if it's deemed necessary.

### DIFF
--- a/internal/clientgen/javascript.go
+++ b/internal/clientgen/javascript.go
@@ -559,7 +559,7 @@ if (authData) {
 					w.WriteString("\n")
 				}
 			} else {
-				w.WriteString("init.headers[\"Authorization\"] = \"Bearer \" + authData\n")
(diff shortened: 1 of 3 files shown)


internal/clientgen/javascript.go
```diff
@@ -559,7 +559,7 @@ if (authData) {
 					w.WriteString("\n")
 				}
 			} else {
-				w.WriteString("init.headers[\"Authorization\"] = \"Bearer \" + authData\n")
+				w.WriteString("init.headers[\"x-api-key\"] = \"Bearer \" + authData\n")
```
eandre, line 562: Is this change intended? If so why?

eandre (approved):

PR #516, opened 2022-11-07
Add `encore alpha exec`

eandre:
This adds a (pre-release version of) a new command (currently `encore alpha exec`) that adds a way to execute one-off scripts against an Encore application running locally.

Usage:
`encore alpha exec path/to/command [...args]`
`encore alpha exec path/to/main.go [...args]` (single file)

The command directory should contain Go files with `package main` with a `main` function. The additional args are passed directly to the built binary.


compiler/exec_script.go
```diff
@@ -0,0 +1,236 @@
...
+				mainFile, mainFunc = f, fd
+				break FileLoop
+			}
+		}
+	}
+	if mainFunc == nil {
+		return fmt.Errorf("unable to find main function in %s", mainPkgPath)
+	}
+
+	//rw := rewrite.New(mainFile.Contents, mainFile.Token.Base())
```
ekerfelt, line 169: delete?

ekerfelt (approved):

ekerfelt (approved):

PR #522, opened 2022-11-21
Add Prometheus exporter

campesel:
This PR adds a Prometheus exporter for custom metrics, so that we can push metrics to any system that supports Prometheus remote write.


runtime/go.mod
```diff
@@ -24,9 +24,10 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nsqio/go-nsq v1.1.0
+	github.com/prometheus/prometheus v0.40.2
```
eandre, line 27: Prometheus is a pretty big dependency and we only use the protobuf part. Could we vendor those files instead?

PR #525, opened 2022-11-23
Store external request id in traces

eandre:
This adds support for storing any incoming `X-Request-Id` in the trace,
to allow for easier matching of requests sent from frontends.


runtime/appruntime/api/server.go
```diff
@@ -191,7 +192,9 @@ func (s *Server) register(reg HandlerRegistration, logRegistration bool) {
 
 		adapter := func(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
 			params := toUnnamedParams(ps)
-			s.processRequest(h, s.NewIncomingContext(w, req, params, model.AuthInfo{}))
+			traceID, _ := model.GenTraceID()
+			w.Header().Set("X-Encore-Trace-ID", traceID.String())
```
DomBlack, line 196: I think we need to echo back the `X-Request-Id` based on this: https://http.dev/x-request-id

> When receiving a token for the HTTP X-Request-ID request header, any reverse proxies are recommended to pass the identifier to the backend servers, which in turn are recommended to include it in the HTTP response generated and send the token back to the client.

I think also if one wasn't sent, we should respond with this Trace ID in that field, rather than our own Trace ID field.

DomBlack (approved):

PR #526, opened 2022-11-23
[ENC-1069] Support Private Network Access in CORS

DomBlack:
As of Chrome 107, a new CORS header is present if the requested resource is on a private network: `Access-Control-Allow-Private-Network: true` The response _must_ include this header as well for the preflight to pass.

This commit adds support for this only on applications running locally (i.e. run through `encore run`).

This commit also adds support for `encore.app` files to declare a list of allowed headers which should be permitted on the application. Previously this was configuration within the Encore platform but did not affect local development.

By adding the following to `encore.app`, an app can opt into allowing the header `X-Test-Header`:

```json
{
    "global_cors": {
        "allow_headers": [
            "X-Test-Header",
        ],
    },
}
```

### DIFF
--- a/cli/internal/appfile/appfile.go
+++ b/cli/internal/appfile/appfile.go
@@ -31,6 +31,19 @@ type File struct {
 	//
 	// Do not use these features in production without consulting the Encore team.
 	Experiments []experiments.Name `json:"experiments,omitempty"`
+
+	// Configure global CORS settings for the application which
+	// will be applied to all API gateways into the application.
+	GlobalCORS *CORS `json:"global_cors,omitempty"`
+}
+
+type CORS struct {
+	// Debug is a flag to enable debug logging for CORS
+	Debug bool `json:"debug,omitempty"`
+
+	// AllowHeaders allows an app to specify additional headers that should be
+	// accepted by the app
+	AllowHeaders []string `json:"allow_headers"`
 }
 
 // Parse parses the app file data into a File.
@@ -75,4 +88,13 @@ func Experiments(appRoot string) ([]experiments.Name, error) {
 		return nil, err
 	}
 	return f.Experiments, nil
+}
+
+// GlobalCORS returns the global CORS settings for the app located
+func GlobalCORS(appRoot string) (*CORS, error) {
+	f, err := ParseFile(filepath.Join(appRoot, Name))
+	if err != nil {
+		return nil, err
+	}
+	return f.GlobalCORS, nil
 }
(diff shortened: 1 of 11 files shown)


cli/internal/appfile/appfile.go
```diff
@@ -31,6 +31,19 @@ type File struct {
...
+	// will be applied to all API gateways into the application.
+	GlobalCORS *CORS `json:"global_cors,omitempty"`
+}
+
+type CORS struct {
+	// Debug is a flag to enable debug logging for CORS
+	Debug bool `json:"debug,omitempty"`
+
+	// AllowHeaders allows an app to specify additional headers that should be
+	// accepted by the app
```
eandre, line 45: Does this support wildcard (`*`)? Probably worth doing and documenting

runtime/appruntime/config/config.go
```diff
@@ -81,6 +84,12 @@ type CORS struct {
 	// the default set of {"Origin", "Authorization", "Content-Type"}.
 	// As a special case, if the list contains "*" all headers are allowed.
 	ExtraAllowedHeaders []string `json:"raw_allowed_headers,omitempty"`
+
+	// AllowAccessWhenOnPrivateNetwork, if true, allows requests to Encore apps running
+	// on private networks from websites.
+	//
+	// See: https://wicg.github.io/private-network-access/
+	AllowAccessWhenOnPrivateNetwork bool `json:"allow_private_network_access,omitempty"`
```
eandre, line 92: The field name is different from the JSON key, make them consistent? I think the JSON field is clearer

runtime/appruntime/cors/cors.go
```diff
@@ -23,12 +29,14 @@ func Options(cfg *config.CORS) cors.Options {
 	hasWildcardOriginWithoutCreds := cfg.AllowOriginsWithoutCredentials == nil || sortedSliceContains(originsWithoutCreds, "*")
 	hasUnsafeWildcardOriginWithCreds := sortedSliceContains(originsCreds, config.UnsafeAllOriginWithCredentials)
 
-	allowedHeaders := append([]string{"Authorization", "Content-Type"}, cfg.ExtraAllowedHeaders...)
+	allowedHeaders := append([]string{"Origin", "Authorization", "Content-Type"}, cfg.ExtraAllowedHeaders...)
```
eandre, line 32: Is this needed? Seems surprising, how could CORS work without Origin headers?

eandre: I'd rather fix the docs; it's surprising to have code which is misleading or doesn't have a purpose

eandre (approved):

PR #528, opened 2022-11-23
Fix the JavaScript base auth else and add a test case

Minivera:
In a previous PR, I unfortunately had changed the JavaScript auth generation in a way that would have broken the base auth. I've fixed that and added a test case for base auth to make sure that line is covered and tested.


Minivera:
Not entirely sure why CI is failing here? The tests pass perfectly on my end without issue and I haven't touched how the middleware is generated in the TS files. Could I be missing something here? @eandre

eandre (approved):

PR #529, opened 2022-11-24
Add experiments to app metadata

campesel:
This PR adds the list of enabled experiments to the app metadata, so that the platform can compute the app config accordingly.


internal/experiments/experiment.go
```diff
@@ -17,12 +17,16 @@ const (
 	// LocalSecretsOverride is an experiment to allow for secrets
 	// to be overridden with values from a ".secrets.local" file.
 	LocalSecretsOverride Name = "local-secrets-override"
+
+	// DefaultEncoreMetrics is an experiment to enable default Encore metrics.
+	DefaultEncoreMetrics Name = "default-encore-metrics"
```
eandre, line 22: This PR is more self contained if you remove this experiment from here, and then introduce it in the Metrics PR.

I would also rename the experiment to just Metrics while you're at it, since we're working on custom metrics at the same time

eandre (approved):

PR #531, opened 2022-11-24
[ENC-1072] Fix over watching of the filesystem

DomBlack:
The new FSnotify library we introduced in v1.10.5 had a bug where it would recursively recursive and watch the file system (i.e. we expontentaionally watch the file system).

This resulted in two side effects:
 - Our attempt to ignore `node_modules` got bypassed, so we ended up watching everything inside node_modules
 - We spent a lot of time going over the same folders again and again

This has been fixed by moving the main logic into the `WalkDir` call within the watch library.

Then we've modified the `walkDir` in the parser to also ignore the same set of folders, so we walk the app as quickly as possible.

This commit also updates the `eerror` package and intergrates it better with zerolog, so if an error with a stack trace  or an error with meta data is logged, we will extract them and add the meta data as new parameters to the log line, and add the stack trace under the log line. (This change will also work for `srcerrors` stack traces too as they both implement the same stack trace interface)

Screenshot 2022-11-24 at 15 02 10

### DIFF
--- a/cli/cmd/encore/daemon/daemon.go
+++ b/cli/cmd/encore/daemon/daemon.go
@@ -36,11 +36,15 @@ import (
 	"encr.dev/cli/daemon/sqldb/external"
 	"encr.dev/cli/internal/xos"
 	"encr.dev/internal/conf"
+	"encr.dev/pkg/eerror"
+	"encr.dev/pkg/watcher"
 	daemonpb "encr.dev/proto/encore/daemon"
 )
 
 // Main runs the daemon.
 func Main() {
+	watcher.BumpRLimitSoftToHardLimit()
+
 	if err := redirectLogOutput(); err != nil {
 		log.Error().Err(err).Msg("could not setup daemon log file, skipping")
 	}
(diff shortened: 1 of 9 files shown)


cli/cmd/encore/daemon/daemon.go
```diff
@@ -36,11 +36,15 @@ import (
...
 	"encr.dev/cli/internal/xos"
 	"encr.dev/internal/conf"
+	"encr.dev/pkg/eerror"
+	"encr.dev/pkg/watcher"
 	daemonpb "encr.dev/proto/encore/daemon"
 )
 
 // Main runs the daemon.
 func Main() {
+	watcher.BumpRLimitSoftToHardLimit()
```
DomBlack, line 46: Go bumps the hard limits, but not the soft limits, so I've added this as a satefy net, even though the rest of the PR means we don't need it in the cases we've identified so far

pkg/eerror/error.go
```diff
@@ -83,7 +81,15 @@ func WithMeta(err error, meta map[string]any) error {
...
-	return fmt.Sprintf("%s: %s", e.Module, e.Message)
+	if e.cause != nil {
+		cause := e.cause.Error()
+
+		// Remove the module prefix if it's the same
+		cause = strings.TrimPrefix(cause, "["+e.Module+"]: ")
+
+		return fmt.Sprintf("[%s]: %s: %s", e.Module, e.Message, cause)
+	}
+	return fmt.Sprintf("[%s]: %s", e.Module, e.Message)
```
DomBlack, line 92: This was because we didn't show the wrapped error before, which was annouying

PR #532, opened 2022-11-24
[ENC-1073] Automatically add headers used in types to CORS config

DomBlack:
This commit automatically adds any headers used within data types used within the Encore application to be added to the list of accepted headers used by the CORS middleware within the Encore runtime.

### DIFF
--- a/compiler/internal/codegen/codegen_main.go
+++ b/compiler/internal/codegen/codegen_main.go
@@ -130,8 +130,8 @@ func (b *Builder) computeCORSHeaders() (Code, error) {
 
 	// Walk all the structures looking for headers
 	decls := b.res.Meta.Decls
-	for _, decl := range decls {
-		err := schema.Walk(decls, decl, func(node any) error {
+	for _, param := range b.res.App.ParamTypes() {
+		err := schema.Walk(decls, param, func(node any) error {
 			switch n := node.(type) {
 			case *schema.Struct:
 				for _, field := range n.Fields {
--- a/compiler/internal/codegen/codegen_main.go
+++ b/compiler/internal/codegen/codegen_main.go
@@ -2,14 +2,18 @@ package codegen
 
 import (
 	"fmt"
+	"net/http"
 	"path"
+	"sort"
 
 	. "github.com/dave/jennifer/jen"
 
 	"encr.dev/internal/gocodegen"
 	"encr.dev/parser"
 	"encr.dev/parser/est"
+	"encr.dev/pkg/eerror"
 	"encr.dev/pkg/errlist"
+	schema "encr.dev/proto/encore/parser/schema/v1"
 )
 
 const JsonPkg = "github.com/json-iterator/go"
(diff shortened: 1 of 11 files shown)


compiler/internal/codegen/codegen_main.go
```diff
@@ -113,6 +123,49 @@ func (b *Builder) computeStaticPubsubConfig() Code {
...
 }
 
+func (b *Builder) computeCORSHeaders() (Code, error) {
+	// Find all used headers
+	usedHeaderSet := map[string]struct{}{}
+	usedHeaders := []string{}
+
+	// Walk all the structures looking for headers
+	decls := b.res.Meta.Decls
+	for _, decl := range decls {
```
eandre, line 133: Better to start from APIs, and only check request structs for non-private APIs? Otherwise this will end up adding a bunch of extraneous headers (such as for private endpoints, response headers, etc)

DomBlack: What about middleware?

That's why I started at decls

eandre: Middleware cannot add headers I don't believe? Auth handlers can so they need to be checked also

DomBlack: I meant Auth handlers :P

parser/est/est.go
```diff
@@ -27,6 +27,31 @@ type Application struct {
 	Middleware    []*Middleware
 }
 
+// ParamTypes returns all used parameter types used for input to an Encore application
+func (a *Application) ParamTypes() (rtn []*schema.Type) {
```
eandre, line 31: Nit: unnecessary named return

parser/est/est.go
```diff
@@ -27,6 +27,31 @@ type Application struct {
...
+	for _, s := range a.Services {
+		for _, rpc := range s.RPCs {
+			if rpc.Request != nil {
+				used[rpc.Request.Type] = struct{}{}
+			}
+		}
+	}
+
+	// Convert the set to an array
+	for typ := range used {
```
eandre, line 49: Probably good to return things in a determinstic order. Either sort the `rtn` slice or append to it as we iterate over `Services` and `RPCs` (which are already in deterministic order).

parser/est/est.go
```diff
@@ -27,6 +27,30 @@ type Application struct {
 	Middleware    []*Middleware
 }
 
+// ParamTypes returns all used parameter types used for input to an Encore application
+func (a *Application) ParamTypes() (rtn []*schema.Type) {
+	used := make(map[*schema.Type]struct{})
+
+	// Check the auth handler
+	if a.AuthHandler != nil && a.AuthHandler.Params != nil {
+		used[a.AuthHandler.Params] = struct{}{}
```
eandre, line 36: This doesn't add to `rtn`

eandre (approved):

eandre (approved):

PR #534, opened 2022-11-25
Add custom metrics as code concept

campesel:
This PR adds the code to support custom metrics as a code concept. See `metrics.md` for examples of how to use them.


proto/encore/parser/meta/v1/meta.proto
```diff
@@ -287,7 +288,31 @@ message CacheCluster {
...
+    string         doc          = 4;
+    Path           path_pattern = 5;
   }
 }
+
+message Metric {
+  string            name         = 1; // the name of the metric
+  schema.v1.Builtin value_type   = 2;
+  string            doc          = 3; // the doc string
+  MetricKind        kind         = 4;
```
campesel, line 303: @eandre I renamed this field to `kind` to differentiate it from the metric value type.

parser/feature_metrics.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		var metricKind meta.Metric_MetricKind
+		switch con.FuncName {
+		case "NewCounter", "NewCounterGroup":
+			metricKind = meta.Metric_COUNTER
+		case "NewGauge", "NewGaugeGroup":
+			metricKind = meta.Metric_GAUGE
+		default:
+			panic("unknown metric constructor name")
+		}
```
campesel, line 76: @eandre Should I add the metric kind to `con` rather than switching on the function name to get it?

parser/feature_metrics.go
```diff
@@ -0,0 +1,213 @@
...
+			ConfigLit: cfg.Lit(),
+		}
+
+		// Resolve labels
+		if con.HasLabels {
+			if _, isPtr := labelTypeArg.(*ast.StarExpr); isPtr {
+				p.errInSrc(srcerrors.MetricLabelsIsPointer(p.fset, typeArgs[0], "metrics."+con.FuncName))
+				return nil
+			}
+			metric.LabelsType = p.resolveType(file.Pkg, file, typeArgs[0], nil)
```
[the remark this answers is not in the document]
campesel: It is used to in `codegen_metrics.go`. Should we change that code to use the label slice instead?

eandre (approved):

PR #539, opened 2022-12-07
Implement Encore Cloud metrics exporter

campesel:
The platform prefixes an Encore app metrics with their resource ID. To be able to export metrics to Encore Cloud, the runtime needs to know what the metric names in Encore Cloud are: these are injected into the runtime config by the platform, so that the runtime can use them to export metrics to Encore Cloud.

### DIFF
--- a/runtime/appruntime/metrics/gcp/cloud_monitoring.go
+++ b/runtime/appruntime/metrics/gcp/cloud_monitoring.go
@@ -31,6 +31,13 @@ func New(svcs []string, cfg *config.GCPCloudMonitoringProvider, rootLogger zerol
 	}
 }
 
+func NewEncoreCloudExporter(svcs []string, cfg *config.GCPCloudMonitoringProvider, rootLogger zerolog.Logger, metricNames map[string]string) *Exporter {
+	exp := New(svcs, cfg, rootLogger)
+	exp.exportsToEncoreCloud = true
+	exp.encoreCloudMetricNames = metricNames
+	return exp
+}
+
 type Exporter struct {
 	svcs       []string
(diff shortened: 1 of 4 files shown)


runtime/appruntime/metrics/gcp/cloud_monitoring.go
```diff
@@ -109,7 +119,11 @@ func (x *Exporter) getMetricData(newCounterStart, endTime time.Time, collected [
 		}
 
 		svcNum := m.Info.SvcNum()
-		metricType := "custom.googleapis.com/" + m.Info.Name()
+		encoreCloudMetricName, ok := x.encoreCloudMetricNames[m.Info.Name()]
+		if !ok {
+			continue
```
eandre, line 124: Is this expected to happen, or is it guarding against something that never should happen unless something is wrong? If the latter it would be useful to log an error like we do above

runtime/appruntime/metrics/gcp_cloud_monitoring_exporter.go
```diff
@@ -11,10 +11,14 @@ func init() {
 	registerProvider(providerDesc{
```
eandre, line 11: Would it be cleaner to treat Encore Cloud as a separate provider, instead of overloading it? Not sure, up to you

runtime/appruntime/metrics/gcp/cloud_monitoring.go
campesel: This should never happen. I'll add a log line 👍

eandre (approved):

PR #544, opened 2022-12-09
pkg/pgproxy: fix ssl negotiation

eandre:
If the client requested SSL but the server does not support
SSL we reject the SSL request. However, instead of reading
a new startup message we would incorrectly proceed to
actually doing the SSL handshake.

Thanks Cilla for the report.


DomBlack (approved):
Nice simple change :-)

PR #550, opened 2022-12-18
Add *sqldb.Driver() to get underlying db driver

eandre:
This adds a mechanism to get the underlying database driver.
With this change Encore commits to not changing the underlying
database driver for the foreseeable future, and not without providing
a graceful migration path to a hypothetical new driver.

This is stacked on top of #543 so that should be merged first.


runtime/storage/sqldb/db.go
```diff
@@ -262,3 +261,17 @@ func (db *Database) Begin(ctx context.Context) (*Tx, error) {
 
 	return &Tx{mgr: db.mgr, txid: txid, std: tx}, nil
 }
+
+// Driver returns the underlying database driver for this database connection pool.
+//
+// It is currently of type [*pgxpool.Pool] and it's safe to immediately do a type conversion
+// to that type:
+//
+//	var db = sqldb.Named("mydatabase").Driver().(*pgxpool.Pool)
```
DomBlack, line 270: I think the example should include a if guard
```suggestion
//	var db, ok = sqldb.Named("mydatabase").Driver().(*pgxpool.Pool)
//      if !ok {
//         panic("driver is not pgx")
/       }
```

DomBlack (approved):

DomBlack (approved):

PR #553, opened 2022-12-19
Add GraphQL Operation Tracing

DomBlack:
This commit adds the concept of a GraphQL operation event to our tracing implementation. This will allow the trace list view to show the name of the query, mutation or subscription to be shown rather than just the GraphQL endpoint name.

To achieve this, we attempt to parse a raw endpoints request body as a GraphQL query document (using the same parser that `gqlgen` uses). If this parse succeeds, we assume that the endpoint is processing GraphQL and add the trace data


cli/daemon/dash/dashapp/src/components/app/AppTraces.tsx
```diff
@@ -85,11 +85,34 @@ export default class AppTraces extends React.Component<Props, State> {
...
+
+                    switch (graphQLOp[0].type) {
+                      case "QUERY":
+                        endpoint = "query " + graphQLOp[0].name
+                        break
+                      case "MUTATION":
+                        endpoint = "mutation " + graphQLOp[0].name
+                        break
+                      case "SUBSCRIPTION":
+                        endpoint = "subscription " + graphQLOp[0].name
```
eandre, line 112: Add a `break` for safety in case we add more cases

runtime/appruntime/api/capture.go
```diff
@@ -19,9 +20,12 @@ type rawRequestBodyCapturer struct {
...
 
-	bufMu      sync.Mutex
-	buf        *bytes.Buffer
-	overflowed bool // whether the buffer overflowed
+	bufMu       sync.Mutex
+	buf         *bytes.Buffer
+	overflowed  bool        // whether the buffer overflowed
+	httpMethod  string      // The HTTP method of the request
+	headers     http.Header // The headers of the request
+	queryString url.Values  // The query string of the request captured
```
eandre, line 28: Pretty sure these are all available through other means already, why do they need to be added here?

DomBlack: I don't think it is for raw endpoints

marcuskohlberg:
@DomBlack Do we need to print the greyed "graphql.Query"? Does the icon convey enough information already?
If the information is not critical, we should omit it since it makes the interaction weird when the links are not in a straight line?

DomBlack:
> @DomBlack Do we need to print the greyed "graphql.Query"? Does the icon convey enough information already? If the information is not critical, we should omit it since it makes the interaction weird when the links are not in a straight line?

I think we do; as there's no guarantee the app has only one graph endpoint, so you might want to know _which_ endpoint is the one being used. For example if you had a `public` GraphQL endpoint and an `internal tooling` one for your staff to use

marcuskohlberg:
> > @DomBlack Do we need to print the greyed "graphql.Query"? Does the icon convey enough information already? If the information is not critical, we should omit it since it makes the interaction weird when the links are not in a straight line?
> 
> I think we do; as there's no guarantee the app has only one graph endpoint, so you might want to know _which_ endpoint is the one being used. For example if you had a `public` GraphQL endpoint and an `internal tooling` one for your staff to use

In that case maybe make the whole thing a link?

DomBlack:
It is a link, I just removed the underline from it and faded it out, can put the line back in, but the line will be solid black when under the endpoint name - or we could just make the entire endpoint solid black - I only did the opacity on the endpoint name to remove the focus from it somewhat

DomBlack:
Closing for now; we'll come back to this in the future - but with the large amount of refactoring we've done since this isn't really mergable any more

PR #557, opened 2022-12-19
internal/clientgen: fix fetch use

### LINKED ISSUE #556: Generated TypeScript client doesn't work with built-in fetch out of the box
The built-in fetch function isn't bound correctly when using the generated TypeScript client. It needs to be bound using `fetch.bind(self)`. But what's more, it doesn't work correctly when running in server-side Node environments. A better way to fix this is to depend on a library like `cross-fetch` that provides an abstraction that automatically selects the correct `fetch` implementation.

eandre:
Properly use `fetch.bind(this)` to bind it to the `Window` object. Improve a few things while we're at it, like adding a `PreviewEnv` helper.

Fixes #556


internal/clientgen/javascript.go
```diff
@@ -80,6 +81,7 @@ func (js *javascript) Generate(buf *bytes.Buffer, appSlug string, md *meta.Data)
 	}
 
 	js.WriteString("// " + doNotEditHeader() + "\n\n")
+	js.WriteString("import fetch from \"cross-fetch\";\n")
```
DomBlack, line 84: This adds an implicit dependancy which we've not documented in the generate command's help. We should add that if we go which this approach.

On the backend this was solvable without a new library, but by simply applying a bind against the default fetcher function. The issue was we defaulted in a way which apply the wrong prototype when calling the function

```typescript
return new Client(APIServer(), {
    auth: key,
    fetcher: fetch.bind(globalThis), // This is needed in cloudflare workers, otherwise it will report a TypeError: Illegal invocation
  })
```

So I wonder if we can fix this without needing `cross-fetch`

e2e-tests/testdata/echo_client/js/client.js
```diff
@@ -13,6 +13,13 @@ export function Environment(name) {
     return `https://${name}-slug.encr.app`
 }
 
+/**
+ * PreviewEnv returns a BaseURL for calling the preview environment with the given PR number.
+ */
+export function PreviewEnv(pr) {
+    return Environment(`pr${pr}`)
+}
```
DomBlack, line 21: Should we add this function to the Go client too?

DomBlack (approved):

e2e-tests/testdata/echo_client/js/client.js
eandre: Good idea, done

DomBlack (approved):

PR #561, opened 2022-12-22
ENC-1166 - Fix idents part parsing

ekerfelt:
If new parts started with digits, they would be truncated, e.g.

`encore-123abc` -> `["encore","abc"]`


DomBlack (dismissed):

pkg/idents/identifiers.go
```diff
@@ -75,6 +75,9 @@ func parseIdentifier(goIdentifier string) (parts []string) {
 		case unicode.IsLower(r):
 			return lower
 		case unicode.IsDigit(r):
+			if lastType == other {
```
eandre, line 78: Would be nice to add a test case :)

eandre (approved):

PR #567, opened 2023-01-02
Define default metric e_requests_total as custom metric

campesel:
We defined `e_requests_total` as a logs-based metric, but considered the quota limitations of logs-based metrics, we've decided to treat default metrics like `e_requests_total` as custom metrics.

Custom metrics, as opposed to logs-based metrics, have higher quotas in GCP. This makes it easier to use them in Encore Cloud.

### DIFF
--- a/runtime/appruntime/api/server.go
+++ b/runtime/appruntime/api/server.go
@@ -14,11 +14,11 @@ import (
 	encore "encore.dev"
 	"encore.dev/appruntime/config"
 	"encore.dev/appruntime/cors"
-	"encore.dev/appruntime/metrics"
 	"encore.dev/appruntime/model"
 	"encore.dev/appruntime/platform"
 	"encore.dev/appruntime/reqtrack"
 	"encore.dev/beta/errs"
+	"encore.dev/metrics"
 )
 
 type Access string
@@ -59,12 +59,18 @@ type Handler interface {
 	Handle(c IncomingContext)
 }
 
+type RequestsTotalLabels struct {
+	Service  string // Service name.
+	Endpoint string // Endpoint name.
+	Code     string // Human-readable HTTP status code.
+}
+
 type Server struct {
 	cfg            *config.Config
 	rt             *reqtrack.RequestTracker
 	pc             *platform.Client // if nil, requests are not authenticated against platform
 	encoreMgr      *encore.Manager
-	metrics        *metrics.Manager
+	requestsTotal  *metrics.CounterGroup[RequestsTotalLabels, uint64]
 	clock          clock.Clock
 	rootLogger     zerolog.Logger
 	json           jsoniter.API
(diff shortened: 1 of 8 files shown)


runtime/appruntime/api/server.go
```diff
@@ -113,7 +119,7 @@ func NewServer(
 		pc:             pc,
 		rt:             rt,
 		encoreMgr:      encoreMgr,
-		metrics:        metrics,
+		requestsTotal:  metrics.NewCounterGroupInternal[RequestsTotalLabels, uint64]("e_requests_total", metrics.CounterConfig{}, reg),
```
eandre, line 122: Can you extract this into a variable?

runtime/metrics/metrics.go
```diff
@@ -64,6 +64,11 @@ func NewCounterGroup[L Labels, V Value](name string, cfg CounterConfig) *Counter
 	return newCounterGroup[L, V](Singleton, name, cfg)
 }
 
+//publicapigen:drop
+func NewCounterGroupInternal[L Labels, V Value](name string, cfg CounterConfig, reg *Registry) *CounterGroup[L, V] {
```
eandre, line 68: Move the registry to the first argument

runtime/appruntime/api/server.go
```diff
@@ -59,12 +59,18 @@ type Handler interface {
 	Handle(c IncomingContext)
 }
 
+type RequestsTotalLabels struct {
+	Service  string // Service name.
+	Endpoint string // Endpoint name.
+	Code     string // Human-readable HTTP status code.
```
eandre, line 65: Should this be an int? Do we support that?

campesel: `Code` is a string like `ok` or `not_found`. If we can't convert a HTTP code into a human-readable string, we set `Code` to `http_X` (e.g. `http_418`).

runtime/appruntime/api/server.go
```diff
@@ -88,11 +94,21 @@ func NewServer(
...
 ) *Server {
+	requestsTotal := metrics.NewCounterGroupInternal[RequestsTotalLabels, uint64](reg, "e_requests_total", metrics.CounterConfig{
+		EncoreInternal_LabelMapper: func(labels RequestsTotalLabels) []metrics.KeyValue {
+			return []metrics.KeyValue{
+				{Key: "service", Value: labels.Service},
+				{Key: "endpoint", Value: labels.Endpoint},
+				{Key: "code", Value: labels.Code},
+			}
+		},
+	})
```
campesel, line 110: I haven't set `EncoreInternal_SvcNum` as this metric is shared by all services. My understanding is that the service number is read from the request tracker.

eandre (approved):

runtime/appruntime/api/server.go
campesel: That's the case where it's used.

PR #577, opened 2023-01-12
Change API Explorer to use JSON field names and JSON output

simon-johansson:
Main change: 
* Request input (headers, query string and body) are now in the format in which the user would interact with the endpoint and not the GO request struct.

**Before**
Screenshot 2023-01-12 at 16 49 47

**New**
Notice also that we are now showing the URL even when it does not have dynamic path params

Screenshot 2023-01-12 at 16 49 21

Other changes:
* Always show the request path even when the path does not have dynamic parts (see screenshot above)
* Show the request url in the tracing view
* Replace the JSON viewer to the same lib that we use on app.encore.dev

**Old JSON viewer**
Screenshot 2023-01-12 at 16 50 10

**New JSON viewer**
Screenshot 2023-01-12 at 16 50 04

### DIFF
--- a/cli/daemon/dash/dashapp/src/components/api/RPCCaller.tsx
+++ b/cli/daemon/dash/dashapp/src/components/api/RPCCaller.tsx
@@ -8,15 +8,16 @@ import Input from "~c/Input";
 import { decodeBase64, encodeBase64 } from "~lib/base64";
 import JSONRPCConn from "~lib/client/jsonrpc";
 import { copyToClipboard } from "~lib/clipboard";
-import { APIMeta, PathSegment, RPC, Service } from "./api";
+import { APIEncoding, APIMeta, ParameterEncoding, PathSegment, RPC, Service } from "./api";
 import CM from "./cm/CM";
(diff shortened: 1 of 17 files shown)


cli/daemon/dash/dashapp/src/components/api/RPCCaller.tsx
```diff
@@ -696,3 +613,119 @@ const RPCPathEditor = React.forwardRef<
...
+  const { authorization } = apiEncoding;
+  if (authorization.query_parameters?.length || authorization.header_parameters?.length) {
+    const queryParams = authorization.query_parameters ?? [];
+    const headerParams = authorization.header_parameters ?? [];
+    processStruct([...queryParams, ...headerParams], authBody);
+  }
+
+  reqBody = JSON.stringify(newBody);
+
+  const defaultMethod = hasBody ? "POST" : "GET";
```
eandre, line 709: I think we should use `reqEncoding.http_methods[0]` instead

eandre (approved):

PR #578, opened 2023-01-12
Fix no-SSL Client Authentication

cezary-marczak:


eandre:
Hey @cezary-marczak, thanks very much for the contribution! It looks like you and I independently fixed this (I committed this early December in #544). Are you running the latest release of Encore?

PR #579, opened 2023-01-13
docs: Update documentation to build from source

sammous:
When building from source, the `git-remote-encore` pluging needs to be manually installed to be able to clone projects (`encore app clone`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ This also works just as well with [Visual Studio Code's Remote Development](http
 
 
 ## Building the encore command from source
-To build from source, [build the dashboard](#Building-the-Development-Dashboard) and simply run `go build ./cli/cmd/encore` and `go build ./cli/cmd/git-remote-encore`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,7 +15,7 @@ This also works just as well with [Visual Studio Code's Remote Development](http
 
 
 ## Building the encore command from source
-To build from source, [build the dashboard](#Building-the-Development-Dashboard) and simply run `go build ./cli/cmd/encore`.
+To build from source, [build the dashboard](#Building-the-Development-Dashboard) and simply run `go build ./cli/cmd/encore` and `go build ./cli/cmd/git-remote-encore`.
```
eandre, line 18: `git-remote-encore` doesn't really require ongoing development so it might be better to suggest to `go install` it, so it ends up in `$PATH` automatically

eandre (commented):
Thanks! One minor comment and then this should be good to go

CONTRIBUTING.md
sammous: yes my bad 🙇🏽

eandre (approved):
Thakn you!

PR #582, opened 2023-01-17
Remove per-service time series slice

campesel:
Currently, we store time series values in a slice where the value at position `i` is for service `i`. This PR simplifies how we store time series by only storing the latest value for a metric (instead of the latest value per service).

If a custom metric is declared within a service, we keep track of that so that it's easy to automatically add a `service` label to a time series when exporting metrics.


campesel:
Won't do, see PR #583.

PR #583, opened 2023-01-18
Keep track of which services have emitted metrics

campesel:
We currently don't keep track of which services have emitted metrics for metrics defined outside services.

This PR adds a `valid` slice to time series to keep track of which services have emitted metrics.

### DIFF
--- a/runtime/metrics/metrics.go
+++ b/runtime/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"fmt"
+	"sync/atomic"
 )
 
 type Labels interface {
@@ -41,8 +42,8 @@ type Counter[V Value] struct {
 func (c *Counter[V]) Increment() {
 	if idx, ok := c.svcIdx(); ok {
 		c.inc(&c.ts.value[idx])
-		if idx != 0 {
-			c.ts.valid[idx] = true
+		if !c.ts.valid[idx].Load() {
+			c.ts.valid[idx].Store(true)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


runtime/metrics/metrics.go
```diff
@@ -41,6 +42,9 @@ type Counter[V Value] struct {
 func (c *Counter[V]) Increment() {
 	if idx, ok := c.svcIdx(); ok {
 		c.inc(&c.ts.value[idx])
+		if !c.ts.valid[idx].Load() {
+			c.ts.valid[idx].Store(true)
+		}
```
eandre, line 47: Better to always store than to load and store

runtime/metrics/metrics.go
```diff
@@ -210,11 +223,15 @@ func (m *metricInfo[V]) getTS(labels any) (ts *timeseries[V], setup bool) {
 
 	// Initialize the values if they haven't yet been set up.
 	if !setup {
-		n := m.reg.numSvcs
 		if m.svcNum > 0 {
-			n = 1
+			ts.value = make([]V, 1)
+			ts.valid = make([]atomic.Bool, 1)
+			ts.valid[0].Store(true)
```
eandre, line 229: Is this needed? Won't it happen in the other code above already?

campesel: It would, yeah. My thinking was that it'd be useful to know if a metric defined within a service is zero (as opposed to not emitting that metric until it's increased/set).

eandre (approved):

runtime/metrics/metrics.go
eandre: Ah, makes sense :) Thanks!

PR #592, opened 2023-01-25
runtime: implement trailing slash redirect

eandre:
This adds trailing-slash-redirect handling to
make using Encore APIs slightly more forgiving.
It particularly makes frontend support better
since frontends otherwise require a trailing slash
for the root node.

### DIFF
--- a/runtime/appruntime/api/server.go
+++ b/runtime/appruntime/api/server.go
@@ -297,6 +297,15 @@ func (s *Server) handler(w http.ResponseWriter, req *http.Request) {
 	if h == nil {
 		h, p, _ = r.Lookup(wildcardMethod, path)
 	}
+
+	if h == nil {
+		// If we still couldn't find a handler, check if there is one
+		// with a trailing slash redirect.
+		if handleTrailingSlashRedirect(r, w, req, path) {
+			return
+		}
+	}
+
 	if h != nil {
 		// Found an endpoint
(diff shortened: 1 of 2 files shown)


runtime/appruntime/api/server.go
```diff
@@ -364,3 +373,44 @@ func toUnnamedParams(ps httprouter.Params) UnnamedParams {
...
+		// Couldn't find a handler.
+		return false
+	}
+
+	// Moved Permanently, request with GET method
+	code := http.StatusMovedPermanently
+	if req.Method != http.MethodGet {
+		// Permanent Redirect, request with same method
+		code = http.StatusPermanentRedirect
+	}
```
DomBlack, line 412: Should this be temporary rather than permanent?

`Permanent` which only impact browser behaviour, rather than say an iOS or Android client - where as it could result in a bad time if the user changes the aPI path and the browser has a permanent redirect stored

eandre: This was based on the existing trailing slash redirect from httprouter, so I kept this as-is.

DomBlack: Yeah I'm still not sure it's the right behaviour for us though.

Will approve; but I'd prefer we changed it to a temporary redirect

DomBlack (approved):

PR #595, opened 2023-01-27
Add snippet tab to local dev dash

simon-johansson:
I have started adding snippets from https://encore.dev/docs/primitives/code-snippets
Added stuff under "APIs" and "Databases" so far.


cli/daemon/dash/dashapp/src/App.tsx
```diff
@@ -37,6 +38,9 @@ function App() {
     <ConnContext.Provider value={conn}>
       <Router>
         <Routes>
+          <Route path="/:appID/snippets" element={<SnippetPage />}>
```
DomBlack, line 41: Nit; but do snippets really sit under the app? They're not app specific right?

cli/daemon/dash/dashapp/src/encore-dark.css
```diff
@@ -0,0 +1,155 @@
+/*
```
DomBlack, line 1: dark mode? 😢

cli/daemon/dash/dashapp/src/components/snippets/snippetData.tsx
```diff
@@ -0,0 +1,683 @@
...
+      {content}
+    </a>
+  );
+};
+
+const docLink = (path: string) => {
+  return <span>{extLink(`https://encore.dev/docs${path}`, "Read the docs")} to learn more.</span>;
+};
+
+const apiSection: SnippetSection = {
```
DomBlack, line 36: This would be great if we actually built it / generated it from the docs within this repo so we don't now need to keep these and the markdown file in sync

PR #596, opened 2023-01-27
fix: use info for logging failed requests

melkstam:
As we have tried to utilise logging more in our project, we have noticed that all failed request automatically are logged as errors by encore.

However, we find this a bit annoying as many failed requests are not really errors. Especially the `4XX` range are user errors, and logging them clutters the logs. We think it would make sense to give more logging control to the developers.

Example from our logs:

image

You might have better ideas or other opinions or there are more things that need changing in the codebase. So please feel free to do what you wish with this PR!


DomBlack:
I think a better outcome here would be to report errors for 5xx style errors, and warnings for 4xx errors.

I think we'd need to think through a design which is obvious and clear to the end developer what the outcome is; while not overcomplicating the API design

PR #613, opened 2023-02-13
Track system metrics

campesel:
We want to monitor a number of system metrics and send them to the app metric provider.

This PR adds the code to track memory usage (in terms of allocated heap objects) and number of Go routines.


runtime/appruntime/metrics/aws/cloudwatch.go
```diff
@@ -156,3 +158,19 @@ func (x *Exporter) getClient() *cloudwatch.Client {
...
+	return []types.MetricDatum{
+		{
+			MetricName: aws.String("memory_usage_bytes"),
+			Timestamp:  aws.Time(now),
+			Value:      aws.Float64(float64(sysMetrics.MemoryUsageBytes)),
+		},
+		{
+			MetricName: aws.String("num_go_routines"),
+			Timestamp:  aws.Time(now),
+			Value:      aws.Float64(float64(sysMetrics.NumGoRoutines)),
```
DomBlack, line 173: I thought all encore inbuilt metrics started with `e_` so that users own metrics wouldn't overlap with ours?

runtime/appruntime/metrics/system/system.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func ReadSysMetrics() SysMetrics {
+	var memStats runtime.MemStats
+	runtime.ReadMemStats(&memStats)
+	return SysMetrics{
+		MemoryUsageBytes: memStats.Alloc,
+		NumGoRoutines:    runtime.NumGoroutine(),
+	}
+}
```
DomBlack, line 19: Given the repetition and chance of typo's I feel like this would be better if it returned something like:

```suggestion
func ReadSysMetrics() map[string]uint64 {
```

Where the key is the metric name `e_memory_usage` and the value is the current value.  (This works given all of them are gauges right now).

This means name's would be consistent, and if we add a new metric we only need to add it in one place

runtime/appruntime/metrics/aws/cloudwatch.go
```diff
@@ -156,3 +158,19 @@ func (x *Exporter) getClient() *cloudwatch.Client {
 	}
 	return x.client
 }
+
+func getSysMetrics(now time.Time) []types.MetricDatum {
```
DomBlack, line 162: System metrics need an instance ID / name field.

In GCP we can run up to two instances during a deploy - so you'd get two metrics being returned; 1 for the old instance, 1 for the new instance.

Without labels for the instances, your gauges are going to be bouncing between two values, depending on which instance reported last.


Also when some customers scale up beyond 1 instance on purpose, this isn't going to be great without being able to view _per instance_

runtime/appruntime/metrics/system/system.go
```diff
@@ -0,0 +1,19 @@
+package system
+
+import (
+	"runtime"
+)
+
+type SysMetrics struct {
+	MemoryUsageBytes uint64
+	NumGoRoutines    int
```
DomBlack, line 9: Why num goroutines and not CPU usage?

runtime/appruntime/metrics/aws/cloudwatch.go
campesel: Good catch!

runtime/appruntime/metrics/system/system.go
campesel: Sounds good. I'll also add a list of constants to have the key names in one place.

runtime/appruntime/metrics/aws/cloudwatch.go
campesel: That's already handled in GCP as we add `node_id` to the set of labels every metric has. It's not handled properly for AWS, tho.

I'll try and look for an instance ID we can attach to all metrics in AWS.

runtime/appruntime/metrics/system/system.go
campesel: I haven't found a way to get the current CPU usage.

runtime/appruntime/metrics/prometheus/prometheus.go
```diff
@@ -155,6 +157,32 @@ func (x *Exporter) getMetricData(now time.Time, collected []metrics.CollectedMet
...
+			}},
+			Samples: []*prompb.Sample{{
+				Value:     float64(sysMetrics.MemoryUsageBytes),
+				Timestamp: FromTime(now),
+			}},
+		},
+		{
+			Labels: []*prompb.Label{{
+				Name:  "__name__",
+				Value: "num_go_routines",
```
eandre, line 176: Maybe to stick with the "<what>_<unit>" scheme this should be "goroutine_count" (also goroutine is one word, not two). But would be good to review all metrics we intend to add and make sure we have a consistent naming scheme for all of them.

eandre (changes requested):

runtime/appruntime/metrics/prometheus/prometheus.go
campesel: Since the other metric is called `e_requests_total`, then we should call this one `e_goroutines_total`, then?

campesel: Now that I think about it, I avoided using the `total` suffix for the number of goroutines because that's a gauge. I think `count` would also suggest that being a counter?

eandre: True. What does prometheus call it?

runtime/appruntime/metadata/metadata.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"time"
+
+	gcemetadata "cloud.google.com/go/compute/metadata"
```
eandre, line 12: This is dependent on the GCP libraries. We've introduced specific build tags for each cloud provider to ensure we don't include dependencies on cloud providers that aren't used by the application.

I think this needs to be broken out into separate packages with their own build tags, or alternatively they could be the same package but each cloud provider has its own file, and we make sure the functions only get referenced from packages that are imported only for that cloud provider.

runtime/appruntime/metrics/prometheus/prometheus.go
campesel: The Prometheus Go client calls it `go_goroutines`.

eandre: Ok. Do we have a list of the runtime metrics they track? I think we should look at the whole list of runtime metrics and come up with a consistent naming strategy instead of naming things one-by-one.

runtime/appruntime/metrics/gcp/cloud_monitoring.go
```diff
@@ -126,10 +131,14 @@ func (x *Exporter) getMetricData(newCounterStart, endTime time.Time, collected [
 		metricType = "custom.googleapis.com/" + cloudMetricName
 
 		doAdd := func(val *monitoringpb.TypedValue, svcIdx uint16) {
-			labels := make(map[string]string, len(baseLabels)+1)
+			containerMetadataLabels := containerMetadataLabels(x.containerMetadata)
```
eandre, line 134: Same here

runtime/appruntime/metrics/prometheus/prometheus.go
```diff
@@ -63,16 +68,19 @@ func (x *Exporter) getMetricData(now time.Time, collected []metrics.CollectedMet
 	data := make([]*prompb.TimeSeries, 0, len(collected))
 
 	doAdd := func(val float64, metricName string, baseLabels []*prompb.Label, svcIdx uint16) {
-		labels := make([]*prompb.Label, len(baseLabels)+2)
+		containerMetadataLabels := containerMetadataLabels(x.containerMetadata)
```
eandre, line 71: Same here

eandre (approved):

eandre (approved):

eandre (approved):

PR #619, opened 2023-02-22
[ENC-1288] Fix indeterministic build/encore run

DomBlack:
This commit fixes a bug introduced in #505 when we made the parser run in parallel.

The issue ultimately was a race condition within the main goroutine, such that when the last worker finished parsing a package, the main goroutine could either read from the `pkgCh` or the `workerDone` channel.

If it picked the `workerDone` channel, then the last package would never be added to the list of known packages, which could result in random errors later in the encore parse/compile pipeline.

I've also updated the unit tests to ensure the test cache is cleared for each PR


.github/workflows/go.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       run: cd encr.dev && go build ./...
 
     - name: Test
-      run: cd encr.dev && go test -short ./...
+      run: cd encr.dev && go clean -testcache && go test ./...
```
eandre, line 53: This should not be necessary; go's test cache should be precise. I think the issue is maybe how we're using `testscript`. Can we revert this and look into that issue separately?

eandre: The `parser` package has been modified a bunch of times since then, which would cause the test to be re-run a bunch of times. That can't be the issue.

DomBlack: Only one test failed time in about 20-30 runs of all the tests with the default parallel settings we had, if I dropped `numberOfWorkers` down to 1, then about 10 would fail each run. (i.e. the less workers the more likely the bug was to occur).

It was simply a probability game, but allowing our tests to be cached we reduce the chances of us detecting an issue like this in the future - by forcing CI to always run on all tests, we increase the probability of catching a bug like this.

eandre (approved):

PR #621, opened 2023-02-22
Simple first pass on JSON-based metrics exporter

mason-stewart:
This PR was a result of asking the Encore team if there was a way to view metrics locally during development. The answer was "not yet", so I asked if they could provide some direction so that I could help get this implemented. @campesel gave some guidance, so here's a pull request as promised.

This also adds a development-only route at /__encore/metrics that renders the result of this exporter. This page will now render JSON response that look something like this:

```json
{
  "metrics": [
    {
      "name": "orders_processed",
      "type": "Counter",
      "labels": [],
      "value": 12
    },
    {
      "name": "e_requests_total",
      "type": "Counter",
      "labels": [
        {
          "endpoint": "Index"
        },
        {
          "code": "ok"
        }
      ],
      "value": 2
    }
  ]
}
```

Initially I wrote this just a few functions in the `runtime/appruntime/api/encore_routes.go` file, but it was clear that I was mimicking a lot of the functionality in the existing metrics exporters, and there was no simple, obvious (to me at least) place to test this if it was defined in `runtime/appruntime/api/encore_routes.go`, so I rewrote this to be a new exporter called `JSONBased`, following the existing patterns. 

I borrowed the tests from `runtime/appruntime/metrics/prometheus/prometheus_test.go` and adjusted them based on the shape of the JSON that @campesel requested. This was helpful since there were lots of little details I would have missed. I ended up borrowing heavily from the `getMetricData` in the `runtime/appruntime/metrics/prometheus/prometheus.go`.

This is enabled in development mode only (so far as I can understand) via a new field on `config.Config.Metrics`, but I believe it would be possible to enable this is non-local environments if desired.

I'm still a Golang rookie and I'm sure I've made some rookie mistakes here, so please don't hold back on feedback. Additionally, if the exporter-based approach is wrong, I'm happy to scrap this and take another approach. Cheers! 🤠

### DIFF
--- a/runtime/appruntime/metrics/json_based/json_based.go
+++ b/runtime/appruntime/metrics/json_based/json_based.go
@@ -1,0 +1,141 @@
+package json_based
+
+import (
+	"context"
+	"time"
+
+	"github.com/rs/zerolog"
+
+	"encore.dev/appruntime/config"
+	"encore.dev/metrics"
+)
+
+func New(svcs []string, cfg *config.JSONBasedMetricsProvider, rootLogger zerolog.Logger) *Exporter {
+	return &Exporter{
+		svcs:       svcs,
+		cfg:        cfg,
+		rootLogger: rootLogger,
+	}
+}
+
+type Exporter struct {
+	svcs       []string
+	cfg        *config.JSONBasedMetricsProvider
+	rootLogger zerolog.Logger
+}
+
+// Can be easily marshaled into a reasonable JSON object
+type JSONMetrics struct {
+	Metrics []JSONMetric `json:"metrics"`
+}
+
+type Label map[string]string
+
+type JSONMetric struct {
+	Name   string  `json:"name"`
+	Type   string  `json:"type"`
+	Labels []Label `json:"labels"`
+	Value  any     `json:"value"`
+}
+
+func (x *Exporter) Shutdown(_ context.Context) {}
+
+func (x *Exporter) Export(ctx context.Context, collected []metrics.CollectedMetric) error {
+	// This is a no-op since there's currently no need to export JSON metrics;
+	// they're only emitted via an internal api endpoint in runtime/appruntime/api/encore_routes.go
+	return nil
+}
+
+// Use to export a slice of CollectedMetric into a shape that's useful when rendering
+// metrics into JSON. Marshalling the return value will result in nicely formatted JSON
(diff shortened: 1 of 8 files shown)


runtime/appruntime/metrics/json_based/json_based.go
```diff
@@ -0,0 +1,141 @@
...
+	"context"
+	"time"
+
+	"github.com/rs/zerolog"
+
+	"encore.dev/appruntime/config"
+	"encore.dev/metrics"
+)
+
+func New(svcs []string, cfg *config.JSONBasedMetricsProvider, rootLogger zerolog.Logger) *Exporter {
```
eandre, line 13: I think the JSON metrics should not be thought of as a metrics provider/exporter, but as simply a pure function that takes `[]metrics.CollectedMetric` and returns a data a data structure. Essentially what you have with `(*Exporter).GetMetricData`. But we don't need to treat it as a "JSONBasedMetricsProvider" or change the configuration, I don't think. That way exposing metrics via API is orthogonal to the notion of a metrics exporter. (Thanks @campesel for pointing this out!)

runtime/appruntime/metrics/json_metrics_exporter.go
```diff
@@ -0,0 +1,18 @@
+package metrics
+
+import (
+	"encore.dev/appruntime/config"
+	"encore.dev/appruntime/metrics/json_based"
+)
+
+func init() {
+	registerProvider(providerDesc{
```
eandre, line 9: See earlier comment: I think we shouldn't need the JSON metrics to be considered a metrics exporter

runtime/metrics/registry_internal.go
```diff
@@ -68,6 +68,21 @@ func (r *Registry) Collect() []CollectedMetric {
...
 
+// Converts MetricType enum values into their common English names
+func (mt MetricType) Name() string {
+	switch mt {
+	case CounterType:
+		return "Counter"
+	case GaugeType:
+		return "Gauge"
+	case HistogramType:
+		return "Historgram"
```
eandre, line 79:
```suggestion
		return "Histogram"
```

eandre (commented):
Thanks Mason for doing this! Overall it looks good, but a few points that would be nice to address before merging.

mason-stewart:
Ah geez, I completely forgot about this PR for almost a year. Whoops. I'm going to close this since it has quite a few conflicts and we no longer need this. Thanks!

PR #625, opened 2023-03-03
Implement Datadog exporter

campesel:
This PR adds a new metric exporter to export metrics to Datadog. This is dependent on the platform setting the new `datadog` config field appropriately.


runtime/appruntime/metrics/datadog_exporter.go
```diff
@@ -0,0 +1,50 @@
...
+				return nil
+			}
+
+			cfg := m.cfg.Runtime.Metrics.Datadog
+			if cfg.Site == "" {
+				m.rootLogger.Err(err).Msg("unable to initialize metrics exporter: Datadog site unset")
+				return nil
+			}
+
+			err = os.Setenv("DD_SITE", cfg.Site)
```
eandre, line 30: Is this necessary? Is there no other way of getting the site information and API key to the client? It makes it easy for rogue libraries to sniff credentials. If the libraries are designed that way I guess there's not too much we can do about it, but would be nice to avoid.

runtime/appruntime/metrics/datadog/datadog.go
```diff
@@ -0,0 +1,188 @@
...
+func (x *Exporter) Export(ctx context.Context, collected []metrics.CollectedMetric) error {
+	now := time.Now()
+	data := x.getMetricData(now, collected)
+	data = append(data, x.getSysMetrics(now)...)
+	body := datadogV2.MetricPayload{Series: data}
+
+	ctx = x.newContext(ctx)
+	configuration := datadog.NewConfiguration()
+	apiClient := datadog.NewAPIClient(configuration)
+	api := datadogV2.NewMetricsApi(apiClient)
```
eandre, line 50: Do we need to instantiate the client at export time? Intuitively I would expect that to happen during Exporter initialization.

campesel: We don't, I'll change that 👍

eandre (approved):

PR #629, opened 2023-03-11
Allow access to underlying *pgx.Conn

brentd:
Hi 👋 

This PR attempts to expand upon `sqldb.Driver`, which provides access to `*pgxpool.Pool`, by adding a way to access the underlying `*pgx.Conn` of a connection as well. 

Either implementation offered here would work, and there might be another method I wasn't able to think of. We only need one. I am happy to add tests with some guidance on where they should be.

## Use case

The addition of `sqldb.Driver` helps when we need lower level access to DB features that pgx provides, such as `COPY FROM STDIN`. However, since the rest of our code expects stdlib's `*sql.DB` interfaces, it becomes all-or-nothing: once we've dropped down to pgx for a transaction, we can't go back to a `*sql.DB` compatible interface for the duration of a transaction.

The addition of access to `*pgx.Conn` would allow us to weave the use of pgx in more surgically so we can keep the rest of our non-pgx code focused on `*sql.DB` which, being from the stdlib, should be a more stable interface. We also use sqlc which by default is compatible with the stdlib interface -- although it has a `pgx/v5` adapter, it is difficult for us to justify using it in some of our services but not others.

## Usage

Usage of the `DriverConn` implementation would look something like this

```golang
// Checkout a *sql.Conn
conn, _ := testdb.Stdlib().Conn(ctx)

// Start a transaction
tx, _ := conn.BeginTx(ctx, nil)
defer tx.Rollback()

sqldb.DriverConn(conn, func(driverConn *pgx.Conn) error {
	// do low level stuff with *pgx.Conn
})

// Do something with a package that expects a *sql.DB-like interface
somepackage.RunQuery(ctx, tx)

_ := tx.Commit()
```

### DIFF
--- a/runtime/storage/sqldb/db.go
+++ b/runtime/storage/sqldb/db.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 
 	"encore.dev/appruntime/config"
@@ -280,4 +281,33 @@ func Driver[T SupportedDrivers](db *Database) T {
 // Currently only [*pgxpool.Pool] is supported.
 type SupportedDrivers interface {
 	*pgxpool.Pool
+}
+
+// DriverConn provides access to the underlying driver connection given a stdlib
+// *sql.Conn connection. The driverConn must not be used outside of f, and conn
+// must be a *sql.Conn originating from sqldb.Database or this function panics.
+//
+//	conn, _ := db.Stdlib().Conn(ctx) // Checkout a connection from the pool
+//	sqldb.DriverConn(conn, func(driverConn *pgx.Conn) error) error {
+//	  // do stuff with *pgx.Conn
+//	}
+//
+// This is defined as a generic function to allow compile-time type checking
+// that the Encore application is expecting a driver that is supported.
+//
+// At some point in the future where Encore adds support for a different
+// database driver this will be made with backwards compatibility in mind,
+// providing ample notice and time to migrate in an opt-in fashion.
+func DriverConn[T SupportedDriverConns](conn *sql.Conn, f func(driverConn T) error) error {
+	return conn.Raw(func(c any) error {
+		parentConn := (c).(wrappedConn).parent
+		rawConn := (parentConn).(*stdlibdriver.Conn).Conn()
+		return f(rawConn)
+	})
+}
+
+// SupportedDriverConns is a type list of all supported database drivers
+// connections. Currently only [*pgx.Conn] is supported.
+type SupportedDriverConns interface {
+	*pgx.Conn
 }


runtime/storage/sqldb/stdlib_wrapper_internal.go
```diff
@@ -321,6 +323,17 @@ var (
...
+// Conn exposes the underlying pgx database connection so that it can be
+// accessed via (*sql.Conn).Raw.
+//
+//	conn.Raw(func(driverConn any) error {
+//		pgxConn := driverConn.(interface{ Conn() *pgx.Conn }).Conn()
+//		// do stuff with *pgx.Conn
+//	})
+func (c wrappedConn) Conn() *pgx.Conn {
+	return (c.parent).(*stdlibdriver.Conn).Conn()
+}
```
brentd, line 335: This is the second way which is a bit more direct, piggy-backing on `(*sql.Conn).Raw`. I took some inspiration from the implementation of [`errors.Unwrap`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/errors/wrap.go;l=16-24) in the example code which type asserts that `driverConn` is actually wrapping `*pgx.Conn` and exposes it via `Conn()`.

I suppose the benefit of this approach is that client code could type assert `driverConn` in a switch statement without having to import any packages from Encore, so hypothetically a client package could support a `*sql.Conn` that _doesn't_ have another layer of wrapping over `*pgx.Conn` like Encore's.

I suspect though that the compile-time type checking of the `DriverConn` approach is superior in practice.

runtime/storage/sqldb/db.go
```diff
@@ -281,3 +282,32 @@ func Driver[T SupportedDrivers](db *Database) T {
...
+		rawConn := (parentConn).(*stdlibdriver.Conn).Conn()
+		return f(rawConn)
+	})
+}
+
+// SupportedDriverConns is a type list of all supported database drivers
+// connections. Currently only [*pgx.Conn] is supported.
+type SupportedDriverConns interface {
+	*pgx.Conn
+}
```
brentd, line 313: This is the first approach which follows the shape of [`(*sql.Conn).Raw`](https://pkg.go.dev/database/sql#Conn.Raw) to expose the underlying `*pgx.Conn` in a short-lived fashion, with the benefit of compile-time type checking in case something changes in Encore.

eandre:
Hey @brentd, I think this is a compelling use case and we should support this. I'm leaning towards mirroring the `SupportedDriverConns` approach. (The reason we went with that approach for getting the database driver is to statically be able to determine which database drivers are in use by a program, which makes it much easier to gracefully migrate over to another default driver in the future while preserving backwards compatibility).

brentd:
@eandre thanks for the feedback. I've reduced this PR to only the `DriverConn` approach using the `SupportedDriverConns` type list.

If there's a good spot for me to add a test for this, let me know. Otherwise, I did test this out on our app when I first drafted this PR and this approach works and will help us use pgx more surgically than we can now.

eandre (approved):

eandre:
Thanks for this, and sorry for dropping the ball on getting it merged. Looks great.

PR #634, opened 2023-03-21
Allow serving a root path

azr:
Hello there !

I'd like to have an endpoint serve things from the service's root path.

I'm totally hacking around things here, but I'd like to serve some HTML from that index, 'cause YOLO.

Basically I want to have everything runing in that encore server.


eandre:
Hey @azr, thanks for the pull request! It would be useful to understand what it is you're trying to do exactly. Typically if you want to serve HTML you would use a catch-all route: `path=/*path`. Does that not work?

PR #645, opened 2023-03-31
v2/app: fix metadata for auth handlers, subscriptions

eandre:


v2/app/service_discovery.go
```diff
@@ -49,6 +49,8 @@ func discoverServices(pc *parsectx.Context, result *parser.Result) []*Service {
 			sd.possibleServiceRoot(r.Package(), false)
 		case resource.PubSubSubscription:
 			sd.possibleServiceRoot(r.Package(), false)
+		case resource.AuthHandler:
```
DomBlack, line 52: Is an auth handler the cause of a service to be created? I didn't think an auth handler was part of a service, but more library code

v2/internals/schema/schema_parser.go
```diff
@@ -144,7 +144,8 @@ func (r *typeResolver) parseType(file *pkginfo.File, expr ast.Expr) Type {
 			for _, field := range expr.Fields.List {
 				typ := r.parseType(file, field.Type)
 				if len(field.Names) == 0 {
-					r.errs.AddPos(field.Pos(), "cannot use anonymous fields in Encore struct types")
+					//r.errs.AddPos(field.Pos(), "cannot use anonymous fields in Encore struct types")
```
DomBlack, line 147: We're allowing this now?

v2/app/api_framework.go
```diff
@@ -110,5 +112,14 @@ func configureAPIFramework(pc *parsectx.Context, services []*Service, res *parse
...
 
+	// Call modifySvcDesc for subscriptions and auth handlers, even though they don't actually
+	// modify the framework description, to actually tag the package as a service.
+	for _, ss := range subscriptions {
+		modifySvcDesc(ss.Package(), nil, func(svc *Service, desc *apiframework.ServiceDesc) {})
+	}
+	for _, ah := range authHandlers {
+		modifySvcDesc(ah.Package(), nil, func(svc *Service, desc *apiframework.ServiceDesc) {})
+	}
+
```
DomBlack, line 123: For subscriptions I didn't do this on purpose, as a subscription does not create an API service, so there's no `apiframework.ServiceDesc` - but it still can create a service.

What was the need to do this here?

PR #648, opened 2023-04-03
Update zig

DomBlack:
We had an issue with our cross compile failing, because Zig was too old to understand a flag that go uses.

See ziglang/zig#11863

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -51,7 +51,7 @@ jobs:
     # for failures. We will also fail fast so that we don't waste time running tests
     # that are already failing.
     - name: Run all tests multiple times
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     # for failures. We will also fail fast so that we don't waste time running tests
     # that are already failing.
     - name: Run all tests multiple times
-      run:  cd encr.dev && go test -v --count=25 -failfast -shuffle=on -tags=dev_build ./...
+      run:  cd encr.dev && go test -v --count=5 -failfast -shuffle=on -tags=dev_build ./...
```
DomBlack, line 54: We where getting a 10 minute timeout on our nightly tests, so I dropped this down only 5 reptitions for now, given we run it every night, we should still pickup flaky tests

PR #658, opened 2023-04-05
v2/parser/infra/sqldb: ignore non-service migrations

eandre:


DomBlack (approved):
I approve this for now, but I still think a better option is to have the resource unregistered during the service discovery phase

PR #668, opened 2023-04-14
Add note about redaction not applying in local dev

jdbann:
When running an encore app locally, the `encore:"sensitive"` tags are ignored. This is intentional, but was not explained in the documentation.


eandre (approved):
Thank you!

PR #677, opened 2023-04-24
Fix issues with the VFS under windows

DomBlack:
This commit fixes an issue with our VFS implementation, where when walking a directory and its subdirectories to add them into a new VFS we had been using `os.PathSeperator` to split the directory names out, but it turns out that `fs.WalkDir` only ever uses `/` as the path separator. This lead to issues on windows where we would append directories into the datastructure incorrectly, leading them to be listed as directory listings, but not openable during walks.

I made it to any errors during computeConfig will point back to the original source code Node which loads the config

While here I added some detection code to disable/enable colours in errors by default depending on if the terminal supports ansi colours.

This commit also allows errors being wrapped by `srcerrors` to have the bottom level stack trace used. (This function is copied over from the `eerror` pkg)

Thanks to @simonvc for the bug report on slack: https://encoredev.slack.com/archives/C0356ERKBEH/p1682296355179289


pkg/cueutil/build.go
```diff
@@ -131,28 +131,28 @@ func writeFSToPath(filesys fs.FS, targetPath string) error {
 	// Copy the files into the temporary directory
 	err := fs.WalkDir(filesys, ".", func(path string, info fs.DirEntry, err error) error {
 		if err != nil {
-			return err
+			return eerror.Wrap(err, "config", "unable to walk VFS", nil)
```
DomBlack, line 134: These wraps are to expose the underlying error stack

pkg/vfs/vfs.go
```diff
@@ -197,7 +196,7 @@ func (v *VFS) AddFile(path string, bytes []byte, time time.Time) (*fileContents,
 
 // AddDir records a directory into the VFS.
 func (v *VFS) AddDir(dirPath string) *directoryContents {
-	dirParts := strings.Split(dirPath, string(os.PathSeparator))
+	dirParts := strings.Split(dirPath, "/")
```
DomBlack, line 199: This is the actual bug fix

pkg/errinsrc/srcrender.go
```diff
@@ -20,8 +21,12 @@ const endEscape = "\x1b[0m"
 
 const tabSize = 4
 
-var aurora = auroraPkg.NewAurora(true)
-var enableColors = true
+var aurora auroraPkg.Aurora
+var enableColors bool
+
+func init() {
+	ColoursInErrors(supportscolor.Stdout().SupportsColor)
```
eandre, line 28: This code runs in many different environments including the build system on the web platform, and we don't always pipe things to stdout. This seems like it could easily do the wrong thing as a result.

Probably we'd be better off with a setting to turn it off explicitly, and filtering it out at the time of rendering if we know we're outputting to a terminal that doesn't support colors?

pkg/errinsrc/stack.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
+
+	"github.com/pkg/errors"
```
eandre, line 8: Is this needed? Yet another error library dependency

pkg/errinsrc/srcrender.go
DomBlack: True, but this is a step to better behaviour on older windows terminals, other locations explicitly turn the colours on or off as needed (such as the build system)

This init change just moves us from always on to mostly on

pkg/errinsrc/stack.go
DomBlack: This isn't a new library in the code base (we already used it, as do some of our dependencies), what it is is a semi-standard interface / struct type for different error libraries to use when they want to expose stack trace information.

pkg/errinsrc/srcrender.go
eandre: But that could cause a regression on the platform side, for example

DomBlack: What regression are you worried about (i.e. I'm not following)?

The platform doesn't use this package directly, apart from the builder images, which are unix based and their Stdout does support colour. Plus the builder images already have this init function:

```go
func init() {
	errinsrc.ColoursInErrors(true)
	errinsrc.TerminalWidth = 80
}
```

PR #679, opened 2023-04-25
Add support for Encore Cloud's new PubSub system

DomBlack:
This commit adds support for Encore Cloud's new pubsub system, which will unlock the ability to do Exactly Once delivery on PubSub subscriptions.


runtime/pubsub/manager_internal.go
```diff
@@ -136,7 +136,7 @@ func registerProvider(p func(mgr *Manager) provider) {
 
 var _ types.PushEndpointRegistry = (*Manager)(nil)
 
-func (mgr *Manager) RegisterPushSubscriptionHandler(id types.SubscriptionID, handler types.PushEndpointHandler) {
+func (mgr *Manager) RegisterPushSubscriptionHandler(id types.SubscriptionID, handler http.HandlerFunc) {
```
DomBlack, line 139: I needed access to the raw `ResponseWriter` to do the SSE, which meant dropping the nice `func(req) error` we had before :(

cli/daemon/run/infra/infra.go
```diff
@@ -260,12 +268,25 @@ func (rm *ResourceManager) UpdateConfig(cfg *config.Runtime, md *meta.Data, dbPr
...
+				EncoreCloud: &config.EncoreCloudPubsubProvider{},
+			})
+			providerID = 1
+		}
+
 		cfg.PubsubTopics = make(map[string]*config.PubsubTopic)
 		for _, t := range md.PubsubTopics {
+			providerName := t.Name
+			if useLocalEncoreCloudAPI {
+				providerName = "res_0o9ioqnrirflhhm3t720"
```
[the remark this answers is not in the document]
DomBlack: These ID's are in the other PR, the reason for specific ID's is the datatype is an EID within the API server, and if you read the README in that PR, you'll see I've listed setup steps which use these same hard coded ID's to testing locally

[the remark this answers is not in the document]
eandre: Makes sense. Can we make them constants and document them?

cli/daemon/run/infra/infra.go
```diff
@@ -223,7 +226,12 @@ func isDockerRunning(ctx context.Context) bool {
 // UpdateConfig updates the given config with infrastructure information.
 // Note that all the requisite services must have started up already,
 // which in practice means that (*optracker.AsyncBuildJobs).Wait must have returned first.
-func (rm *ResourceManager) UpdateConfig(cfg *config.Runtime, md *meta.Data, dbProxyPort int) {
+func (rm *ResourceManager) UpdateConfig(cfg *config.Runtime, md *meta.Data, dbProxyPort int) error {
+	useLocalEncoreCloudAPI, err := rm.setTestEncoreCloud(cfg)
```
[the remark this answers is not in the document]
DomBlack: Originally I had it panic, but I changed it because the deamon would be the thing which panics, which would not give the error visibly to the CLI tool.

Given that the only call site of `UpdateConfig` already had error handling which passed the errors to the CLI, this felt less error prone in terms of us using it and not using it 100% correctly and seeing a random panic which might take time to debug - vs just immediately seeing a useful error

cli/daemon/run/infra/encorecloudtesting.go
```diff
@@ -0,0 +1,54 @@
...
+	"encore.dev/appruntime/exported/config"
+)
+
+// setTestEncoreCloud sets the Encore Cloud API configuration to use a local
+// Encore Cloud API server.
+//
+// It returns true if one has been configured, or false if not.
+//
+// To use it the `encore run` command must be started with the following environment variables:
+// - ENCORE_CLOUD_LOCAL_SERVER: the URL of the local Encore Cloud API server
```
eandre, line 18: Maybe make these `ENCORECLOUD_...`? We have special code in the runtime that picks up all environment variables beginning with `ENCORE_` and these seem unrelated to the runtime itself

runtime/pubsub/internal/encorecloud/push_handler.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+// pushPayload is the payload that Encore Cloud will generate
+// when pushing a subscription attempt to a push endpoint
+type pushPayload struct {
+	Data            []byte            `json:"data"`
+	Attributes      map[string]string `json:"attributes"`
+	MessageID       string            `json:"messageId"`
+	PublishTime     time.Time         `json:"publishTime"`
+	DeliveryAttempt int               `json:"deliveryAttempt"`
```
eandre, line 25: We generally use `kebab_case` for JSON keys

runtime/pubsub/internal/encorecloud/push_handler.go
```diff
@@ -0,0 +1,187 @@
...
+
+			// Start the event stream
+			w.Header().Set("Content-Type", "text/event-stream")
+			w.Header().Set("Cache-Control", "no-cache")
+			w.Header().Set("Connection", "keep-alive")
+			w.WriteHeader(http.StatusOK)
+			flusher.Flush()
+
+			// Run the subscription function in a goroutine
+			response := make(chan error)
```
eandre, line 80: It's easy for cases like these to result in a leaking goroutine. Better to do:
```suggestion
			response := make(chan error, 1)
```

DomBlack: I don't understand your point here, this change could result in the defer deadlocking if the context got cancelled and then the callback paniced

eandre: What do you mean?

eandre: By leaving the channel as unbuffered, if the parent goroutine doesn't read the value then the goroutine will block forever trying to send on `response`, which leaks a goroutine

v2/internals/schema/schema_parser.go
```diff
@@ -194,7 +194,7 @@ func (r *typeResolver) parseType(file *pkginfo.File, expr ast.Expr) Type {
 				// We have a slice, not an array.
 
 				// Translate list of bytes to the builtin Bytes kind.
-				if elem.Family() == Builtin && elem.(BuiltinType).Kind == Uint8 {
+				if elem != nil && elem.Family() == Builtin && elem.(BuiltinType).Kind == Uint8 {
```
eandre, line 197: Add a bailout above instead

eandre (approved):

PR #685, opened 2023-04-27
Fix possible build failures

eandre:
It seems in some cases the error list is empty even though there is
an error. It's hard to debug as it's intermittent. Try to handle
this a bit more gracefully so we can track down the root cause.

### DIFF
--- a/v2/compiler/build/build.go
+++ b/v2/compiler/build/build.go
@@ -242,7 +242,7 @@ func (b *builder) buildMain() {
 
 		if b.cfg.Ctx.Build.Debug {
 			// Disable inlining for better debugging.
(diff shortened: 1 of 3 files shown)


v2/compiler/build/build.go
```diff
@@ -242,7 +242,7 @@ func (b *builder) buildMain() {
 
 		if b.cfg.Ctx.Build.Debug {
 			// Disable inlining for better debugging.
-			args = append(args, `-gcflags "all=-N -l"`)
+			args = append(args, "-gcflags", "all=-N -l")
```
eandre, line 245: This is mostly to bring it in line with the other flags above.

PR #686, opened 2023-04-27
ENC-1448 Add Runtime Configurations for gateways and remote services

ekerfelt:
As we add support for deploying an encore app across multiple containers, we need to inject configuration for service discovery and routing.

This PR introduces two additional Runtime config fields which specifies:
* Which Gateways an instance should be running
* Where remote services can be reached

The Gateway/API group feature is not yet implemented, but the idea is to provide common public API gateways for a collection of encore services

### DIFF
--- a/runtime/appruntime/exported/config/config.go
+++ b/runtime/appruntime/exported/config/config.go
@@ -37,15 +37,15 @@ type Runtime struct {
 	AuthKeys      []EncoreAuthKey `json:"auth_keys,omitempty"`
 	CORS          *CORS           `json:"cors,omitempty"`
 
-	SQLDatabases    []*SQLDatabase          `json:"sql_databases,omitempty"`
-	SQLServers      []*SQLServer            `json:"sql_servers,omitempty"`
-	PubsubProviders []*PubsubProvider       `json:"pubsub_providers,omitempty"`
-	PubsubTopics    map[string]*PubsubTopic `json:"pubsub_topics,omitempty"`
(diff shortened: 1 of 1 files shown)


runtime/appruntime/exported/config/config.go
```diff
@@ -44,13 +44,32 @@ type Runtime struct {
 	RedisServers    []*RedisServer          `json:"redis_servers,omitempty"`
 	RedisDatabases  []*RedisDatabase        `json:"redis_databases,omitempty"`
 	Metrics         *Metrics                `json:"metrics,omitempty"`
+	Services        []*Service              `json:"services,omitempty"`
```
eandre, line 47: Would it make sense to call this "service discovery" instead? Should it be a `map[string]Service` instead? (Also probably not worth having pointers)

runtime/appruntime/exported/config/config.go
```diff
@@ -44,13 +44,32 @@ type Runtime struct {
...
+	Gateways        []*Gateway              `json:"gateways,omitempty"`
```
eandre, line 48: It's not clear exactly what it means to have an entry here. Is it saying all gateways listed should be served by the running container?

runtime/appruntime/exported/config/config.go
ekerfelt: 👍

runtime/appruntime/exported/config/config.go
ekerfelt: Added a comment

eandre (approved):

PR #687, opened 2023-04-27
Improve error handling of panics

eandre:
We were only handling bailouts, not unintended panics.

### DIFF
--- a/v2/internals/perr/perr.go
+++ b/v2/internals/perr/perr.go
@@ -295,6 +295,27 @@ func CatchBailout(recovered any) (l *List, ok bool) {
 	return nil, false
 }
 
+// CatchBailoutAndPanic is like CatchBailout but also catches other panics.
+// In both types of panics it converts the result to an error.
+//
+// If there is no panic it returns the error provided in the first argument
+// back to the caller. That way, it can usually be used like so
+// (inside a deferred function):
+//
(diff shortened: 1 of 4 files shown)


v2/internals/perr/perr.go
```diff
@@ -295,6 +296,28 @@ func CatchBailout(recovered any) (l *List, ok bool) {
...
+//	err, _ = perr.CatchBailoutAndPanic(err, recover())
+func CatchBailoutAndPanic(currentErr error, recovered any) (err error, caught bool) {
+	if recovered == nil {
+		return currentErr, false
+	}
+
+	if b, ok := recovered.(bailout); ok {
+		err = b.l.AsError()
+	} else {
+		err = fmt.Errorf("internal compiler error: unhandled panic: %v\n%s",
```
DomBlack, line 315: In the old v1 parser we had this function (which you could call; https://github.com/encoredev/encore/blob/main/pkg/errinsrc/srcerrors/errors.go#L20

It handles detecting the bailout, and returning the list _or_ if it's an unknown time, creating a new errinsrc object with the stack trace otherwise... this means that the error is rendered the same way in the CLI

I'd prefer if we continue to use that approach

DomBlack (changes requested):

DomBlack (approved):

PR #703, opened 2023-05-08
configgen: fix handling of imported types

eandre:
Also copy over a bunch of the cuegen tests since they weren't testing the compilation step.

Thanks Richie Jenkins for reporting.


v2/codegen/infragen/configgen/configgen_test.go
```diff
@@ -16,3 +16,11 @@ func TestCodegen(t *testing.T) {
 
 	codegentest.Run(t, fn)
 }
+
+func TestCodegen_Cue(t *testing.T) {
```
eandre, line 20: Note to self: remove

DomBlack: Remove ;)

v2/codegen/infragen/configgen/configgen.go
```diff
@@ -384,7 +394,24 @@ func (cb *configUnmarshalersBuilder) typeUnmarshalerFunc(typ schema.Type) (f *St
...
+	// Otherwise allocate a new name.
+	suffix := ""
+	if n := cb.seenNames[decl.Name]; n > 0 {
+		suffix = fmt.Sprintf("%d", n+1)
+	}
+	name := fmt.Sprintf("encoreInternalTypeConfigUnmarshaler_%s%s", decl.Name, suffix)
+	cb.seenNames[decl.Name]++
+	cb.allocatedNames[key] = name
+
+	return name, rtnTyp
```
DomBlack, line 414: Nit; While this works, I wonder if it'd be better to also key off the package name too? - although we'd still need the numerical version fallback incase we saw `pkg_bar.Foo` and `pkg.bar_foo` and `pkg_bar_foo` at the same time.

DomBlack (changes requested):
Just remove the method you've already flagged.

Nice to see we brought the tests over to v2 from the older v1 path

PR #705, opened 2023-05-09
Implement initial Ordered PubSub Support

DomBlack:
This commit adds initial support for ordered pubsub on AWS, GCP and EncoreCloud.

It uses a new `OrderingAttribute` configuration from the topic to extract the ordering key from the message attributes at the time of the publish.

### DIFF
--- a/runtime/pubsub/internal/types/public.go
+++ b/runtime/pubsub/internal/types/public.go
@@ -90,10 +90,49 @@ type TopicConfig struct {
 	// This field is required.
 	DeliveryGuarantee DeliveryGuarantee
 
-	// OrderingKey is the name of the message attribute used to group
-	// messages and delivery messages with the same OrderingKey value
-	// in the order they were published.
+	// OrderingAttribute is the message attribute to use as a ordering key for
+	// messages and delivery will ensure that messages with the same value will
+	// be delivered in the order they where published.
 	//
-	// If OrderingKey is not set, messages can be delivered in any order.
-	// OrderingKey string - OrderingKey is currently not supported.
+	// If OrderingAttribute is not set, messages can be delivered in any order.
+	//
+	// It is important to note, that in the case of an error being returned by a
+	// subscription handler, the message will be retried before any subsequent
+	// messages for that ordering key are delivered. This means depending on the
+	// retry configuration, a large backlog of messages for a given ordering key
+	// may build up. When using OrderingAttribute, it is recommended to use reason
+	// about your failure modes and set the retry configuration appropriately.
+	//
+	// Once the maximum number of retries has been reached, the message will be
+	// forwarded to the dead letter queue, and the next message for that ordering
+	// key will be delivered.
+	//
+	// To create attributes on a message, use the `pubsub-attr` struct tag:
+	//
+	//	type UserEvent struct {
+	//		UserID string `pubsub-attr:"user-id"`
+	//		Action string
+	//	}
+	//
+	//  var topic = pubsub.NewTopic[UserEvent]("user-events", pubsub.TopicConfig{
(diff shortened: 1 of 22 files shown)


runtime/pubsub/internal/types/public.go
```diff
@@ -90,10 +90,49 @@ type TopicConfig struct {
...
+	//  topic.Publish(ctx, &UserEvent{UserID: "1", Action: "login"})  // This message will be delivered before the logout
+	//  topic.Publish(ctx, &UserEvent{UserID: "2", Action: "login"})  // This could be delivered at any time because it has a different user id
+	//  topic.Publish(ctx, &UserEvent{UserID: "1", Action: "logout"}) // This message will be delivered after the first message
+	//
+	// By using OrderingAttribute, the throughput will be limited depending on the cloud provider:
+	//
+	// - AWS: 300 messages per second for the topic (see [AWS SQS Quotas]).
+	// - GCP: 1MB/s for each ordering key (see [GCP PubSub Quotas]).
+	//
+	// Note: OrderingAttribute is not supported for local development.
```
eandre, line 131:
```suggestion
	// Note: OrderingAttribute currently has no effect during local development.
```

runtime/pubsub/internal/types/public.go
```diff
@@ -90,10 +90,49 @@ type TopicConfig struct {
...
+	//
+	// [AWS SQS Quotas]: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html
+	// [GCP PubSub Quotas]: https://cloud.google.com/pubsub/quotas#resource_limits
+	//
+	//publicapigen:drop
```
eandre, line 136: Remove, if we're ready to go live with it

v2/app/testdata/pubsub_err_ordering_attribute_missing.txt
```diff
@@ -0,0 +1,42 @@
+# Verify that pub sub is parsed
+! parse
+err 'PubSub message attributes must not be prefixed with "encore".'
```
eandre, line 3: This seems wrong; there is no such attribute

DomBlack: The v2 parser tests don't actually look at the `err` flag, so I've removed this (it was copy/pasted from a previous test)

runtime/pubsub/internal/types/public.go
```diff
@@ -90,10 +90,49 @@ type TopicConfig struct {
 	// This field is required.
 	DeliveryGuarantee DeliveryGuarantee
 
-	// OrderingKey is the name of the message attribute used to group
```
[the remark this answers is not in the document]
eandre: It's specifying a schema, so it is specifying what is the ordering key. Just like when you create a table in postgres you're defining a "primary key" (by naming a column), not a "primary column".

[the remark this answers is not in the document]
DomBlack: To be clear; we're actually not renaming anything here (officially) as we never released / publicised the original API design. 

I get what you're saying around the Primary Key thing, but given we already confused the field once, and in the GCP API there's a field called "OrderingKey" (AWS it's "GroupingID") I felt we wanted to reduce the risk of that gain.

The other thing I prefer about this name is it's clear it's for an Attribute, meaning if we wanted to change it to a FieldName in the future, there's no abimgilotly as we could add a "OrderingField" option for that.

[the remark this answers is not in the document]
eandre: Yeah let's merge it as-is. Will mull it over

PR #719, opened 2023-05-16
allow passing of RequestCredential for typescript client library

PhakornKiong:
https://encoredev.slack.com/archives/C0356ERKBEH/p1684146428092659

Allow configuration of `RequestInit` so dev have ability to pass along more default configuration of the client.

User of the client library will be able to do and make sure `cookies` are passed for all subsequent request

`const encoreClient = new Client(baseUrl(), requestInit: { credentials: 'include' });`


eandre:
Thanks @PhakornKiong! You can update the golden files with `GOLDEN_UPDATE=1 go test ./...`

PhakornKiong:
> Thanks @PhakornKiong! You can update the golden files with `GOLDEN_UPDATE=1 go test ./...`

i just tried using codescape, and got this error, any idea?

```
root@codespaces-<id>:/workspaces/encore# GOLDEN_UPDATE=1 go test ./...
# encr.dev/pkg/option
pkg/option/option.go:43:9: val.Comparable undefined (type reflect.Value has no field or method Comparable)
pkg/option/option.go:44:14: val.Equal undefined (type reflect.Value has no field or method Equal)
note: module requires Go 1.20
# encr.dev/pkg/paths
pkg/paths/paths.go:89:18: undefined: filepath.IsLocal
pkg/paths/paths.go:236:24: undefined: strings.CutPrefix
note: module requires Go 1.20
FAIL    encr.dev/cli/daemon/dash [build failed]
```

internal/clientgen/testdata/expected_baseauth_typescript.ts
```diff
@@ -72,6 +72,9 @@ export interface ClientOptions {
      */
     fetcher?: Fetcher
 
+	/** indicates whether the user agent should send or receive cookies from the other domain in the case of cross-origin requests. */
+	credentials?: RequestCredentials;
```
eandre, line 76: Having thought about this some more, I think it would be better if we had a field `request?: RequestInit` instead, and we merge those options with the per-request data. That seems more flexible and would support other `RequestInit` fields at the same time, without needing to copy each of those fields to `ClientOptions` over time. What do you think?

PhakornKiong: Make sense, let me try to update it

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -70,6 +70,9 @@ export interface ClientOptions {
      */
     fetcher?: Fetcher
 
+    /** Default RequestInit to be used for the client */
+    requestInit?: RequestInit & { headers?: Record<string, string> };
```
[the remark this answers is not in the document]
PhakornKiong: @eandre Do you have any other comment on this?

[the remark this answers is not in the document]
eandre: Makes sense, thanks!

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -815,6 +820,7 @@ class BaseClient {
     public async callAPI(method: string, path: string, body?: BodyInit, params?: CallParameters): Promise<Response> {
         let { query, ...rest } = params ?? {}
         const init = {
+            ...this.requestInit,
             ...rest,
```
eandre, line 824: If `rest` contains the `headers` field it will entirely overwrite the `headers` field from `requestInit`, whereas we want to merge the two

eandre:
Thanks @PhakornKiong for doing this, I merged it via #737 as I had to make a few minor changes on top. I kept you as the author. Hope that's okay!

PR #726, opened 2023-05-18
clientgen: fix handling of optionals

### LINKED ISSUE #725: Optional query parameter being sent as 'undefined'
Hey Encore team! I’m using the typescript generated client and a route with an optional query parameter. It builds the optional type in TS correctly however it’s still being sent in the query when not included (being sent as “undefined”). Let me know if I might be missing something.

```
// backend
type GetDocumentsRequest struct {
	FiscalYear int `query:"fiscal_year" encore:"optional"`
}
```
```
// frontend codegen client
export interface GetDocumentsRequest {
    FiscalYear?: number
}
```

When it’s building the query it’s always including the parameter even though it could be undefined
```
public async GetDocuments(params: GetDocumentsRequest): Promise<GetDocumentsResponse> {
            // Convert our params into the objects we need for the request
            const query = makeRecord<string, string | string[]>({
                "fiscal_year": String(params.FiscalYear),
            })

            const resp = await this.baseClient.callAPI("GET", `/documents`, undefined, {query})
            return await resp.json() as GetDocumentsResponse
        }
```
Since `fiscal_year` is a number | undefined - using `String()` returns `'undefined'` (as a string) so it’s not being caught by `makeRecord`

eandre:
When a value is optional the ts/js clients need to handle the possibility the value is undefined.

Thanks Ryan Content for the report.
Fixes #725


DomBlack (approved):
Are we not also golden testing the OpenAPI generated client in the e2e-tests? Are we happy with their query strings being generated correctly here?

internal/clientgen/openapi/openapi.go
```diff
@@ -152,7 +153,7 @@ func (g *Generator) newOperationForEncoding(rpc *meta.RPC, method string, reqEnc
 				AllowEmptyValue: true,
 				AllowReserved:   false,
 				Deprecated:      false,
```
DomBlack, line 155: Ohh nice, in the future we could even detect this and include it in the meta

DomBlack (approved):

internal/clientgen/openapi/openapi.go
eandre: Yeah that's a good idea :)

PR #736, opened 2023-05-31
Enable unified streaming trace protocol

eandre:
This updates the runtime to use the new streaming trace protocol, and updates the local development dashboard to support it.


cli/daemon/dash/server.go
```diff
@@ -27,17 +28,23 @@ var upgrader = websocket.Upgrader{
...
 
 // NewServer starts a new server and returns it.
-func NewServer(runMgr *run.Manager, tr *trace.Store) *Server {
+func NewServer(runMgr *run.Manager, tr trace2.Store) *Server {
 	assets, err := fs.Sub(assets, "dashapp/dist")
 	if err != nil {
 		log.Fatal().Err(err).Msg("could not get dash assets")
 	}
 
+	proxy, err := dashproxy.New("https://devdash.encore.dev")
```
DomBlack, line 37: Could we add a `var DevDashURL string = "https://devdash.encore.dev"` as a package level variable - such that during dev builds we can change this to point at a local `npm run dev` version of the dev dash.

I suggest we also add a `var CacheDevDash bool` as well, so we can disable caching

cli/daemon/dash/dashproxy/dashproxy.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "get user cache dir")
+	}
+
+	cache := diskcache.NewWithDiskv(diskv.New(diskv.Options{
+		BasePath:     filepath.Join(cacheDir, "encore", "dashcache"),
+		CacheSizeMax: 1024 * 1024 * 1024, // 1GiB
+		Compression:  diskv.NewGzipCompression(),
+	}))
+	transport := httpcache.NewTransport(cache)
```
DomBlack, line 34: When we deploy a new version of the dev dash to the website, how does the local cache know it needs to be cleared because there is a bug fix available online?

How do we make backwards incompatible changes to the dev dash without _breaking_ old installs? - Because if a users OS has caching disbaled or clears their cache - the next time they start up encore, we'll download an incompatible dashboard.

Maybe the URL needs to include the major Encore CLI version, so we can serve older versions if needed.

cli/daemon/dash/server.go
```diff
@@ -60,21 +68,23 @@ type Server struct {
...
+		//	// We've seen cases where net/http's content type detection gets it wrong
+		//	// for the bundled javascript files. Work around it by specifying it manually.
+		//	if filepath.Ext(path) == ".js" {
+		//		w.Header().Set("Content-Type", "application/javascript")
+		//	}
+		//	fs.ServeHTTP(w, req)
+		//default:
+		//	// Serve the index page for all other paths since we use client-side routing.
+		//	req.URL.Path = "/"
+		//	fs.ServeHTTP(w, req)
```
DomBlack, line 87: Can this be deleted now?

cli/daemon/dash/server.go
```diff
@@ -60,21 +68,23 @@ type Server struct {
 
 func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	path := req.URL.Path
-	fs := http.FileServer(http.FS(s.assets))
+	//fs := http.FileServer(http.FS(s.assets))
```
DomBlack, line 71: Delete?

runtime/appruntime/apisdk/api/reqtrack.go
```diff
@@ -253,57 +300,51 @@ func CheckAuthData(uid model2.UID, userData any) error {
...
 		SpanID: spanID,
 		DefLoc: defLoc,
 	}
 
-	if curr := s.rt.Current(); curr.Trace != nil {
-		curr.Trace.BeginAuth(call, curr.Goctr)
-	}
+	//if curr := s.rt.Current(); curr.Trace != nil {
+	//	curr.Trace.BeginAuth(call, curr.Goctr)
+	//}
```
DomBlack, line 341: Remove this and the next commented out block?

DomBlack (approved):

PR #739, opened 2023-06-01
pubsub: add MethodHandler

eandre:
This adds support for having Pub/Sub handlers that reference a method on a service struct.

### DIFF
--- a/runtime/pubsub/internal/nsq/topic.go
+++ b/runtime/pubsub/internal/nsq/topic.go
@@ -118,13 +118,19 @@ func (l *topic) Subscribe(logger *zerolog.Logger, ackDeadline time.Duration, ret
 		return nil
 	}))
 
-	// connect the consumer to the NSQD
-	err = consumer.ConnectToNSQD(l.addr)
-	if err != nil {
-		panic(fmt.Sprintf("failed to connect %s to nsqd for topic %s: %v", implCfg.EncoreName, l.name, err))
-	}
 	// add the consumer to the known consumers
 	l.consumers[implCfg.EncoreName] = consumer
+
+	go func() {
+		// Allow the rest of the service to initialize before we connect to NSQD.
+		// This is necessary because NSQD is so fast the receiver can process messages
+		// before all package-level initialization functions have been called.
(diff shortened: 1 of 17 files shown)


runtime/pubsub/internal/nsq/topic.go
```diff
@@ -118,13 +118,19 @@ func (l *topic) Subscribe(logger *zerolog.Logger, ackDeadline time.Duration, ret
...
-		panic(fmt.Sprintf("failed to connect %s to nsqd for topic %s: %v", implCfg.EncoreName, l.name, err))
-	}
 	// add the consumer to the known consumers
 	l.consumers[implCfg.EncoreName] = consumer
+
+	go func() {
+		// Allow the rest of the service to initialize before we connect to NSQD.
+		// This is necessary because NSQD is so fast the receiver can process messages
+		// before all package-level initialization functions have been called.
+		time.Sleep(100 * time.Millisecond)
```
DomBlack, line 128: Is there a better way to do this, than "100ms sleep" - this seems like we're just asking for trouble in the future; could there not be a  `apisdk.WaitForInit()` method, which returns true once all services are init'ed

runtime/pubsub/subscription.go
```diff
@@ -278,3 +279,25 @@ func marshalParams[Resp any](json jsoniter.API, resp Resp) []byte {
...
+//	//encore:service
+//	type Service struct {}
+//
+//	func (s *Service) Method(ctx context.Context, msg *Event) error { /* ... */ }
+//
+//	var _ = pubsub.NewSubscription(Topic, "subscription-name", pubsub.SubscriptionConfig[*Event]{
+//		Handler: pubsub.MethodHandler((*MyService).MyMethod),
+//		// ...
+//	})
+func MethodHandler[T, SvcStruct any](handler func(s SvcStruct, ctx context.Context, msg T) error) func(ctx context.Context, msg T) error {
```
DomBlack, line 296: ❤️ Perfect, adding support while keeping type satefy in the IDE

DomBlack (approved):

PR #744, opened 2023-06-05
Add docs for pubsub consumers around ctx deadlines

DomBlack:
This commit adds to the subscription docs the information that the `ctx` of the handler will be cancelled after the `AckDeadline` has passed.

Thanks to @Willyham for the suggestion

### DIFF
--- a/docs/primitives/pubsub.md
+++ b/docs/primitives/pubsub.md
@@ -209,6 +209,10 @@ subscription to a single topic receives the events independently of any other su
 that if one subscription is running very slowly, it will grow a backlog of unprocessed events.
 However, any other subscriptions will still be processing events in real-time as they are published.
 
+The `ctx` passed to the handler function is cancelled when the `AckDeadline` for the subscription is reached, this is the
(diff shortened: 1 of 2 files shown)


docs/primitives/pubsub.md
```diff
@@ -209,6 +209,10 @@ subscription to a single topic receives the events independently of any other su
 that if one subscription is running very slowly, it will grow a backlog of unprocessed events.
 However, any other subscriptions will still be processing events in real-time as they are published.
 
+The `ctx` passed to the handler function is cancelled when the `AckDeadline` for the subscription is reached, this is the
+time when the message is considered to have timed out and can be redelivered to another subscriber. By default if you
+do not specify an `AckDeadline` the default value of 30 seconds is used.
```
eandre, line 214:
```suggestion
The `ctx` passed to the handler function is cancelled when the `AckDeadline` for the subscription is reached.
This is the time when the message is considered to have timed out and can be redelivered to another subscriber.
The timeout defaults to 30 seconds if you don't explicitly configure `AckDeadline`.
```

PR #745, opened 2023-06-05
Merge initial work on supporting HTTP calls between services

DomBlack:
This PR merges the initial work we've been doing on supporting making HTTP calls between services; it is currently behind a feature flag and will not impact existing applications.

There's a lot of new stuff in this PR, so I'll try and summarise the changes here:
- If `ENCORE_DAEMON_DEV` is set, we will now watch the `runtime` source code during `encore run`. This will allow quicker iteration on the runtime library.
- The service discovery config map is now populated for local development, for now always pointing at the daemons proxy server.
- A new experiment, `external-calls` has been introduced; when switched on all service-to-service calls will be routed over HTTP, rather than internally within the running binary. This is experimental and really is not ready for use by _anybody_.
  - A new `CallMeta` data structure exists within the API SDK, which tracks external information about a call, as well as internal meta data. This structure exists to act as a layer between the transport (HTTP in this case) and the handler invocation code. This object is constructed at the highest level `handler` in the HTTP service, and passed via the context to the running handler invocation function.
  - A new `errmarshalling` package has been added. The role of this package is to enable an "internal" version of marshalling to occur for the wire level, which can be used on service-to-service calls to propagate more information between services. For calls not originating from another service in the same Encore app, this package is not used.
  - A new `svcauth` package has been added to the API SDK; it's use is to allow services to sign and verify service-to-service requests. Currently there is just a no-op implementation (which always approves the auth), but we plan to add several authentication options in the future _before_ HTTP based service-to-service calls to live.
  - A new `transport` package has been added to the API SDK; it's designed to abstract away the transport layer from the `svcauth` and server code, allowing us to quickly and easily support other transport methods (such as GRPC) in the future. It's make use case right now is to allow us to read and write meta data on requests being made in that transport.
- The Github actions have been updated, to also run the unit tests for the `runtime`


runtime/appruntime/apisdk/api/call_meta.go
```diff
@@ -0,0 +1,216 @@
...
+	if meta.Internal != nil {
+		// Add a marker to the request to indicate that this is an internal call
+		req.SetMeta("Internal-Call", meta.Internal.SendingService)
+
+		// TODO(domblack): Add the auth UID and data to the request
+
+		// If we're making an internal call, sign the request
+		if err := svcauth.Sign(svcauth.Noop, req); err != nil {
+
+		}
```
eandre, line 99: This is a bit weird; use `_ = ...` if we want to deliberately ignore the error (or is the if statement not filled in?)

runtime/appruntime/apisdk/api/errmarshalling/jsonextension.go
```diff
@@ -0,0 +1,159 @@
...
+	serdeByName = map[string]*jsonMarshaller{}
+
+	// errType is the reflect2 type of the error interface
+	errType = reflect2.Type2(reflect.TypeOf((*error)(nil)).Elem())
+)
+
+const (
+	// TypeKey is the key used to identify the type of an error
+	// used to unmarshal the correct error type
+	TypeKey = "$type$"
```
eandre, line 26: Maybe use `@type` to align with Protobuf's JSON mapping convention?

runtime/appruntime/apisdk/app/app.go
```diff
@@ -41,11 +41,19 @@ func (app *App) Run() error {
 
 	app.Start()
 
+	// Begin serving requests.
+	serveCh := make(chan error, 1)
+	go func() {
+		serveCh <- app.api.Serve(ln)
+	}()
```
eandre, line 48: Why this change? Seems like you want to initialize the services before you begin serving

runtime/appruntime/exported/config/config.go
```diff
@@ -69,10 +74,21 @@ type Gateway struct {
...
+	Protocol SvcProtocol `json:"protocol"`
 }
 
+type SvcProtocol uint8
+
+const (
+	Http SvcProtocol = iota
+	ProtoJson
+	ProtoBinary
+	GRPC
```
eandre, line 89: Should these be strings instead since it's exported as JSON? Also the values don't really make sense, should we start with just HTTP for now?

v2/parser/apis/api/apienc/encoding.go
```diff
@@ -171,6 +171,13 @@ func DescribeResponse(errs *perr.List, responseSchema schema.Type) *ResponseEnco
 		return &ResponseEncoding{}
 	}
 
+	// Check for reserved header prefixes
+	for _, field := range fields[Header] {
+		if strings.HasPrefix(field.WireName, "x-encore") {
```
eandre, line 176: Is the wire name lowercase, or `X-Encore`? Maybe worth lowercasing to be safe? Also maybe should be `x-encore-` and not just `x-encore`

v2/parser/apis/api/apienc/encoding.go
```diff
@@ -251,6 +258,13 @@ func DescribeRequest(errs *perr.List, requestSchema schema.Type, httpMethods ...
 			return nil
 		}
 
+		// Check for reserved header prefixes
+		for _, field := range fields[Header] {
+			if strings.HasPrefix(field.WireName, "x-encore") {
```
eandre, line 263: Same here

runtime/appruntime/apisdk/api/errmarshalling/jsonextension.go
DomBlack: Sure, I was following another convention that I thought was GraphQL, but in hindsight I'm not actually sure. Happy to change though!

runtime/appruntime/apisdk/app/app.go
DomBlack: So this is due to the test app, it has an `init` function which calls an API, with external calls this fails because the service isn't actually running yet - which then causes it's service to fail to start up

eandre (approved):

PR #746, opened 2023-06-06
Transmit auth information for external service-to-service calls

DomBlack:
This PR adds support for transmitting the user ID and auth data of an authenticated user between services when services are communicating over the wire.

It adds new validation checks at compile time for verifying that auth data is marshallable, which is a new restriction on auth data types.


runtime/appruntime/apisdk/api/reqtrack.go
```diff
@@ -281,6 +281,15 @@ func GetCallOptions(ctx context.Context) *CallOptions {
 
 var RegisteredAuthDataType reflect.Type
```
eandre, line 282: Document this needs to be a pointer type

runtime/appruntime/apisdk/api/authmarshalling/marshal.go
```diff
@@ -0,0 +1,40 @@
+package authmarshalling
+
+import (
+	jsoniter "github.com/json-iterator/go"
+)
+
+var jsonInstance jsoniter.API
```
eandre, line 7: As discussed let's use standard JSON instead, and round-trip auth data via JSON instead

runtime/appruntime/apisdk/api/call_meta.go
```diff
@@ -90,7 +95,18 @@ func (meta CallMeta) AddToRequest(req transport.Transport) error {
 		// Add a marker to the request to indicate that this is an internal call
 		req.SetMeta("Internal-Call", meta.Internal.SendingService)
 
-		// TODO(domblack): Add the auth UID and data to the request
+		// Add the auth data
+		if meta.Internal.AuthUID != "" {
+			req.SetMeta("Auth-UID", meta.Internal.AuthUID)
```
eandre, line 100: `X-Encore-Auth-UID` -> `X-Encore-UserID`

runtime/appruntime/apisdk/api/call_meta.go
```diff
@@ -90,7 +95,18 @@ func (meta CallMeta) AddToRequest(req transport.Transport) error {
...
+
+			if meta.Internal.AuthData != nil {
+				authData, err := authmarshalling.Marshal(meta.Internal.AuthData)
+				if err != nil {
+					return errs.B().Cause(err).Msg("failed to marshal auth data").Err()
+				}
+				req.SetMeta("Auth-Data", authData)
```
eandre, line 107: `X-Encore-Auth-Data` -> `X-Encore-AuthData`

runtime/appruntime/apisdk/api/call_meta.go
```diff
@@ -142,7 +158,18 @@ func MetaFromRequest(req transport.Transport) (meta CallMeta, err error) {
 			SendingService: sendingService,
 		}
 
-		// TODO(domblack): Read the auth UID and data from the request
+		// Pull the auth data out of the request
+		if uid, found := req.ReadMeta("Auth-UID"); found && uid != "" {
+			meta.Internal.AuthUID = uid
+
+			if data, found := req.ReadMeta("Auth-Data"); found && data != "" {
```
eandre, line 165: Same

runtime/beta/auth/auth.go
```diff
@@ -60,7 +61,23 @@ func (mgr *Manager) Data() interface{} {
 // the empty string then data may not be nil. If these requirements are not met,
 // API calls made with these options will not be made and will immediately return
 // a client-side error.
+//
+// Note: this function will panic if the type of the data parameter is not
+// the same type returned by the auth handler. If the auth handler does not
+// return any data, then this value must be nil.
 func WithContext(ctx context.Context, uid UID, data interface{}) context.Context {
+	// If we're being passed a uid and data, make sure the data is valid.
+	if uid != "" || data != nil {
```
eandre, line 70: Revert and preserve existing semantics (return a client-side error on the request)

runtime/appruntime/exported/experiments/set.go
```diff
@@ -0,0 +1,127 @@
...
+type Set struct {
+	enabled map[Name]struct{}
+}
+
+// NewForRuntime constructs a new Experiments object from both the static and runtime configs.
+//
+// It is used by the Encore runtime to construct the Experiments set in a running application.
+//
+// Unknown experiments are ignored.
+func NewForRuntime(static *config.Static, runtime *config.Runtime) *Set {
```
eandre, line 23: It's a bit confusing that "ForRuntime" checks the compiled-in config, and "ForCompiler" checks the environment variable at runtime. It's not wrong but it's confusing.

Maybe rename this `FromConfig` and rename the other to `FromEnviron`?

runtime/appruntime/shared/health/health.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/rs/zerolog/log"
+	"golang.org/x/exp/slices"
+)
+
+// CheckRegistry is a registry of health checks from the API and Infra SDKs
+// and other parts of the runtime.
+type CheckRegistry struct {
+	checks []Check
+	m      sync.Mutex
```
eandre, line 16: Nit: mutex goes before the fields it protects
```suggestion
	m      sync.Mutex
	checks []Check
```

runtime/appruntime/shared/health/health.go
```diff
@@ -0,0 +1,115 @@
...
+	c.checks = append(c.checks, check)
+}
+
+// RegisterFunc registers a new health check from a function with a given name
+//
+// Checks must complete within 5 seconds, otherwise
+// they will be terminated and considered failed.
+//
+// Checks can be called at any time and could have
+// multiple goroutines calling them concurrently.
```
eandre, line 45: Nit: no need to duplicate all the documentation; better to refer to the other.

```suggestion
// RegisterFunc is like Register but constructs the Check from the given name and func.
```

runtime/appruntime/shared/health/health.go
```diff
@@ -0,0 +1,115 @@
...
+func (c *CheckRegistry) RunAll(ctx context.Context) []CheckResult {
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	checks := c.GetChecks()
+
+	// Run all checks in parallel.
+	results := make(chan []CheckResult, len(checks))
+	var wg sync.WaitGroup
+	wg.Add(len(checks))
```
eandre, line 69: Use an error group?

runtime/appruntime/exported/config/config.go
```diff
@@ -62,10 +66,12 @@ type Runtime struct {
 	// If zero, it shuts down immediately.
 	ShutdownTimeout time.Duration `json:"shutdown_timeout"`
 
-	// ExperimentUseExternalCalls is a flag that indicates whether the app should
-	// use external calls for all RPCs. This is a temporary flag that is only
-	// for testing purposes and will be removed in the future.
-	ExperimentUseExternalCalls bool `json:"experiment_use_external_calls,omitempty"`
+	// EnabledExperiments is a list of experiments that are enabled for this app
+	// which impact runtime behaviour and where turned on _after_ compilation.
```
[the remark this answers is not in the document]
DomBlack: By the Encore platform - so for instance we have a binary already built, but in the PR environment we switch on some experiment, which we don't use during prod.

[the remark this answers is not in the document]
eandre: Ah, makes sense. Should we call it something else? Like `DynamicExperiments` (in contrast with "static experiments" that were compiled in)?

runtime/appruntime/shared/health/health.go
DomBlack: An error group only returns the first error, I'd like to be able to report on all errors (or successes), so I'd still want channel, at which point I might as well just use a WaitGroup

PR #756, opened 2023-06-12
[ENC-1575] Implement Preshared Key Auth for Service to Service calls

DomBlack:
This commit implements a new preshared key authentication mechisam for service-to-service calls made over the wire.

### DIFF
--- a/runtime/appruntime/apisdk/api/svcauth/encoreauth.go
+++ b/runtime/appruntime/apisdk/api/svcauth/encoreauth.go
@@ -1,0 +1,160 @@
+package svcauth
+
+import (
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"go.encore.dev/platform-sdk/pkg/auth"
+	"golang.org/x/crypto/sha3"
+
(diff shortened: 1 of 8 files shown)


runtime/appruntime/apisdk/api/svcauth/encoreauth.go
```diff
@@ -0,0 +1,160 @@
...
+		default:
+			// Read all values for this key, and sort them
+			values, found := req.ReadMetaValues(key)
+			if !found {
+				return "", errs.B().Code(errs.Internal).Msg("failed to read metadata value").Err()
+			}
+			sort.Strings(values)
+
+			for _, value := range values {
+				if _, err := fmt.Fprintf(hash, "%s=%s\n", key, value); err != nil {
```
eandre, line 147: Can values contain newlines or equal signs? Do we need to escape them?

DomBlack: I guess yeah it can be, we control all values right now so it's a non-issue - but I'll add a ticket to come back and fix it before go-live

PR #758, opened 2023-06-12
Fix Stack traces on panics

DomBlack:
This commit fixes an issue where panic stack traces where not be propagated via the `RoundTrip` function, because `gob` did not know about the stack type.

This also improves panic stacks by removing all frames back to the user's code which caused the panic

### DIFF
--- a/runtime/appruntime/exported/stack/stack.go
+++ b/runtime/appruntime/exported/stack/stack.go
@@ -16,11 +16,38 @@ type Stack struct {
 }
 
 func Build(skip int) Stack {
+	return BuildWithFilters(skip+1, nil)
+}
+
+func BuildWithoutGoRuntime(skip int) Stack {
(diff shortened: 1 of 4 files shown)


ekerfelt (approved):

runtime/appruntime/exported/stack/stack.go
```diff
@@ -16,13 +16,40 @@ type Stack struct {
 }
 
 func Build(skip int) Stack {
+	return BuildWithFilters(skip+1, nil)
+}
+
+func BuildWithoutGoRuntime(skip int) Stack {
```
eandre, line 22: Maybe make this a built-in filter function instead? `func ExcludeRuntime(pc uintptr) bool` or whatever, with usage `stack.BuildWithFilters(1, stack.ExcludeRuntime)`?

PR #759, opened 2023-06-13
.github: add docker image to release process

eandre:


.github/workflows/release.yml
```diff
@@ -79,8 +79,74 @@ jobs:
...
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}
+            type=ref,event=pr
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .github/dockerimg
+          platforms: linux/amd64,linux/arm64
```
DomBlack, line 137: Do we need darwin as well here?

```suggestion
          platforms: linux/amd64,linux/arm64,darwin/amd64,darwin/arm64
```

.github/workflows/release.yml
```diff
@@ -79,8 +79,74 @@ jobs:
...
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: encoredotdev/encore
```
DomBlack, line 120: damn was `encoredev` not free?

eandre: Nope, unfortunately not, but this matches our Twitter handle so it's not without precedent

DomBlack (approved):

PR #767, opened 2023-06-15
.github/dockerimg: fix platform mismatch

eandre:
Use the target platform instead of the build platform.


.github/dockerimg/Dockerfile
```diff
@@ -1,12 +1,12 @@
 # syntax=docker/dockerfile:1.4
-FROM --platform=$BUILDPLATFORM ubuntu:22.04 AS build
+FROM --platform=$TARGETPLATFORM ubuntu:22.04 AS build
```
DomBlack, line 2: This should be the build platform, as that's where we're building the code I'd though

DomBlack (approved):

PR #770, opened 2023-06-15
v2/parser: fix implicit database usage

eandre:


v2/parser/infra/sqldb/implicit.go
```diff
@@ -83,11 +83,16 @@ func ComputeImplicitUsage(errs *perr.List, pkgs []*pkginfo.Package, binds []reso
 
 				switch qn.Name {
 				case "Exec", "ExecTx", "QueryRow", "QueryRowTx", "Query", "QueryTx", "Begin":
+					validUsage := false
 					if bind, ok := findBind(file.Pkg.ImportPath); ok {
```
DomBlack, line 87: Should bind finding like this happen later in the `app` validation package - once we've decided what the services are;

How does this handle:

`/foo/migrations` as an implicit database and then `/foo/bar` calling `sqldb.Query` ?

PR #773, opened 2023-06-21
Re-add schema validation for Header & Query String fields

DomBlack:
This PR re-adds schema validation for header and query string fields, which in the v1 parser we had, but lost during the v2 rewrite.

It also fixes an issue where we where double sending compiler errors to the CLI.

### DIFF
--- a/internal/optracker/optracker.go
+++ b/internal/optracker/optracker.go
@@ -161,7 +161,6 @@ func (t *OpTracker) refresh() {
 		return ops[i].start.Before(ops[j].start)
 	})
 
-	// var errlistToSend *errlist.List
 	for _, o := range ops {
 		started := o.start.Before(now)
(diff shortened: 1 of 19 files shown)


internal/optracker/optracker.go
```diff
@@ -194,10 +200,15 @@ func (t *OpTracker) refresh() {
...
-	}
+
+	// For now we don't send the error list to the stream
+	// as the OpTracker is only used for the initial build
+	// and we don't want to send the error list to the stream
+	// because the Run command will send the error list to the stream
+	// if errlistToSend != nil {
+	// 	// We sent this after we clear and repaint the screen
+	// 	errlistToSend.SendToStream(t.stream)
+	// }
```
eandre, line 211: Remove?

eandre (approved):

PR #775, opened 2023-06-26
Introduce no-op infrastructure

DomBlack:
This commit introduces the concept of no-op infrastructure. This is required so that Encore can build a single binary containing all the services, however deploy a container with just a single service running inside it - all infrastructure used by other services which are not running will not work as the container will not be provisioned permissions to that infrastructure. Instead any operations on that infrastructure should result in an error.

### DIFF
--- a/runtime/config/manager_internal.go
+++ b/runtime/config/manager_internal.go
@@ -49,21 +49,21 @@ func NewManager(rt *reqtrack.RequestTracker, json jsoniter.API) *Manager {
 	}
 }
 
-func (m *Manager) getComputedCUE(serviceName string) ([]byte, error) {
+func (m *Manager) getComputedCUE(serviceName string) (jsonBytes []byte, found bool, err error) {
 	if m == nil {
-		return nil, fmt.Errorf("config subsystem has not been initialized")
+		return nil, true, fmt.Errorf("config subsystem has not been initialized")
 	}
 
 	// Fetch the raw JSON config for this service
 	envVar := encoreenv.Get(envName(serviceName))
 	if envVar == "" {
-		return nil, fmt.Errorf("configuration for service `%s` not found, expected it in environmental variable %s", serviceName, envName(serviceName))
+		return nil, false, fmt.Errorf("configuration for service `%s` not found, expected it in environmental variable %s", serviceName, envName(serviceName))
 	}
 	cfgBytes, err := base64.RawURLEncoding.DecodeString(envVar)
 	if err != nil {
-		return nil, fmt.Errorf("failed to decode configuration for service `%s`: %v", serviceName, err)
+		return nil, true, fmt.Errorf("failed to decode configuration for service `%s`: %v", serviceName, err)
 	}
-	return cfgBytes, nil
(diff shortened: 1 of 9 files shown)


runtime/config/manager_internal.go
```diff
@@ -49,21 +49,21 @@ func NewManager(rt *reqtrack.RequestTracker, json jsoniter.API) *Manager {
 	}
 }
 
-func (m *Manager) getComputedCUE(serviceName string) ([]byte, error) {
+func (m *Manager) getComputedCUE(serviceName string) (jsonBytes []byte, found bool, err error) {
```
eandre, line 52: Instead of a separate return, maybe it would be clearer to introduce a specific error type for the case where it's not found:

```go
type NotFoundError struct {
    Service string
    EnvVar string
}

func (e NotFoundError) Error() string {
		return fmt.Sprintf("configuration for service `%s` not found, expected it in environmental variable %s", e.Service, e.EnvVar)
}
```

And then use `errors.As` to check for it

runtime/pubsub/internal/noop/topic.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"encore.dev/appruntime/exported/config"
+	"encore.dev/pubsub/internal/types"
+)
+
+type Topic struct{}
+
+var ErrNoop = errors.New(
+	"pubsub: this service is not configured to use this topic. " +
+		"Use pubsub.TopicRef in the service to get a reference and access to the topic from this service",
```
eandre, line 18: If a service isn't configured to use the topic, how would it get a `TopicRef` in the first place? A `TopicRef` can't be passed over the wire. I think we should just drop the second sentence from the error.

runtime/storage/sqldb/stdlib_noop_internal.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	registerNoopDriverOnce sync.Once
+)
+
+const noopDriverName = "__encore_noop"
+
+// noopDriver is a driver.Driver that returns will "connect" and manage
+// a connection to a database without an error, however whenver you try and use
```
eandre, line 16:
```suggestion
// a connection to a database without an error, however whenever you try and use
```

runtime/pubsub/internal/noop/topic.go
DomBlack: I thought the whole point of a TopicRef was to allow you to say "this service needs a reference to this topic"

eandre: True, but I suppose the same could be said for all other infrastructure. "sqldb: this service is not configured to use this database. Use `sqldb.Named` in the service to get a reference and access to the database."

runtime/config/manager_internal.go
DomBlack: Given this is simply for an internal one use function, doesn't it add far more complication to create a new type, and support errors as on it?

eandre: yeah fair :)

PR #776, opened 2023-06-28
cli: add device auth login flow

eandre:
This adds support for the OAuth2 device authorization login flow. The flow supports logging in with a different device from where the browser runs, allowing use of the Encore CLI in environments that don't run browsers, like developing over SSH or using GitHub Codespaces/Gitpod.

Thanks Peter Stewart for the suggestion.

### DIFF
--- a/cli/cmd/encore/auth/auth.go
+++ b/cli/cmd/encore/auth/auth.go
@@ -4,14 +4,11 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"time"
 
-	"github.com/briandowns/spinner"
 	"github.com/spf13/cobra"
 
 	"encr.dev/cli/cmd/encore/cmdutil"
 	"encr.dev/cli/cmd/encore/root"
-	"encr.dev/cli/internal/browser"
 	"encr.dev/cli/internal/login"
 	"encr.dev/internal/conf"
 )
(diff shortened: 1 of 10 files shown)


cli/cmd/encore/auth/auth.go
```diff
@@ -30,14 +27,14 @@ func init() {
 
 		DisableFlagsInUseLine: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := DoLogin(); err != nil {
+			if err := DoLogin(DeviceAuth); err != nil {
```
ekerfelt, line 30: why deviceauth here?

ekerfelt (approved):

cli/cmd/encore/auth/auth.go
eandre: Because it's more general. I kept the original flow around for the `encore app create` flow because it doesn't require a bunch of instructions that are unrelated to the task at hand (creating an app). If you explicitly run `encore auth login` that's not as much of a problem: you're explicitly invoking the login flow.

If the auto-detection is wrong we don't currently have a way of selecting a specific flow, so it seemed easier to default this flow to always use the device auth flow.

ekerfelt: fair. I guess an optional flag could be an option too.

eandre: Yeah, it's just not very discoverable and if you have an issue it's not super-clear what to do. I figured it's easier to just always use the new flow in the `encore auth login` case.

PR #779, opened 2023-06-29
Refactor proc into procgroup

DomBlack:
This PR is prepping for when locally we run multiple processes of the Encore application, one for each microservice and one for each gateway.

This starts by refactoring our existing `Proc` struct into `ProcGroup` while keeping most of the API the same.

Then within a ProcGroup we introduce the concept of `Proc`'s which can be started independently of each other.

### DIFF
--- a/cli/daemon/run/proc_groups.go
+++ b/cli/daemon/run/proc_groups.go
@@ -1,0 +1,107 @@
+package run
+
+import (
+	"context"
+	"io"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/hashicorp/yamux"
+	"github.com/rs/zerolog"
+
+	"encore.dev/appruntime/exported/config"
+	"encore.dev/appruntime/exported/experiments"
+	"encr.dev/cli/daemon/internal/sym"
+	meta "encr.dev/proto/encore/parser/meta/v1"
+)
+
+// ProcGroup represents a running Encore process.
+type ProcGroup struct {
+	ID          string           // unique process id
+	Run         *Run             // the run the process belongs to
+	Pid         int              // the OS process id
+	Meta        *meta.Data       // app metadata snapshot
+	Started     time.Time        // when the process started
+	Experiments *experiments.Set // enabled experiments
+
+	ctx      context.Context
+	log      zerolog.Logger
+	exit     chan struct{} // closed when the process has exited
+	cmd      *exec.Cmd
+	reqWr    *os.File
+	respRd   *os.File
+	buildDir string
+	Client   *yamux.Session
+	authKey  config.EncoreAuthKey
+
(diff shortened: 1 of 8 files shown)


cli/daemon/run/proc_groups.go
```diff
@@ -0,0 +1,297 @@
...
+	ID          string           // unique process id
+	Run         *Run             // the run the process belongs to
+	Meta        *meta.Data       // app metadata snapshot
+	Experiments *experiments.Set // enabled experiments
+
+	Gateway *Proc // the API gateway process
+
+	procMu       sync.Mutex // protects both AllProcesses and runningProcs
+	procCond     sync.Cond  // used to signal a change in runningProcs
+	AllProcesses []*Proc    // all processes in the group
```
eandre, line 41: Should this be public if there's a mutex protecting it?

cli/daemon/run/proc_groups.go
```diff
@@ -0,0 +1,297 @@
...
+	buildDir   string
+	workingDir string
+
+	authKey   config.EncoreAuthKey
+	sym       *sym.Table
+	symErr    error
+	symParsed chan struct{} // closed when sym and symErr are set
+}
+
+// Done returns a channel that is closed when the process has exited.
```
eandre, line 56:
```suggestion
// Done returns a channel that is closed when all processes in the group have exited.
```

cli/daemon/run/proc_groups.go
```diff
@@ -0,0 +1,297 @@
...
+
+	return c
+}
+
+// Close closes the process and waits for it to shutdown.
+// It can safely be called multiple times.
+func (pg *ProcGroup) Close() {
+	var wg sync.WaitGroup
+	wg.Add(len(pg.AllProcesses))
+	for _, p := range pg.AllProcesses {
```
eandre, line 79: This needs to be guarded according to the docs

cli/daemon/run/proc_groups.go
```diff
@@ -0,0 +1,297 @@
...
+// If the process has already been started, this is a no-op.
+func (p *Proc) Start() error {
+	if !p.Started.CompareAndSwap(false, true) {
+		return nil
+	}
+
+	p.group.procMu.Lock()
+	defer p.group.procMu.Unlock()
+
+	if err := p.cmd.Start(); err != nil {
```
eandre, line 228: We can lock the proc group after `p.cmd.Start()` I guess?

DomBlack: I did it before the `Start` so `runningProcs` would be correct for all readers at the time of read, if we took the lock after `Start` it might be possible for a reader to read `0` just before we increment it to `1`.

eandre (approved):

PR #780, opened 2023-06-30
External calls runs per svc processes

DomBlack:


cli/daemon/run/proc_groups.go
```diff
@@ -71,6 +73,18 @@ func (pg *ProcGroup) Done() <-chan struct{} {
 	return c
 }
 
+// Start starts all the processes in the group.
+func (pg *ProcGroup) Start() (err error) {
+	for _, p := range pg.allProcesses {
```
DomBlack, line 78: There's a deadlock I need to solve here in the next PR if I use the `procMu` lock, as I want `p.Start` to be synchronous, yet it also wants the lock.

cli/daemon/run/proc_groups.go
```diff
@@ -165,58 +210,61 @@ func (pg *ProcGroup) NewAllInOneProc(binPath string, environ []string, secrets m
...
+	}
+
+	p.cmd = cmd
 
 	return nil
 }
 
+func (pg *ProcGroup) NewProcForGateway(params *NewProcParams) error {
+	// TOOD: implement gateway processes
+	return pg.NewAllInOneProc(params)
```
DomBlack, line 251: This function will be implemented in the next PR

cli/daemon/run/run.go
```diff
@@ -389,21 +416,53 @@ func (r *Run) StartProcGroup(params *StartProcGroupParams) (p *ProcGroup, err er
...
 	p.procCond.L = &p.procMu
 	go p.parseSymTable(params.BinPath)
 
-	if err := p.NewAllInOneProc(params.BinPath, params.Environ, params.Secrets, params.ServiceConfigs); err != nil {
-		return nil, err
+	newProcParams := &NewProcParams{
+		BinPath: params.BinPath,
+		Environ: params.Environ,
+		Secrets: params.Secrets,
+		Configs: params.ServiceConfigs,
```
DomBlack, line 428: I think I can remove these next week

cli/daemon/run/infra/infra.go
```diff
@@ -311,3 +314,76 @@ func (rm *ResourceManager) UpdateConfig(cfg *config.Runtime, md *meta.Data, dbPr
 
 	return nil
 }
+
+// SQLConfig returns the SQL server and database configuration for the given database.
+func (rm *ResourceManager) SQLConfig(db *meta.SQLDatabase) (config.SQLServer, config.SQLDatabase, error) {
```
DomBlack, line 319: These changes will make more sense when you read `runtime_config.go` so I suggest you come back to this file after reading the changes to `proc_groups.go` and `runtime_config.go`

cli/daemon/run/runtime_config.go
```diff
@@ -0,0 +1,411 @@
...
+		PlatformOrLocalID() string
+	}
+
+	// The infra manager to use
+	InfraManager interface {
+		SQLConfig(db *meta.SQLDatabase) (config.SQLServer, config.SQLDatabase, error)
+		PubSubTopicConfig(topic *meta.PubSubTopic) (config.PubsubProvider, config.PubsubTopic, error)
+		PubSubSubscriptionConfig(topic *meta.PubSubTopic, sub *meta.PubSubTopic_Subscription) (config.PubsubSubscription, error)
+		RedisConfig(redis *meta.CacheCluster) (config.RedisServer, config.RedisDatabase, error)
+	}
```
DomBlack, line 44: These are interfaces as this process is complicated enough I want to be able to mock them in tests. I just ran out of time to write the tests today.

It might also be useful to keep these interfaces outside of the needs of unit tests, so things like our CI/CD pipeline can reuse the EnvGenerator without needing to have modifications to the `ResourceManager` struct made

cli/daemon/run/runtime_config.go
```diff
@@ -0,0 +1,411 @@
...
+		if err != nil {
+			return "", errors.Wrap(err, "failed to generate service discovery map")
+		}
+
+		runtimeCfg.ServiceDiscovery = svcDiscovery
+	}
+
+	sqlServers := newIndexTracker[config.SQLServer]()
+	pubsubProviders := newIndexTracker[config.PubsubProvider]()
+	redisServers := newIndexTracker[config.RedisServer]()
```
DomBlack, line 155: So this is why the Interfaces return values and not pointers; these make the config types `comparable` which allow me to basically say "have I seen this SQLServer before, if yes return it's `idx` in the list, otherwise add it to the list now and return `len(list)`"

cli/daemon/run/runtime_config.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+		}
+	})
+}
+
+// ListenAddresses is a list of listen address and port numbers for services to run on
+type ListenAddresses struct {
+	Gateway  netip.AddrPort            // The entrypoint to the application
+	Services map[string]netip.AddrPort // Map from service name to listen address
+}
```
DomBlack, line 325: I kept this as it's own type, so we can use it as a type for managing service discovery map generation outside of the CLI too.

Hence the use of `netip.AddrPort` instead of `int16`, it allows for the option of the host being changed

runtime/appruntime/apisdk/api/handler.go
```diff
@@ -440,21 +440,11 @@ type CallContext struct {
...
-			}
 
-			return d.externalCall(c, config.Service{
-				Name:        d.Service,
-				URL:         c.server.runtime.APIBaseURL,
-				Protocol:    config.Http,
-				ServiceAuth: c.server.runtime.ServiceAuth[0],
-			}, req)
+		if !found {
+			// FIXME(domblack): In the future add a check to see if we're hosting this service within this instance
```
DomBlack, line 446: I'll do this next week

runtime/appruntime/apisdk/api/handler.go
```diff
@@ -691,15 +681,20 @@ func (d *Desc[Req, Resp]) externalCall(c CallContext, service config.Service, re
 				if err != nil {
 					return resp, errs.B().Code(errs.Internal).Msg("request failed: unable to read response").Err()
 				}
-				return resp, errs.B().Code(errs.Internal).Msg("request failed: " + string(bodyBytes)).Err()
+
+				if len(bodyBytes) == 0 {
+					return resp, errs.B().Code(errs.Internal).Msgf("request failed: status %d (%s)", httpResp.StatusCode, httpResp.Status).Err()
+				} else {
+					return resp, errs.B().Code(errs.Internal).Msgf("request failed: status %d: %s", httpResp.StatusCode, string(bodyBytes)).Err()
+				}
```
DomBlack, line 689: Turn's out we had a shit error experience if there was no body returned, so this is just to make it easier to understand what's gong on

runtime/appruntime/apisdk/app/setup.go
```diff
@@ -0,0 +1,27 @@
...
+
+func Listen() (net.Listener, error) {
+	listenAddr := encoreenv.Get("ENCORE_LISTEN_ADDR")
+	if listenAddr != "" {
+		addrPort, err := netip.ParseAddrPort(listenAddr)
+		if err != nil {
+			return nil, err
+		}
+		return net.Listen("tcp", addrPort.String())
+	}
```
DomBlack, line 20: This is so we just work with the `netip.AddrPort` data type throughout locally

runtime/appruntime/exported/config/config.go
```diff
@@ -51,7 +51,7 @@ type Runtime struct {
 	RedisServers     []*RedisServer          `json:"redis_servers,omitempty"`
 	RedisDatabases   []*RedisDatabase        `json:"redis_databases,omitempty"`
 	Metrics          *Metrics                `json:"metrics,omitempty"`
-	ServiceDiscovery map[string]Service      `json:"services,omitempty"`
+	ServiceDiscovery map[string]Service      `json:"service_discovery,omitempty"`
```
DomBlack, line 54: I'll need `services` next week so I can tell an instance what services it is hosting.

cli/daemon/run/proc_groups.go
eandre: I guess we can have a private method `p.start()` that assumes the lock is held, that can be called from here

e2e-tests/app_test.go
```diff
@@ -157,28 +154,20 @@ func (l testRunLogger) RunStderr(r *Run, line []byte) {
...
-			continue
-		}
+	go srv.Serve(ln)
 
-		go io.Copy(conn, clientConn)
-		go io.Copy(clientConn, conn)
-	}
+	// So we're going to sleep here to give the server a chance to start up.
+	// as otherwise we see the application startup before this proxy is ready
+	time.Sleep(1 * time.Second)
```
eandre, line 170: Is there a better solution than this? What exactly is the impact of this (i.e. what does it delay)?

e2e-tests/echo_app_test.go
```diff
@@ -325,6 +327,7 @@ func doTestEndToEndWithApp(t *testing.T, env []string) {
 		{
 			w := httptest.NewRecorder()
 			req := httptest.NewRequest("GET", "/echo.Env", nil)
+			log.Println("making request")
```
eandre, line 330: Remove

e2e-tests/testscript_test.go
```diff
@@ -206,6 +206,8 @@ func doRun(t *testing.T, experiments []string) {
 						}
 					}
 				}
+
+				time.Sleep(1 * time.Second)
```
eandre, line 210: If this is needed this test is probably quite flakey already

runtime/appruntime/apisdk/api/handler.go
eandre: Makes sense. Maybe just use `httpResp.Status` in both cases instead of `status %d`? `httpResp.Status` includes the HTTP status code but also what it is ("500 Internal Server Error")

runtime/appruntime/apisdk/api/handler.go
```diff
@@ -691,15 +681,20 @@ func (d *Desc[Req, Resp]) externalCall(c CallContext, service config.Service, re
...
 			}
 		}
 
 		return d.DecodeExternalResp(httpResp, c.server.json)
 	})()
 
 	if respErr != nil {
-		c.server.rootLogger.Err(respErr).Msg("call failed")
+		c.server.rootLogger.Err(respErr).Str("calling", fmt.Sprintf("%s.%s", d.Service, d.Endpoint)).Str("url", reqURL).Msg("call failed")
```
eandre, line 697: `calling` is a confusing name since it's also a present tense verb, but the call already failed. How about `target`?

```suggestion
		c.server.rootLogger.Err(respErr).Str("target", fmt.Sprintf("%s.%s", d.Service, d.Endpoint)).Str("url", reqURL).Msg("call failed")
```

runtime/appruntime/apisdk/app/setup.go
```diff
@@ -0,0 +1,27 @@
...
+	"net"
+	"net/netip"
+	"os"
+	"strconv"
+
+	"encore.dev/appruntime/shared/encoreenv"
+)
+
+func Listen() (net.Listener, error) {
+	listenAddr := encoreenv.Get("ENCORE_LISTEN_ADDR")
```
eandre, line 13: Why two different ways of listening?

runtime/appruntime/apisdk/service/service.go
```diff
@@ -70,7 +71,21 @@ func doSetupService[T any](mgr *Manager, decl *Decl[T]) (err error) {
...
 	}
 
-	i, err := setupFn()
+	var i *T
+
+	// backoff retry for service initialization
+	//
+	// this is to handle the case where a service is dependent on another service during a cold start
+	// of the application, and the dependent service is not yet ready
+	for attempt := 0; attempt < 5; attempt++ {
```
eandre, line 80: I don't think we should do this; we haven't documented or required that the initialization function be idempotent or designed to handle being retried within the same process

e2e-tests/app_test.go
DomBlack: Without it we have flacky end to end tests. Normally at least one of them fails. 

This is because the proxy starts _after_ the test try’s to call the endpoint.

DomBlack: I couldn’t think of a bretter way to solve it today

e2e-tests/testscript_test.go
DomBlack: Yeah it is. This only happened when I brought in the change from your GRPC branch so I’m not sure why it made it flakey; I assume because of the background go routine now needing to start and get to a point of seving

runtime/appruntime/apisdk/service/service.go
DomBlack: I don’t agree for now. The reason is this change broke the test app without; and most things in there are because customers have also got that behaviour. In which case we either need to document a breaking change and give people time to migrate - or support this and document a new retry method. 

We could maybe only retry on connection failed errors; so the retry only applies to errors coming from our own external calls

eandre: Let's discuss on Monday, but this is mainly a concern when developing locally when all services spin up at the same time and there's only one instance of them. A proxy that waits for the service to start up would solve this.

DomBlack: This wouldn’t solve a cold start in the cloud though; sign up a new environment of an existing app with this behaviour. 

There’s no proxy waiting to delay the requests there.

eandre: Relying on another service being up to start up is always going to be error-prone. But you can implement such retries yourself if it's appropriate for your use case.

pkg/svcproxy/svcproxy.go
```diff
@@ -0,0 +1,154 @@
...
+			logger.Err(err).Msg("error serving")
+		}
+	}()
+
+	return proxy, nil
+}
+
+// Close stops running the service proxy
+func (p *SvcProxy) Close() {
+	if p.httpServer != nil {
```
eandre, line 68: When would this be nil?

pkg/svcproxy/svcproxy.go
```diff
@@ -0,0 +1,154 @@
...
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}}).DialContext,
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		},
+		Director: func(request *http.Request) {
```
eandre, line 110: Probably better to use `Rewrite` instead of `Director` and call `req.SetURL(target)`

pkg/svcproxy/svcproxy.go
DomBlack: in an earlier iteration there was a path when it could have been, gone now

eandre (approved):

PR #784, opened 2023-07-04
Get gateway instances running without Auth

DomBlack:
This commit makes the following changes:

- Instances can be defined as gateways and brought up by themselves
- `initService` functions are not being called if the instance isn't hosting that service
- API endpoints are not registered if the instance isn't hosting that service
- Only the gateway logs when it is listening for HTTP requests and the registered endpoints locally - this is to reduce log spam on startup
- Missing secrets as a warning as been moved from the runtime to the daemon (again to reduce duplication)

### DIFF
--- a/runtime/appruntime/apisdk/service/service.go
+++ b/runtime/appruntime/apisdk/service/service.go
@@ -114,11 +114,11 @@ type Manager struct {
 }
 
 func (mgr *Manager) RegisterService(i Initializer) {
-	if !mgr.IsHosting(i) {
+	name := i.ServiceName()
+	if !mgr.IsHosting(name) {
 		return
 	}
 
-	name := i.ServiceName()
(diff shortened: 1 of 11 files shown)


runtime/appruntime/apisdk/service/service.go
```diff
@@ -120,6 +126,21 @@ func (mgr *Manager) RegisterService(i Initializer) {
 	mgr.svcInit = append(mgr.svcInit, i)
 }
 
+func (mgr *Manager) IsHosting(i Initializer) bool {
```
eandre, line 129: Maybe clearer to have this take a `name string` argument? And then in `RegisterService` move the `!mgr.IsHosting(...)` if statement below `name := i.ServiceName()`

eandre (approved):

PR #786, opened 2023-07-05
Remote Auth Handler calls implemented

DomBlack:
This PR adds the ability for an Encore service (or gateway) to make a remote call to another container to run an auth handler.

This is required when the auth handler requires access to infrastructure that only the service it is defined in has access to. So all other containers will route their auth calls to that container.

This PR also fixes an issue where secrets would cause a panic in services when not set, even if those services where not running within the container.

### DIFF
--- a/runtime/appruntime/apisdk/api/auth_remote.go
+++ b/runtime/appruntime/apisdk/api/auth_remote.go
@@ -1,0 +1,163 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/rs/zerolog"
+
+	"encore.dev/appruntime/apisdk/api/transport"
+	"encore.dev/appruntime/exported/config"
+	"encore.dev/appruntime/exported/model"
+	"encore.dev/beta/errs"
+)
+
+// remoteAuthHandler is an AuthHandler that calls a remotely hosted auth handler.
+//
+// This is used when the auth handler is not being hosted in this container, but in
+// another container somewhere else in the cluster.
+type remoteAuthHandler struct {
+	server         *Server        // The server we're running in
+	hostingService config.Service // The service name of the remote auth handler
+	authURL        string         // The URL of the remote auth handler
+	original       AuthHandler
+	logger         zerolog.Logger
+	traceLogs      bool
+}
+
+var _ AuthHandler = (*remoteAuthHandler)(nil)
+
+func (r *remoteAuthHandler) Authenticate(c IncomingContext) (model.AuthInfo, error) {
+	// Quickly check if the auth data is parsable here, if it isn't we can return early
+	// without making a remote call.
+	if err := r.original.ParseAuthData(c); err != nil {
+		return model.AuthInfo{}, err
+	}
+
+	if r.traceLogs {
+		r.logger.Trace().Msg("calling auth handler")
+	}
+
+	// Create the auth request
+	authReq, err := http.NewRequestWithContext(c.ctx, http.MethodPost, r.authURL, nil)
+	if err != nil {
+		r.logger.Err(err).Msg("unable to create auth request")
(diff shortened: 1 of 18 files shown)


runtime/appruntime/apisdk/api/auth_remote.go
```diff
@@ -0,0 +1,163 @@
...
+	// Create the auth request
+	authReq, err := http.NewRequestWithContext(c.ctx, http.MethodPost, r.authURL, nil)
+	if err != nil {
+		r.logger.Err(err).Msg("unable to create auth request")
+		return model.AuthInfo{}, errs.Wrap(err, "unable to create auth request")
+	}
+
+	// Copy over any data which might be needed by the auth handler (query string, headers and cookies)
+	authReq.URL.RawQuery = c.req.URL.RawQuery
+	for k, v := range c.req.Header {
```
eandre, line 50: Probably we could just `maps.Clone` followed by deleting the `X-Encore-Auth` header

runtime/appruntime/apisdk/api/auth_remote.go
```diff
@@ -0,0 +1,163 @@
...
+			continue
+		}
+
+		for _, vv := range v {
+			authReq.Header.Add(k, vv)
+		}
+	}
+
+	// Add call meta data to the request, so the receiving service can use to allow access to the auth handler
+	// This also passes a shared TraceID to the auth handler, so it and the microservice called after the gateway
```
eandre, line 62: Nit: we don't typically specify "microservice" but just "service"
```suggestion
	// This also passes a shared TraceID to the auth handler, so it and the service called after the gateway
```

runtime/appruntime/apisdk/api/encore_routes.go
```diff
@@ -12,6 +12,7 @@ import (
 func (s *Server) registerEncoreRoutes() {
 	s.encore.HandlerFunc(wildcardMethod, "/healthz", s.handleHealthz)
 	s.encore.Handle("POST", "/pubsub/push/:subscription_id", s.handlePubsubPush)
+	s.encore.Handle("POST", "/auth_handler", s.handleRemoteAuthCall)
```
eandre, line 15: Can we change the URL to `/authhandler` instead? Underscores in URLs are a bit unusual I think

v2/codegen/infragen/secretsgen/secretsgen.go
```diff
@@ -27,12 +29,21 @@ func Gen(gen *codegen.Generator, pkg *pkginfo.Package, secrets []*secrets.Secret
...
 		}
 
+		svcName := strconv.Quote(
+			option.
+				Map(
+					svc,
+					func(svc *app.Service) string { return svc.Name },
+				).
+				GetOrElse(""),
+		)
```
eandre, line 39: Can we make this like two lines? If we move the closure to a separate variable the rest should fit in one line

runtime/appruntime/apisdk/api/gateway.go
```diff
@@ -36,18 +38,58 @@ func (s *Server) createGatewayHandlerAdapter(h Handler) httprouter.Handle {
...
+		meta := CallMetaFromContext(req.Context())
+
+		info, proceed := s.runAuthHandler(h, s.NewIncomingContext(w, req, toUnnamedParams(ps), meta))
+		if proceed {
+			meta.Internal = &InternalCallMeta{
+				Caller: GatewayCaller{
+					ServiceName: h.ServiceName(),
+					Endpoint:    h.EndpointName(),
+				},
+				AuthUID:  string(info.UID),
```
eandre, line 55: Are we guaranteed that `info` is non-nil if `proceed` is true? (I forget the semantics of `runAuthHandler`)

DomBlack: `info` isn't a pointer - thus we always garanutee at least a zero value info struct here.

eandre (approved):

PR #804, opened 2023-07-19
Fix possible deadlock in AWS pubsub

DomBlack:


runtime/pubsub/internal/utils/workers.go
```diff
@@ -150,14 +175,17 @@ func workInWorkPool[Work any](ctx context.Context, maxConcurrency int, maxBatchS
 
 	// fetchProcessor is a small wrapper around the fetcher function that passes the fetched work to the workers
 	// it will fetch upto maxConcurrency items at a time in batches of maxBatchSize items
-	var lastFetch time.Time
+	var lastFetch atomic.Pointer[time.Time]
+	var epoch time.Time
+	lastFetch.Store(&epoch)
```
DomBlack, line 180: There was a race on this, so I've moved it into an atomic

runtime/pubsub/internal/aws/topic.go
```diff
@@ -132,33 +148,51 @@ func (t *topic) Subscribe(logger *zerolog.Logger, maxConcurrency int, ackDeadlin
...
 					// Call the callback, and if there was no error, then we can delete the message
-					msgCtx, cancel := context.WithTimeout(t.ctx, ackDeadline)
+					msgCtx, cancel := context.WithTimeout(ctx, ackDeadline)
+					defer cancel()
 					err = f(msgCtx, msgWrapper.MessageId, msgWrapper.Timestamp, int(deliveryAttempt), attributes, []byte(msgWrapper.Message))
 					cancel()
+
+					// Check if the context has been cancelled, and if so, return the error
+					if ctx.Err() != nil {
+						return ctx.Err()
```
eandre, line 158: Why? This might hide other errors from the func if they error for some other reason

runtime/pubsub/internal/utils/workers.go
```diff
@@ -179,7 +207,12 @@ func workInWorkPool[Work any](ctx context.Context, maxConcurrency int, maxBatchS
 
 			// Pass work to workers
 			for _, w := range work {
-				workChan <- w
+				select {
+				case workChan <- w:
+					// success, we passed the work to a worker
+				case <-fetchCtx.Done():
+					return
```
DomBlack, line 214: I think this was the deadlock;

If all the workers panic'ed and quit, that could have resulted in the fetch processor trying to write onto the `workChan` but nothing pulling and reading them.

I've solved this by adding this new select, so if the ctx is done, we don't even try to push to the channel & I've also added additional panic recovery wrappers at other points in the code.

runtime/pubsub/internal/aws/topic.go
```diff
@@ -132,33 +148,51 @@ func (t *topic) Subscribe(logger *zerolog.Logger, maxConcurrency int, ackDeadlin
...
+					ctx, responseCancel := context.WithTimeout(ctx, 30*time.Second)
+					defer responseCancel()
+
 					if err != nil {
+						logger.Err(err).Str("msg_id", msgWrapper.MessageId).Msg("unable to process message")
+
 						// If there was an error processing the message, apply the backoff policy
 						_, delay := utils.GetDelay(retryPolicy.MaxRetries, retryPolicy.MinBackoff, retryPolicy.MaxBackoff, uint16(deliveryAttempt))
-						_, visibilityChangeErr := t.sqsClient.ChangeMessageVisibility(t.ctx, &sqs.ChangeMessageVisibilityInput{
+						_, visibilityChangeErr := t.sqsClient.ChangeMessageVisibility(ctx, &sqs.ChangeMessageVisibilityInput{
```
eandre, line 173: Shouldn't we do this with a context not derived from the input, so we do this even if the input context is canceled?

runtime/pubsub/internal/aws/topic.go
```diff
@@ -87,12 +87,28 @@ func (t *topic) Subscribe(logger *zerolog.Logger, maxConcurrency int, ackDeadlin
...
 				maxConcurrency, 10,
 				func(ctx context.Context, maxToFetch int) ([]sqsTypes.Message, error) {
-					resp, err := t.sqsClient.ReceiveMessage(t.ctx, &sqs.ReceiveMessageInput{
+					// We should only long poll for 20 seconds, so if this takes more than
+					// 30 seconds we should cancel the context and try again
+					//
+					// We do this incase the ReceiveMessage call gets stuck on the server
+					// and doesn't return
+					ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+					defer cancel()
```
DomBlack, line 109: One theory I have is the AWS library _might_ be stalling and blocking under high load, so I've introduced a smaller timeout to try and cause a context cancelled error

runtime/pubsub/internal/utils/workers.go
```diff
@@ -42,19 +44,37 @@ type WorkProcessor[Work any] func(ctx context.Context, work Work) error
 // this again you could end up with 2x maxConcurrency workers running at the same time. (1x from the original run who
 // are still processing work and 1x from the new run).
 func WorkConcurrently[Work any](ctx context.Context, maxConcurrency int, maxBatchSize int, fetch WorkFetcher[Work], worker WorkProcessor[Work]) error {
+	fetchWithPanicHandling := func(ctx context.Context, maxToFetch int) (work []Work, err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				err = errs.B().Msgf("panic: %v", r).Err()
```
eandre, line 50: Can we include the stack like we do elsewhere?

runtime/pubsub/internal/aws/topic.go
DomBlack: No I delribatley wanted both these API calls to release when the fetch context is cancelled, as we immediately go into a loop to go again.

I think them being based on the `t.ctx`, rather than the fetch context could have been an issue; as we never cancel the `t.ctx`, but the `fetchCtx` is cancelled when we want to exit the `WorkConcurrently` code

PR #815, opened 2023-07-26
cli: bump 'encore gen wrappers timeout'

eandre:
We've gotten reports that the command times out.
It's not clear whether it's due to taking a long time to parse
larger projects with a cold cache, or due to a deadlock.

I haven't been able to reproduce the deadlock, so let's start
by bumping the timeout and seeing if that helps.

### DIFF
--- a/cli/cmd/encore/gen.go
+++ b/cli/cmd/encore/gen.go
@@ -117,7 +117,7 @@ which may require the user-facing wrapper code to be manually generated.`,
 		Args: cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			appRoot, _ := determineAppRoot()
-			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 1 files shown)


cli/cmd/encore/gen.go
```diff
@@ -117,7 +117,7 @@ which may require the user-facing wrapper code to be manually generated.`,
 		Args: cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			appRoot, _ := determineAppRoot()
-			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
DomBlack, line 120: I'd maybe go to 30 seconds. On shared CI machines, their can be really little CPU availability and I've seen before it can be bursty too, which means stuff which can complete sub-second on my machine can sometimes take 15 seconds on CI.

DomBlack (approved):

PR #818, opened 2023-07-28
cli: fix missing secret warning

eandre:
The shortcut when all services are running in the same
process was missing the check for missing secrets.

Thanks Neal Lathia for reporting.


DomBlack (approved):

nlathia (commented):
🥳

PR #819, opened 2023-07-30
pubsub: fix 'projectID string is empty' panic

eandre:
Newer versions of the `cloud.google.com/go/pubsub` package
no longer permit creating a client without specifying a project id.

Since we don't use the default project id but always use `TopicInProject`
and `SubscriptionInProject`, work around this by specifying a placeholder
project id.

### DIFF
--- a/runtime/pubsub/internal/gcp/clients.go
+++ b/runtime/pubsub/internal/gcp/clients.go
@@ -10,7 +10,7 @@ import (
 func (mgr *Manager) getClient() *pubsub.Client {
 	mgr.clientOnce.Do(func() {
 		// Create a new client
-		cl, err := pubsub.NewClient(mgr.ctxs.Connection, "")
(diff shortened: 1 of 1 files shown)


runtime/pubsub/internal/gcp/clients.go
```diff
@@ -10,7 +10,7 @@ import (
 func (mgr *Manager) getClient() *pubsub.Client {
 	mgr.clientOnce.Do(func() {
 		// Create a new client
-		cl, err := pubsub.NewClient(mgr.ctxs.Connection, "")
+		cl, err := pubsub.NewClient(mgr.ctxs.Connection, "-")
```
DomBlack, line 13: I would have updated our runtime version as well here, rather than just do this as the newer versions have security fixes we should include

eandre: Agreed, but seemed reasonable to do that as a separate change

PR #820, opened 2023-07-31
feat: add infrastructure namespaces

eandre:
This commit adds the concept of infrastructure namespaces for local
development. This makes all Encore-managed infrastructure resources
isolated to their namespace, enabling easy task switching without
having to tear down and reset stateful infrastructure like databases.

Every application starts with a single `default` namespace that is
created on-demand if there are no namespaces. Additional namespaces
can be created and managed using the `encore namespace` family of
CLI commands.


cli/cmd/encore/daemon/migrations/1_initial_schema.up.sql
```diff
@@ -1,10 +1,3 @@
-CREATE TABLE IF NOT EXISTS schema_migrations (
-    dummy BOOLEAN PRIMARY KEY,
-    version INT NOT NULL
-);
-
-INSERT OR REPLACE INTO schema_migrations(dummy, version) VALUES (true, 1);
-
```
DomBlack, line 7: Should we issue a drop command for this older db?

cli/cmd/encore/namespace/namespace.go
```diff
@@ -0,0 +1,163 @@
...
+var nsCmd = &cobra.Command{
+	Use:     "namespace",
+	Short:   "Manage infrastructure namespaces",
+	Aliases: []string{"ns"},
+}
+
+var listCmd = &cobra.Command{
+	Use:     "list",
+	Short:   "List infrastructure namespaces",
+	Aliases: []string{"ls"},
```
DomBlack, line 26: Maybe add 
```suggestion
	Aliases: []string{"ls"},
	Args: cobra.NoArgs()
```

cli/cmd/encore/namespace/namespace.go
```diff
@@ -0,0 +1,163 @@
...
+				AppRoot: appRoot,
+				Name:    args[0],
+				Create:  create,
+			})
+			if err != nil {
+				cmdutil.Fatal(err)
+			}
+			fmt.Fprintf(os.Stdout, "switched to namespace %s\n", ns.Name)
+		},
+	}
```
DomBlack, line 133: This commman definition feels out of place with the others being pkg level variables, we could still add the flag in an initFunc

cli/daemon/namespace.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	ns, err := s.ns.Switch(ctx, app, namespace.Name(req.Name))
+	if err != nil {
+		return nil, err
+	}
+	return ns.ToProto(), nil
+}
+
+func (s *Server) namespaceOrActive(ctx context.Context, app *apps.Instance, ns *string) (*namespace.Namespace, error) {
```
DomBlack, line 70: What's the behaviour if `ns` is `nil` and there's no active namespace?

cli/daemon/namespace/namespace.go
```diff
@@ -0,0 +1,327 @@
...
+	err = tx.Commit()
+	return errors.Wrap(err, "delete namespace")
+}
+
+func (m *Manager) Switch(ctx context.Context, app *apps.Instance, name Name) (*Namespace, error) {
+	// Resolve the namespace to switch to.
+	var target *Namespace
+
+	// If the name is "-", switch to the previous namespace.
+	if name == "-" {
```
DomBlack, line 214: I didn't see this documented on the commands or else where in the code; just deep in the manager here.

We should add this to the command docs

cli/cmd/encore/namespace/namespace.go
```diff
@@ -0,0 +1,163 @@
...
+				marker = "*"
+			}
+			fmt.Fprintf(os.Stdout, "%s %s\n", marker, ns.Name)
+		}
+	},
+}
+
+var createCmd = &cobra.Command{
+	Use:   "create NAME",
+	Short: "Create a new infrastructure namespace",
```
DomBlack, line 60: Alias `add`?

cli/cmd/encore/daemon/migrations/1_initial_schema.up.sql
eandre: We do issue a `DROP TABLE schema_migrations` but it's being done in code (since it needs to run before we run the migrations)

PR #834, opened 2023-08-06
Add `encore rand` command family

eandre:
This adds a small family of utility commands for generating various kinds of random data. It supports generating UUIDs, bytes, and words.


DomBlack (commented):
I’m not sure this belongs as a cli command in Encore. 

What’s the usecases your thinking about where this is something only Encore can provide?

eandre:
It's not something only Encore can provide, but it's very useful.

DomBlack (dismissed):

PR #853, opened 2023-08-12
Use persistent work dirs

eandre:
It turns out the `go` tool's build cache derives its cache keys
from the physical location of files when using overlays.

As a result, since Encore created temporary work dirs for those files, any packages that Encore generated code for were not cached.

Fix this by using a persistent work directory, keyed by the app id.

The end result is live reloads are dramatically faster, and if nothing
changes between builds the application now starts up instantaneously.


DomBlack (dismissed):
Nice catch

DomBlack (approved):

PR #861, opened 2023-08-16
cli: add `encore app init`

eandre:
This adds a new command, `encore app init`, that initializes
an existing repository by creating a new app on the server
and linking with it.

Stacked on top of #860

### DIFF
--- a/cli/cmd/encore/app/initialize.go
+++ b/cli/cmd/encore/app/initialize.go
@@ -1,0 +1,108 @@
+package app
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"time"
+
+	"github.com/briandowns/spinner"
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+
+	"encr.dev/cli/cmd/encore/auth"
+	"encr.dev/cli/cmd/encore/cmdutil"
+	"encr.dev/internal/conf"
+)
+
+// Create a new app from scratch: `encore app create`
+// Link an existing app to an existing repo: `encore app link <app-id>`
+// Link an existing repo to a new app: `encore app init <name>`
+
+func init() {
+	initAppCmd := &cobra.Command{
+		Use:   "init [name]",
+		Short: "Create a new Encore app from an existing repository",
+		Args:  cobra.MaximumNArgs(1),
+
+		DisableFlagsInUseLine: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			var name string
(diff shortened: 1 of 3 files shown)


cli/cmd/encore/app/initialize.go
```diff
@@ -0,0 +1,108 @@
...
+	"encr.dev/internal/conf"
+)
+
+// Create a new app from scratch: `encore app create`
+// Link an existing app to an existing repo: `encore app link <app-id>`
+// Link an existing repo to a new app: `encore app init <name>`
+
+func init() {
+	initAppCmd := &cobra.Command{
+		Use:   "init [name]",
```
DomBlack, line 25: Not confusing at all to have `init` add an `init` command lol

cli/cmd/encore/app/initialize.go
```diff
@@ -0,0 +1,108 @@
...
+	// Update to latest encore.dev release if this looks to be a go module.
+	if _, err := os.Stat("go.mod"); err == nil {
+		s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
+		s.Prefix = "Running go get encore.dev@latest"
+		s.Start()
+		if err := gogetEncore(name); err != nil {
+			s.FinalMSG = fmt.Sprintf("failed, skipping: %v", err.Error())
+		}
+		s.Stop()
+	}
```
DomBlack, line 105: I actually wonder if we should do this in the daemon regulllary?

i.e. have the daemon check the latest `encore.dev` version, and the update any `go.mod` files to the latest if we update it.

cli/cmd/encore/cmdutil/cmdutil.go
```diff
@@ -16,41 +18,54 @@ import (
...
+// MaybeAppRoot determines the app root by looking for the "encore.app" file,
 // initially in the current directory and then recursively in parent directories
 // up to the filesystem root.
 //
 // It reports the absolute path to the app root, and the
 // relative path from the app root to the working directory.
-//
-// On errors it prints an error message and exits.
-func AppRoot() (appRoot, relPath string) {
+func MaybeAppRoot() (appRoot, relPath string, err error) {
```
DomBlack, line 32: I know I approved this in the other PR, but I wonder if we should look for a `.git` directory in this directory or the directory above?

Then when we create the app on the server, we can automatically set the "folder" within the repo to match the folder we've found, we could also automatically add `encore` as a push target.

Finally if we detect the existing app origin is github, we could display a link to the "Link your Github repo to get free PR environments"

PR #864, opened 2023-08-17
ENC-1666 Add EnvID to all metrics

ekerfelt:


runtime/appruntime/infrasdk/metadata/metadata.go
```diff
@@ -22,12 +22,43 @@ type ContainerMetadata struct {
 	ServiceID  string
 	RevisionID string
 	InstanceID string
+	EnvID      string
```
eandre, line 25: Should we call this EnvName?

eandre (approved):

PR #882, opened 2023-09-07
Expose app root path to front-end

simon-johansson:
So that we can open the files in the users editor


pkg/editors/lookup.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// Resolve a list of installed editors on the user's machine, using the known
+// install identifiers that each OS supports.
+func Resolve(ctx context.Context) ([]FoundEditor, error) {
+	editorCacheMu.Lock()
+	defer editorCacheMu.Unlock()
+
+	// If we don't have a cache yet, populate it
+	if len(editorCache) == 0 && cap(editorCache) == 0 {
```
eandre, line 40: This would be nicer with https://pkg.go.dev/go4.org/syncutil#Once

pkg/editors/lookup_darwin.go
```diff
@@ -0,0 +1,251 @@
...
+			"com.slickedit.SlickEditPro2016",
+			"com.slickedit.SlickEditPro2015",
+		},
+	},
+	{
+		Name:              "IntelliJ",
+		BundleIdentifiers: []string{"com.jetbrains.intellij"},
+	},
+	{
+		Name:              "IntelliJ Community Edition",
```
eandre, line 133: Maybe nicer to shorten this (and others) to `Intellij CE`?

pkg/editors/lookup.go
DomBlack: Hmm, this is already being performed under a lock, `syncutil.Once` would just introduce a new lock and new variable. Is this a readability thing for you?

pkg/editors/lookup_darwin.go
eandre: If you prefer? I don't think the user experience should suffer because we want to align the code base, so if we both agree the shorter name is better we should do something about it :)

pkg/editors/launch_nonwindows.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"syscall"
+)
+
+// detachSysProcAttr returns attributes which ensure the new process is detached from
+// the current process
+func detachSysProcAttr() *syscall.SysProcAttr {
+	return &syscall.SysProcAttr{
+		Setsid: true,
```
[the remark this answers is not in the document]
DomBlack: My reading of `Setpgid` is the opposite of what we want;

> Setpgid sets the process group ID of the child to Pgid, or, if Pgid == 0, to the new child's process ID.

Where as `Setsid` says;

> Create session.

cli/daemon/dash/dash.go
```diff
@@ -171,6 +188,61 @@ func (h *handler) Handle(ctx context.Context, reply jsonrpc2.Replier, r jsonrpc2
...
+			return reply(ctx, nil, err)
+		}
+
+		editor, err := editors.Find(ctx, params.Editor)
+		if err != nil {
+			log.Err(err).Str("editor", params.Editor).Msg("dash: could not find editor")
+			return reply(ctx, nil, err)
+		}
+
+		runs := h.run.ListRuns()
```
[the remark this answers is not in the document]
DomBlack: To be honest I just copied the code used higher up in the dash API for getting an app. We could move to  using the app manager instead, but given the UI only works when the app is open do you think we need to?

pkg/editors/launch_nonwindows.go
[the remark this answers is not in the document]
eandre: Setpgid means create a new process group for the child, instead of inheriting the parent's process group

pkg/editors/lookup.go
eandre: I think with `syncutil.Once` we wouldn't need a separate lock, but up to you. Yes, just for readability

cli/daemon/dash/dash.go
```diff
@@ -152,25 +192,61 @@ func (h *handler) Handle(ctx context.Context, reply jsonrpc2.Replier, r jsonrpc2
...
 		if err != nil {
+			if errors.Is(err, apps.ErrNotFound) {
+				return reply(ctx, nil, fmt.Errorf("app not found, try running encore run"))
+			}
+			log.Err(err).Str("app_id", params.AppID).Msg("dash: could not find app")
 			return reply(ctx, nil, err)
 		}
-		return reply(ctx, sourceContextResponse{Lines: lines, Start: start}, nil)
+
+		params.File = filepath.Join(app.Root(), params.File)
```
eandre, line 241: We should check that the input file is local (filepath.IsLocal) before this to ensure only files within the app are opened.

eandre (approved):

PR #920, opened 2023-10-20
Update react-query syntax to v5

simon-johansson:


melkstam:
Looks better 👏 Another heads-up is that the package name now is `@tanstack/react-query` 👀

marcuskohlberg (approved):

PR #938, opened 2023-11-13
Support graceful pod termination in Kubernetes

DomBlack:
This commit modifies the graceful shutdown system of the Encore runtime such that if we're running inside a Kubernetes pod, we will wait for a period of time after receiving a SIGTERM before we start the shutdown of the running application.

This allows time for Kubernetes to put the pod into a terminating state while external systems such as load balancers get updated that the pod is not in a health state and should not be sent new traffic.

Effectively, this means that the pod will be in a terminating state for the duration of the `graceTerminationPeriodSeconds` value in the pod spec (minus the gracefult shutdown time configured by Encore).

### DIFF
--- a/runtimes/go/appruntime/shared/shutdown/shutdown.go
+++ b/runtimes/go/appruntime/shared/shutdown/shutdown.go
@@ -244,7 +244,7 @@ func (t *Tracker) Shutdown(reasonSignal os.Signal, reasonError error) {
 
 		// If we received a SIGTERM and have a configured keepAcceptingFor duration
 		// then log the fact we're going to continue accepting new requests and then
-		// sleep for that time before being to graceful shutdown.
+		// sleep for that time before begining to graceful shutdown.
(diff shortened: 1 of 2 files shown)


runtimes/go/appruntime/shared/shutdown/shutdown.go
```diff
@@ -188,6 +242,15 @@ func (t *Tracker) Shutdown(reasonSignal os.Signal, reasonError error) {
 			t.logger.Trace().Msg("initiating graceful shutdown")
 		}
 
+		// If we received a SIGTERM and have a configured keepAcceptingFor duration
+		// then log the fact we're going to continue accepting new requests and then
+		// sleep for that time before being to graceful shutdown.
```
ekerfelt, line 247:
```suggestion
		// sleep for that time before begining to graceful shutdown.
```

ekerfelt (approved):

PR #940, opened 2023-11-14
chore - typo update in the shell output

burhan:
The output for the shell doesn't have the prompt. Also updated the minor version to the latest one.


eandre (approved):
Nice catch, thanks!

PR #947, opened 2023-11-20
Add rest api TS tutorial

simon-johansson:


docs/tutorials/rest-api.ts.md
```diff
@@ -0,0 +1,216 @@
...
+
+```ts
+export const shorten = api(
+  { method: "POST", path: "/url" },
+  async ({ url }: ShortenParams): Promise<URL> => {
+    const id = generateID();
+    for await (const row of DB.q`
+        INSERT INTO url (id, original_url)
+        VALUES (${id}, ${url})
+    `) {
```
eandre, line 131: Instead of the loop we can now use `DB.exec`:
```
await DB.exec`INSERT INTO url (id, original_url) VALUES (${id}, ${url})`
```

marcuskohlberg (approved):

PR #964, opened 2023-11-29
Add defaultExpanded field to accordion

simon-johansson:


docs/menu.cue
```diff
@@ -89,6 +89,8 @@
 #AccordionMenuItem: {
 	kind: "accordion"
 	text: string
+	// If the accordion is open by default.
+	defaultExpanded?: bool
```
eandre, line 93: I'd make this a bool with a default value of false, instead of an optional bool:

```suggestion
	defaultExpanded: bool | *false
```

marcuskohlberg (approved):

PR #970, opened 2023-11-30
Add TS content to rest-api tutorial

simon-johansson:
I have skipped the testing section for the typescript content as we will not have that when we go live with the public beta.


eandre:
Merged in #1105

PR #978, opened 2023-12-07
metrics: fix initialization race

### LINKED ISSUE #977: metrics: panic handling request: runtime error: index out of range [1] with length 0
Hey folks 👋🏻 

We've seen this panic a couple of times during tests, and now once in production:

```
panic handling request: runtime error: index out of range [1] with length 0
```

It's coming from this line: https://github.com/encoredev/encore/blob/caaa3b92103af17dc7b2fabfe971f144ae56a10b/runtimes/go/metrics/metrics.go#L38

Here's our code that triggers it:

```Go
eventCounter.With(labels{
	Action:  action,
	Name:    name,
	Success: success,
}).Increment()
```

If it's relevant info, this service is second in the alphabetically sorted list of services 🔡

eandre:
When using a counter group or gauge group, the initialization of the
time series happens on first use. However, if two different goroutines
tried to access the same set of labels concurrently, there was a rare race
where one goroutine could try to increment the time series state before
the other goroutine had finished initializing the data structure.

Fix this by waiting for the initialization to complete before returning
the counter on a call to .With(). This wait is cheap and involves only
a single atomic load in the fast path after initialization has completed.

Thanks Dan Upton and Neal Lathia for the report.

Fixes #977

### DIFF
--- a/runtimes/go/metrics/bits_internal.go
+++ b/runtimes/go/metrics/bits_internal.go
@@ -85,10 +85,11 @@ type initGate struct {
 }
 
 func (g *initGate) Start() {
+	g.mu.Lock()
 	if !atomic.CompareAndSwapUint32(&g.state, 0, 1) {
+		g.mu.Unlock() // don't leave the mutex in a locked state if we panic
 		panic("initGate: already started")
 	}
-	g.mu.Lock()
 }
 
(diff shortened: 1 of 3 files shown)


nlathia (commented):
Thank you!

PR #988, opened 2023-12-19
Improve 'encore app create' behavior

eandre:
- Hide TypeScript examples on non-beta builds
- Error if app directory already exists
- Generate wrappers on creation

### DIFF
--- a/cli/cmd/encore/app/create.go
+++ b/cli/cmd/encore/app/create.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -14,6 +13,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	"github.com/cockroachdb/errors"
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"github.com/tailscale/hujson"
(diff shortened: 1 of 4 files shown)


cli/cmd/encore/app/create_form.go
```diff
@@ -289,51 +305,70 @@ func selectTemplate(inputName, inputTemplate string) (appName, template string)
...
+			},
+		}
+	})()
+
+	// If we're not running the Encore beta (with TypeScript support),
+	// filter out any TypeScript items.
+	haveTypeScript := version.Channel != version.GA
+	if !haveTypeScript {
+		templates = fns.Filter(templates, func(t templateItem) bool {
+			return !strings.Contains(strings.ToLower(t.ItemTitle), "typescript")
```
ekerfelt, line 367: Guess this is a temporary thing. Otherwise it might be useful to introduce a language field to the template items

cli/cmd/encore/app/create.go
```diff
@@ -47,6 +51,7 @@ var createAppCmd = &cobra.Command{
 
 func init() {
 	appCmd.AddCommand(createAppCmd)
+	createAppCmd.Flags().BoolVar(&createAppOnPlatform, "platform", true, "whether to create the app with the Encore Platform")
```
ekerfelt, line 54: What does createAppOnPlatform mean? What's the use case of setting this to false?

ekerfelt (approved):

cli/cmd/encore/app/create.go
eandre: It means the app is only created locally, mainly for local testing purposes

cli/cmd/encore/app/create_form.go
eandre: Yeah, just temporary

ekerfelt (approved):

PR #997, opened 2024-01-09
Support async functions in for auth generator in TS and JS generated clients

### LINKED ISSUE #995: Support async authDataGenerator in generated typescript client
Currently the typescript client generated using the client gen has the type `auth?: string | AuthDataGenerator`
where `AuthDataGenerator` is of type `() => (string | undefined)`

This is great when the auth token is known / a function that return a known auth token. However in situations where auth token itself is returned via an API call / as a promise, the auth data generator doesn't work.

This happens with firebase auth, the auth token of the user is accessed using [this](https://firebase.google.com/docs/reference/js/v8/firebase.User#getidtoken) and `Promise` is returned because of an underlying API call.

The client side api call is already async.
Suggested change in generated code for the `AuthDataGenerator` type:
```
export type AuthDataGenerator = () => Promise<string | undefined> | string | undefined
```

Happy to update the typescript client generator if this is a good idea.

anish749:
fixes #995

### DIFF
--- a/e2e-tests/testdata/echo_client/ts/client.ts
+++ b/e2e-tests/testdata/echo_client/ts/client.ts
@@ -780,7 +780,10 @@ type CallParameters = Omit<RequestInit, "method" | "body" | "headers"> & {
 }
 
 // AuthDataGenerator is a function that returns a new instance of the authentication data required by this API
-export type AuthDataGenerator = () => (echo.AuthParams | undefined)
+export type AuthDataGenerator = () =>
+  | echo.AuthParams
+  | undefined
+  | Promise<echo.AuthParams | undefined>;
 
 // A fetcher is the prototype for the inbuilt Fetch function
 export type Fetcher = typeof fetch;
@@ -815,7 +818,7 @@ class BaseClient {
             if (typeof auth === "function") {
                 this.authGenerator = auth
             } else {
-                this.authGenerator = () => auth                
+                this.authGenerator = () => Promise.resolve(auth)
             }
         }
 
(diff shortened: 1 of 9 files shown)


e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -780,7 +780,10 @@ type CallParameters = Omit<RequestInit, "method" | "body" | "headers"> & {
 }
 
 // AuthDataGenerator is a function that returns a new instance of the authentication data required by this API
-export type AuthDataGenerator = () => (echo.AuthParams | undefined)
+export type AuthDataGenerator = () =>
+  | echo.AuthParams
+  | undefined
+  | Promise<echo.AuthParams | undefined>;
```
eandre, line 786: Can we move the `| undefined` case to the end? It's a bit unintuitive otherwise.

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -815,7 +818,7 @@ class BaseClient {
             if (typeof auth === "function") {
                 this.authGenerator = auth
```
eandre, line 819: What if the function doesn't return a promise; wouldn't the `await this.authGenerator()` fail then?

anish749: yeah, i was thinking the same and then tested, and `await` on a value that is not a promise, returns the value itself in JS.

do you know if there are integrations tests / some other tests for the generated code somewhere in the repo where this behaviour gets tested?

anish749: I had another look at this one, and there might be some performance implications with the added `await`, i'll add a check around this. thanks!

eandre (approved):
Thank you so much!

PR #1012, opened 2024-01-23
[ENC-2013] Add tracing for tests

DomBlack:
**This PR must be merged at the same time as encoredev/go#27**

This commit adds tracing for tests. When tests are run, traces will be captured and stored like traces for API requests, which are exposed to the user via the Trace's view on the local dashboard.

## Screenshots of UI

image

image

image


cli/daemon/engine/trace2/sqlite/write.go
```diff
@@ -203,6 +222,26 @@ func (s *Store) updateSpanEndIndex(ctx context.Context, meta *trace2.Meta, ev *t
 		return nil
 	}
 
+	if msg := end.GetTest(); msg != nil {
+		// TODO: record skipped test spans
```
eandre, line 226: What does this mean?

pkg/traceparser/parser.go
```diff
@@ -268,9 +272,16 @@ func (tp *traceParser) requestSpanStart() *tracepb2.SpanStart {
 				RequestPayload:   tp.ByteString(),
 				ExtCorrelationId: ptrOrNil(tp.String()),
 				Uid:              ptrOrNil(tp.String()),
+				Mocked:           false, // introduced in version 101
```
eandre, line 275: 101?

eandre, line 275: Nicer to do it inline: `(func() bool { ... })()`; that way we don't need the type assertion below

runtimes/go/appruntime/shared/reqtrack/trace_stream.go
```diff
@@ -46,9 +47,17 @@ func (l *lazyTraceInit) Logger() trace2.Logger {
 	return l.log
 }
 
+func (l *lazyTraceInit) WaitForStreamSent() {
+	l.sent.Wait()
```
eandre, line 51: This will return immediately if `l.initStream` hasn't been called. Is that intended?

DomBlack: initStream is always called before WaitForStream; the only call site is:

```go
		op.trace.MarkDone()

		if blockOnTraceSend {
			op.trace.WaitForStreamSent()
		}
```

Where `MarkDone` is:

```go
func (l *lazyTraceInit) MarkDone() {
	l.log.MarkDone()
	l.initStream()
}
```

eandre (approved):

PR #1014, opened 2024-01-25
Quality of Life improvements to local development dashboard

DomBlack:
This PR adds the capability to view all Encore apps via the local dashboard, regardless of if they are currently running or not, as well as several other UI improvements to the dashboard;

- request logs are now displayed when an API call is made
- You get notified when the app is compiling
- You get notified when the app has just restarted (due to a hot-reload)
- You get notified of any compile errors and what they are

### Offline apps are greyed out on lists, but can still be navigated
Screenshot 2024-01-25 at 14 05 04

image

### The API explorer now shows logs which were emitted during the API call
image

### The dashboard now tells you when the app is compiling or just freshly restarted
Screenshot 2024-01-25 at 14 14 22

image

### Compile errors are now shown on the dashboard
image

### DIFF
--- a/cli/daemon/dash/dash.go
+++ b/cli/daemon/dash/dash.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/golang/protobuf/jsonpb"
@@ -27,7 +28,7 @@ import (
 	"encr.dev/pkg/editors"
 	"encr.dev/pkg/errlist"
 	tracepb2 "encr.dev/proto/encore/engine/trace2"
-	v1 "encr.dev/proto/encore/parser/meta/v1"
+	meta "encr.dev/proto/encore/parser/meta/v1"
 )
 
 type handler struct {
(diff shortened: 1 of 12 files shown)


cli/daemon/dash/dash.go
```diff
@@ -66,21 +67,42 @@ func (h *handler) Handle(ctx context.Context, reply jsonrpc2.Replier, r jsonrpc2
...
 		apps := []app{} // prevent marshalling as null
-		seen := make(map[string]bool)
-		for _, r := range runs {
-			id := r.App.PlatformOrLocalID()
-			name := r.App.PlatformID()
-			if name == "" {
-				name = filepath.Base(r.App.Root())
+
+		// Load all the apps we know about
+		allApp, err := h.apps.List()
```
eandre, line 76: Do we ever delete from this list? Like if the user deletes the directory? We probably should

DomBlack: Yes we do already do that:

```go
// List lists all known apps.
func (mgr *Manager) List() ([]*Instance, error) {
	roots, err := mgr.listRoots()
	if err != nil {
		return nil, err
	}

	var apps []*Instance
	for _, root := range roots {
		app, err := mgr.resolve(root)
		if errors.Is(err, fs.ErrNotExist) {
			log.Debug().Str("root", root).Msg("app no longer exists, skipping")
			// Delete the
			_, _ = mgr.db.Exec(`DELETE FROM app WHERE root = ?`, root)
			continue
		} else if err != nil {
			log.Error().Err(err).Str("root", root).Msg("unable to resolve app")
			continue
		}
		apps = append(apps, app)
	}

	return apps, nil
}
```

eandre (dismissed):

eandre (approved):

PR #1017, opened 2024-01-27
Added secret check cli command

### LINKED ISSUE #605: Proposal: CLI command checking missing secrets against environments
General idea is to use this new command in GitHub Workflows, which would allow us to check PR before merge if all secrets for `development` and `production` environments are properly set in Encore. 

Right now it is possible to get very late feedback loop if for example you set the secrets only on `development` environment and after some time you deploy to `production`, where it fails on deploy step.

With possible new command, which would list missing secrets & error if any is missing:
`encore secret check [envs...]`

Example usage:
```bash
encore secret check dev
Secret Key Production   Development 
MySecret   ✗                   ✓

Error: There is 1 secret missing.
```

aneshas:
Added `secret check` cli command

Closes #605


aneshas:
@eandre would we be able to move forward with this?

PR #1021, opened 2024-02-01
Update sqlc booking-system.md

rewop:
This commit updates the doc to use `go install` instead of `go get` to install `sqlc` in the booking tutorial.

When going through the tutorial, using `go get` would not install the binary of `sqlc`.


eandre (approved):
Thanks!

PR #1022, opened 2024-02-02
[ENC-2031] Add new 'encore:trace' annotation + Other small changes

DomBlack:
This PR contains several changes;

# New `//encore:trace` Annotation

Any function can now be annotated with `//encore:trace` and will automatically be picked up as a span in Encore's tracing system, even if it is called _outside_ of the context of a API request, PubSub consumer or test. (i.e. this means your background processes, such as a temporal worker can now be annotated! 🎉 )

The annotation supports two options parameters:
- `name`; by default this is the package name and function name, but if overridden becomes what ever you set this to. This can be a string with no whitespace (like `ping_google`) or a quoted string with whitespace (like `"Wait for message to match"`).
- `type`: by default this is `internal`, however the following additional types are supported; `request_handler`, `call`, `producer` & `consumer`. The only effect of using these types is a change of icon within the trace UI.

All request parameters and response data are automatically captured into the trace and the last `error` return type is used to indicate if the span was successful or not.

### Example Uses

```go
//encore:trace name=ping_google type=call
func makeExternalCall(ctx context.Context) error {
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://google.com", nil)
	// ... rest of check ...
	return nil
}

//encore:trace name="Wait for message to match" type=consumer
func waitForMessageToMatch(ctx context.Context, msgID string, expected string) error {
	// ... rest of function ...
}

//encore:trace name="check users age"
func somethingRandom(ctx context.Context, name string, age int) (isOldEnough bool, err error) {
	return false, errors.New("defiently not old enough")
}

//encore:trace name="system startup"
func systemStartup() {
	// no-op
}
```

### Screenshots

**A call made to `systemStartup` from a background thread is seen as a new trace**
image

**`go-routine-4` and `wait for message to match` are both marked as traced, and you can see API calls made within them as separate spans**
image

**All parameters and response variables are captured, including the error state**
image

# The Dash API now caches the last known compile error

This change was requested by @simon-johansson so that when you navigate pages in the dashboard and change between running apps, the UI can show if the app currently has a compile error or not. Previously the behaviour was the UI would show the app running fine as nothing cached the error state or not.

The error is now also aways syntax highlighted (there was a bug where the daemon wasn't running in a process which supported colour, so it defaulted to no colour, even though we know web browsers support color).

# Addition of a new JSON schema for the `encore.app` file

The schema is already live at https://encore.dev/schemas/appfile.schema.json - however now when new apps are created we'll automatically add the `$schema` tag as the first tag, which means editors like IntelliJ, GoLand and VSCode will automatically provide autocompletitions and documentation for the `encore.app` file.

# Model Design for a new Go Trace system

There's also an unused file being committed which includes a model design for a new tracing system which would be compatible with OpenTelemtry that @eandre will pickup and full flush out in future PR's


cxkoda:
+1 on this. this would be really helpful!

PR #1029, opened 2024-02-13
Fix User-Agent for generated client in JS and TS

### LINKED ISSUE #1020: Javascript generated client triggers errors on User-Agent in browser
### Description of the Problem

While learning encore.dev using the [uptime tutorial](https://encore.dev/docs/tutorials/uptime) I found out the frontend client generates the following errors in the browser console

```
Refused to set unsafe header "User-Agent"
```

The error is thrown because the generated js client [attempts to set the User-Agent](https://github.com/encoredev/encore/blob/main/internal/clientgen/javascript.go#L480) and the browser doesn't allow it. 

### Expected Behavior
Generated JS client for browser should not attempt to set User-Agent. User-Agent should be only set when the client is used in backend services. 

### Environment
- Mac Sonoma 14.2.1
- Chrome 120.0.6099.234

rewop:
It adds the User-Agent header to the requests only if the global `window` is undefined, which means the client is running on node.  

Fixes #1020.

### DIFF
--- a/internal/clientgen/javascript.go
+++ b/internal/clientgen/javascript.go
@@ -477,8 +477,14 @@ class BaseClient {`)
         this.baseURL = baseURL
         this.headers = {
             "Content-Type": "application/json",
-            "User-Agent":   "` + userAgent + `",
         }
+
+        // Add User-Agent header if the script is running in the server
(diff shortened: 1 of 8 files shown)


eandre (approved):
Thank you!

PR #1036, opened 2024-02-23
Add ability to disable color in test log output

Willyham:
This adds the ability to pass a `--no-color` flag to `encore test` to disable color output in zerolog. This is useful when displaying output in windows which don't support colors. This is the case for VSCode test output:

The current output of those runs look like this, which makes debugging failed tests extremely difficult:

image

WARNING: This is not tested, as I couldn't quickly figure out how to do that. Building the binary and attempting to use it caused a daemon version mismatch error though both cli and daemon report the same version (I assume there's some hash checking).


eandre (approved):
Thanks!

PR #1053, opened 2024-02-28
Don't set application/json Content-Type on FormData bodies

### LINKED ISSUE #1052: JS/TS generated clients always set content-type of application/json, which makes uploading multipart/form-data difficult
When performing file uploads to raw endpoints using multipart/form-data we need to set a boundary in the `Content-Type` header to delineate the different parts of the request payload. `fetch` handles this natively for us when a FormData object is passed to it.

The auto-generated JS/TS clients are configured to set the `Content-Type` header to application/json for every request, without exception. Currently we can pass an overridden header for `Content-Type`, but that requires us to handle the boundary manually, which is not great.

Our options are to:
1. In our overridden `fetch` implementation, check the type of the body and delete the `application/json` header if present (as it's already set at this point)
2. Only set the `application/json` header if the body is not `FormData`, as in these cases we should allow fetch to natively handle the setting of `Content-Type` including boundary

macleodmac:
Fixes #1052

### DIFF
--- a/internal/clientgen/javascript.go
+++ b/internal/clientgen/javascript.go
@@ -475,10 +475,8 @@ class BaseClient {`)
 	js.WriteString(`
     constructor(baseURL, options) {
         this.baseURL = baseURL
-        this.headers = {
-            "Content-Type": "application/json",
-        }
-
+		this.headers = {}
+		
         // Add User-Agent header if the script is running in the server
         // because browsers do not allow setting User-Agent headers to requests
         if (typeof window === "undefined") {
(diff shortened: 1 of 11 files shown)


internal/clientgen/javascript.go
eandre, line 1: Maybe a better approach altogether is to drop the `this.headers` on the client altogether, and simply have each
endpoint add its own headers. The endpoint is the thing that knows how the body should be encoded, so it probably makes more sense to do it there.

eandre (commented):
Thanks @macleodmac! One comment that would be good to address; what do you think?

PR #1137, opened 2024-04-10
[FIX] typo in migrate-away.md

tomassar:
There's a small typo in the 'how to migrate away' encore documentation page. In 'What to expect when migrating away', it says this:

> This means that the changes required to migrate away will be almost exactly the same work you would have needed to do if you hadn't used Encore in the first **placem**, e.g. writing infrastructure boilerplate. There's no added migration cost.

Instead of 'placem' it should be 'place'.


eandre (approved):
Thanks!

PR #1143, opened 2024-04-11
Add ability to exclude services from code generation

Willyham:
This adds an `--excluded-services=foo` option to the CLI, which removes services from code generation set.

Defaulting services from metadata is now done outside the `Client` function to centralise the logic for determining the final service set.

> [!WARNING]
> This is untested as I didn't go through the effort to set up a proper environment for encore daemon dev.
> It also needs re-generating with the correct protoc version.


eandre:
Cleaned it up in #1145, so closing this!

PR #1179, opened 2024-04-27
sqldb: support podman

sudomateo:
Previously, running the `uptime` example Encore application using Podman that's correctly configured to emulate Docker would fail with the following error.

```sh
> encore run
  ✔ Building Encore application graph... Done!
  ✔ Analyzing service topology... Done!
  ❌ Creating PostgreSQL database cluster... Failed: failed to start cluster: check docker image: exit status 125
  ✔ Starting PubSub daemon... Done!
  ✔ Fetching application secrets... Done!
  ✔ Generating boilerplate code... Done!
  ✔ Compiling application source code... Done!
failed to start cluster: check docker image: exit status 125
```

This error did not provide enough information to understand what exactly failed without diving into the code. Luckily, I understood Podman well enough to check whether the code was using a short image name without an alias in Podman's configuration.

https://github.com/encoredev/encore/blob/a17db0e408b732ceaa28b95c1712d7b0fe7d9539/cli/daemon/sqldb/docker/docker.go#L369

I fixed that error by running `podman pull encoredotdev/postgres:15` to pull the image. However, other errors persisted.

This patch fixes those other errors and updates the code to support Podman for the `uptime` example Encore application. Additionally, it adds context to the `docker image inspect` error so it will now report the image name as seen below.

```sh
> encore run
  ✔ Building Encore application graph... Done!
  ✔ Analyzing service topology... Done!
  ❌ Creating PostgreSQL database cluster... Failed: failed to start cluster: check docker image: docker image inspect failed: encoredotdev/postgres:15: exit status 125
  ✔ Starting PubSub daemon... Done!
  ✔ Fetching application secrets... Done!
  ✔ Generating boilerplate code... Done!
  ✔ Compiling application source code... Done!
```

### DIFF
--- a/cli/daemon/sqldb/docker/docker.go
+++ b/cli/daemon/sqldb/docker/docker.go
@@ -200,6 +200,10 @@ func (d *Driver) clusterStatus(ctx context.Context, id sqldb.ClusterID) (status 
 			if bytes.Contains(out, []byte("No such container")) {
 				continue
 			}
+			// Podman has slightly different output when a container is not found.
+			if bytes.Contains(out, []byte("no such container")) {
+				continue
+			}
 			return nil, "", errors.Wrapf(err, "docker container inspect failed: %s", out)
 		} else {
 			// Found our container; use it.
@@ -230,7 +234,8 @@ func (d *Driver) clusterStatus(ctx context.Context, id sqldb.ClusterID) (status 
 		return nil, "", errors.Wrap(err, "parse `docker container inspect` response")
 	}
 	for _, c := range resp {
-		if c.Name == "/"+containerName {
+		// Docker prefixes `/` to the container name, Podman doesn't.
+		if c.Name == "/"+containerName || c.Name == containerName {
 			status := &sqldb.ClusterStatus{Status: sqldb.Stopped, Config: &sqldb.ConnConfig{
 				// Defaults if we don't find anything else configured.
 				Superuser: sqldb.Role{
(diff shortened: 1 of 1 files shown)


cli/daemon/sqldb/docker/docker.go
```diff
@@ -200,6 +200,10 @@ func (d *Driver) clusterStatus(ctx context.Context, id sqldb.ClusterID) (status
 			if bytes.Contains(out, []byte("No such container")) {
 				continue
 			}
+			// Podman returns an empty array when a container is not found.
+			if bytes.Contains(out, []byte("[]")) {
```
eandre, line 204: I'm a bit worried that checking for any string containing `[]` might lead to false negatives in certain situations. 

Does the podman output contain **only** `[]`? If so we could check something like `bytes.Equal(bytes.TrimSpace(out), []byte("[]"))`.

cli/daemon/sqldb/docker/docker.go
```diff
@@ -353,8 +366,11 @@ func ImageExists(ctx context.Context) (ok bool, err error) {
 		return true, nil
 	case bytes.Contains(out, []byte("No such image")):
 		return false, nil
+	// Podman returns an empty array when an image is not found.
+	case bytes.Contains(out, []byte("[]")):
```
eandre, line 370: Same here

eandre (commented):
Thanks so much for doing this! Looks great overall; just one minor comment that would be good to address.

cli/daemon/sqldb/docker/docker.go
sudomateo: Stdout will contain only `[]`, but stderr will contain the longer error text.

```sh
> podman container inspect foo > stdout 2> stderr
> cat stdout
[]
> cat stderr
Error: no such container "foo"
```

```sh
> podman image inspect foo > stdout 2> stderr
> cat stdout
[]
> cat stderr
Error: unable to inspect "foo": failed to find image foo: foo: image not known
```

This code path is currently using `CombinedOutput` to search across both stdout and stderr. If we want to use `bytes.Equal(bytes.TrimSpace(out), []byte("[]"))` we'd have to search stdout only. If we want to keep using `CombinedOutput` I can do the following:
- `bytes.Contains(out, []byte("no such container"))` for `docker container inspect`
- `bytes.Contains(out, []byte("failed to find image"))` for `docker image inspect`

What do you think?

eandre: Makes sense. Yes, in that case checking for those strings make more sense I think.

sudomateo: Done!

eandre (approved):
Thanks very much!

PR #1219, opened 2024-05-21
http2: Use http2 transport for h2c only

ekerfelt:


runtimes/go/appruntime/apisdk/api/transport/eh2c.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+	"crypto/tls"
+	"net"
+	"net/http"
+
+	"golang.org/x/net/http2"
+)
+
+func NewH2CTransport() http.RoundTripper {
```
eandre, line 12: Maybe accept the non-h2c transport to use as an argument instead of implicitly using `http.DefaultTransport`?

runtimes/go/appruntime/apisdk/api/transport/eh2c.go
```diff
@@ -0,0 +1,35 @@
...
+			},
+		},
+	}
+}
+
+type H2CTransport struct {
+	h2c http.RoundTripper
+}
+
+func (h H2CTransport) RoundTrip(request *http.Request) (*http.Response, error) {
```
eandre, line 28: Since we're returning a pointer from `NewH2CTransport, it's clearer if the method is on a pointer too

```suggestion
func (h *H2CTransport) RoundTrip(request *http.Request) (*http.Response, error) {
```

eandre (approved):

PR #1248, opened 2024-06-05
typescript: add logs to traces

fredr:
Adds logs rows to traces in typescript

### DIFF
--- a/runtimes/core/src/log/logger.rs
+++ b/runtimes/core/src/log/logger.rs
@@ -1,12 +1,15 @@
 use crate::error::AppError;
 use crate::log::fields::{FieldConfig, DEFAULT_FIELDS, GCP_FIELDS};
 use crate::log::writers::{default_writer, Writer};
-use crate::model;
+use crate::log::LogLevel;
+use crate::model::{self, LogField};
+use crate::trace::protocol::LogMessageData;
+use crate::trace::Tracer;
 use anyhow::Context;
 use env_logger::filter::Filter;
 use log::{Log, Metadata, Record};
 use std::collections::BTreeMap;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time::SystemTime;
 
 pub type Fields = BTreeMap<String, serde_json::Value>;
@@ -19,6 +22,7 @@ pub struct Logger {
     field_config: &'static FieldConfig,
     writer: Arc<dyn Writer>,
     extra_fields: Fields,
+    tracer: Arc<RwLock<Tracer>>,
 }
 
 impl Logger {
@@ -34,7 +38,13 @@ impl Logger {
             field_config,
             writer: default_writer(field_config),
             extra_fields: Fields::new(),
+            tracer: Arc::new(RwLock::new(Tracer::noop())),
         }
+    }
+
+    pub fn set_tracer(&self, tracer: Tracer) {
+        let mut t = self.tracer.write().expect("tracer lock poisoned");
+        *t = tracer;
     }
 
     /// Returns a new logger with the given log level.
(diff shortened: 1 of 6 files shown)


runtimes/core/src/model/mod.rs
```diff
@@ -288,3 +288,55 @@ pub struct AuthSuccessResponse {
...
+impl From<crate::log::LogLevel> for LogLevel {
+    fn from(level: crate::log::LogLevel) -> Self {
+        match level {
+            crate::log::LogLevel::Trace => LogLevel::Trace,
+            crate::log::LogLevel::Debug => LogLevel::Debug,
+            crate::log::LogLevel::Info => LogLevel::Info,
+            crate::log::LogLevel::Warn => LogLevel::Warn,
+            crate::log::LogLevel::Error => LogLevel::Error,
+
+            // TODO(fredr): these are not supported by the go runtime
```
fredr, line 337: Not sure how we should handle this :thinking:

runtimes/core/src/log/logger.rs
```diff
@@ -202,6 +212,8 @@ impl Logger {
             }
         }
 
+        self.write_to_trace(request, level, &msg, &values);
```
eandre, line 215: We're inserting the message as a value above. Do we need to pass it in here, too? It requires cloning the message, per above.

runtimes/core/src/log/logger.rs
```diff
@@ -235,6 +247,64 @@ impl Logger {
...
+                        key,
+                        value: model::LogFieldValue::I64(num.as_i64().unwrap()),
+                    });
+                } else if num.is_u64() {
+                    trace_fields.push(LogField {
+                        key,
+                        value: model::LogFieldValue::U64(num.as_u64().unwrap()),
+                    });
+                } else if num.is_f64() {
+                    println!("is_f64");
```
eandre, line 274: Remove

runtimes/core/src/model/mod.rs
```diff
@@ -288,3 +288,55 @@ pub struct AuthSuccessResponse {
     /// The user data.
     pub user_data: serde_json::Map<String, serde_json::Value>,
 }
+
+// matches go runtime
+pub enum LogLevel {
```
eandre, line 293: In the logging overhaul PR I removed the log level enum we had and relied on `::log::Level` instead. Maybe better to do so here, too?

runtimes/core/src/trace/protocol.rs
```diff
@@ -57,6 +57,57 @@ impl Tracer {
...
+pub struct LogMessageData<'a> {
+    pub source: Option<&'a Request>,
+    pub msg: &'a str,
+    pub level: LogLevel,
+    pub fields: Vec<LogField<'a>>,
+}
+
+impl Tracer {
+    #[inline]
+    pub fn log_message(&self, data: LogMessageData) -> Option<TraceEventId> {
```
eandre, line 69: I think we don't need to return an event id. It's mainly used for log events that need to correlate "start" and "end", like beginning a database query and recording its completion.

eandre (commented):
Overall this looks great! A few minor nitpicks :)

runtimes/core/src/log/logger.rs
```diff
@@ -236,6 +248,65 @@ impl Logger {
...
+
+        let _ = self
+            .tracer
+            .read()
+            .expect("tracer lock poisoned")
+            .log_message(LogMessageData {
+                source: request,
+                msg,
+                level,
+                fields: trace_fields,
```
eandre, line 307: Instead of allocating a vec, could we pass in an iterator instead?

eandre (approved):

PR #1249, opened 2024-06-05
runtimes/core: overhaul runtime logging

eandre:
This improves how we handle log levels for the
Rust runtime.

The runtime now supports `RUST_LOG` matching the
`env_logger` log filter syntax, and uses `ENCORE_LOG`
to set the application log level. Additionally, the
`ENCORE_RUNTIME_LOG` env variable can be used
as a simpler version of `RUST_LOG` to only set the
`encore_runtime_core` log level.


ekerfelt (dismissed):

fredr (approved):
lgtm! :ship:

PR #1262, opened 2024-06-11
runtimes/core: url-encode api call parameters

eandre:


fredr (approved):
Tested in my local env, and it work! lgtm

PR #1266, opened 2024-06-13
runtimes/core: randomize sqldb port

eandre:
To avoid running into issues when another process
is also running, since Tokio uses SO_REUSEPORT.

Thanks Olle Eriksson for the report.


runtimes/core/src/sqldb/manager.rs
```diff
@@ -15,26 +15,48 @@ use crate::trace::Tracer;
...
+    pub secrets: &'a secrets::Manager,
+    pub tracer: Tracer,
+    pub runtime: tokio::runtime::Handle,
+}
+
+impl<'a> ManagerConfig<'a> {
+    pub fn build(self) -> anyhow::Result<Manager> {
+        // Start listening so we can tell which port it is when we generate configuration.
+        let listener =
+            std::net::TcpListener::bind("127.0.0.1:0").context("unable to bind to sqldb port")?;
```
fredr, line 34: will this cause a random port to be selected? or what does `:0` do?

fredr: Ah found the docs for it https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.bind
TIL

fredr (approved):
lgtm

PR #1274, opened 2024-06-24
encore.ts: add `Service` resource to specify service root

eandre:


tsparser/src/parser/parser.rs
```diff
@@ -109,38 +111,132 @@ impl ParseContext {
...
+                "node_modules" | "encore.gen" => true,
+                x => x.starts_with('.'),
+            }
         }
-        self.inject_generated_service_clients();
 
-        let binds = resolve_binds(self.binds)?;
-        let resolver = UsageResolver::new(&self.pc.loader, &self.resources, &binds);
+        fn is_service(e: &walkdir::DirEntry) -> bool {
+            e.path().ends_with("encore.service.ts")
```
fredr, line 139: very much a nit, but this would return true for filenames that ends with this, so a file called blaha.encore.service.ts would also return true (or is that what we want maybe?). Otherwise you could maybe use `e.path().file_name()' and check `.is_file()`

tsparser/src/parser/resources/apis/service.rs
```diff
@@ -0,0 +1,91 @@
...
+                        h.struct_span_err(
+                            r.range,
+                            "cannot have multiple service declarations in the same module",
+                        )
+                        .emit();
+                    });
+                    continue;
+                }
+
+                let _cfg = r.config.unwrap_or_default();
```
fredr, line 49: this doesnt seem to be used? so maybe could be removed

tsparser/src/parser/usageparser/mod.rs
```diff
@@ -202,7 +202,7 @@ impl<'a> UsageResolver<'a> {
 #[derive(Debug)]
 pub enum Usage {
     CallEndpoint(apis::api::CallEndpointUsage),
-    ReferenceEndpoint(apis::api::ReferenceEndpointUsage),
+    // ReferenceEndpoint(apis::api::ReferenceEndpointUsage),
```
fredr, line 205: maybe remove this as well? or are you planning on re-adding it?

fredr (approved):
lgtm! added some optional nits and questions

tsparser/src/parser/parser.rs
fredr: ah wait, ends_with on path only checks whole path components, didnt know that!

eandre: Yeah exactly :)

fredr (approved):

PR #1275, opened 2024-06-25
Refactor gateway in rust runtime

fredr:
Rewrite gateway using cloudflares pingora, so that we can support more advanced upstream proxying, like websockets

Note: I've copied over the cors layer code from tower_http so that we could use it with pingora


runtimes/core/src/api/gateway/mod.rs
```diff
@@ -249,11 +259,15 @@ impl ProxyHttp for Gateway {
 
         upstream_request.set_uri(new_uri);
 
+        // Do we need to set the host header here?
+        // It means the upstream service won't be able to tell
+        // what the original Host header was, which is sometimes useful.
```
fredr, line 264: it depends a bit on how the upstream works, if the upstream server handles multiple hosts or it should terminate tls in will need the host to figure out which certificate it should use

fredr: another thing to consider regarding dns-based upstreams is that we resolve the dns, but just take the first ip, which might be fine depending on how the upstreams are configured

eandre (approved):

runtimes/core/src/api/gateway/mod.rs
```diff
@@ -1,177 +1,430 @@
...
+                test: false,
+                conf: None,
+            })
+            .unwrap(),
+        );
+        let mut proxy = http_proxy_service(&conf, self);
+        let listen_addr = listener.local_addr().unwrap().to_string();
+
+        // unbind the address and let pingora re-bind it
+        drop(listener);
```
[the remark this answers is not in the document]
fredr: I know, we need to create it in the case we let os decide the port (when do add "127.0.0.1:0")

PR #1276, opened 2024-06-25
Import pubsub subscriptions in entrypoint

eandre:
Fixes an issue where they weren't initialized if they
were not otherwise being imported.


fredr (approved):
looks good!

PR #1278, opened 2024-06-26
set level for all targets prefixed with encore_

fredr:
Since we have several targets prefixed with `encore_`, like `encore_runtime_core` and `encore_js_runtime`.

We could also specify them all rather that use the prefix if we want more control.


eandre (approved):

eandre:
Didn't know this syntax worked :)

fredr:
yeah, its not super obvious from the documentation, but the target is just a prefix search

PR #1280, opened 2024-07-01
Add docs for TS logging

ekerfelt:


docs/ts/develop/logging.md
```diff
@@ -0,0 +1,42 @@
...
+  import: "encore.dev/rlog",
+}
+---
+
+Encore offers built-in support for Structured Logging, which combines a free-form log message with structured and type-safe key-value pairs. This enables straightforward analysis of what your application is doing, in a way that is easy for a computer to parse, analyze, and index. This makes it simple to quickly filter and search through logs.
+
+Encore’s logging is integrated with the built-in [Distributed Tracing](/docs/observability/tracing) functionality, and all logs are automatically included in the active trace. This dramatically simplifies debugging of your application.
+
+## Usage
+First, add `import log  from "encore.dev/log";` to your module. Then simply call one of functions `Info`, `Error`, or `Debug`. For example:
```
eandre, line 17:
```suggestion
First, add `import log  from "encore.dev/log";` to your module. Then call one of the logging functions `error`, `warn`, `info`, `debug`, or `trace` to emit a log message. For example:
```

docs/ts/develop/logging.md
```diff
@@ -0,0 +1,42 @@
...
+  })
+log.error(err, "something went terribly wrong!")
+```
+
+The first parameter is the log message (or optionally an error for the error function) . After that follows a single object with key-value pairs for structured logging.
+
+If you’re logging many log messages with the same key-value pairs each time it can be a bit cumbersome. To help with that, use `log.with()` to group them into a Logger object, which then copies the key-value pairs into each log event:
+
+```ts
+const logger = log.with({user_id: 12345})
```
eandre, line 32: Maybe use a different example, since user id is automatically added by Encore already. How about `{is_subscriber: true}` here, and `{login_method: "oauth2"}` below?

ekerfelt: should we update the go logging example too then?

eandre: Yeah :)

docs/ts/develop/logging.md
```diff
@@ -0,0 +1,42 @@
...
+Encore’s logging is integrated with the built-in [Distributed Tracing](/docs/observability/tracing) functionality, and all logs are automatically included in the active trace. This dramatically simplifies debugging of your application.
+
+## Usage
+First, add `import log  from "encore.dev/log";` to your module. Then call one of the logging functions `error`, `warn`, `info`, `debug`, or `trace` to emit a log message. For example:
+
+```ts
+log.info("log message", {
+    user_id: 12345, 
+    is_subscriber: true
+  })
```
eandre, line 23:
```suggestion
log.info("log message", {is_subscriber: true})
```

eandre (approved):

PR #1281, opened 2024-07-01
Add support for debugging TS apps

ekerfelt:


docs/ts/develop/debug.md
```diff
@@ -0,0 +1,56 @@
...
+### Visual Studio Code
+To debug with VS Code you must first add a debug configuration. Press `Run -> Add Configuration`, choose `Node.js -> Attach`. The generated config should look something like this:
+
+```json
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Attach",
+      "port": 9229,
```
eandre, line 36: Is this always 9229 or will it vary?

ekerfelt: it's 9229 by default. You can override it, but I haven't added support for it :). If 9229 isn't available, it will not init the debugger. If we want to make this more fail-safe we could either automatically look for an available port, or add an override debug port option.

eandre (approved):

PR #1282, opened 2024-07-02
Add support for clearing traces from the local dashboard

ekerfelt:


cli/daemon/engine/trace2/sqlite/write.go
```diff
@@ -29,10 +35,103 @@ type Store struct {
...
+					if err != nil {
+						log.Error().Err(err).Msg("failed to get old trace ids")
+						continue
+					}
+					traceIDs, err := scanRows[string](rows)
+					if len(traceIDs) == 0 {
+						continue
+					}
+					idArgs := strings.Join(fns.Map(traceIDs, pq.QuoteLiteral), ",")
+					res, err := s.db.ExecContext(ctx, "DELETE FROM trace_event WHERE app_id = ? AND trace_id IN ("+idArgs+")", appID)
```
eandre, line 93: In Postgres at least you can do `trace_id = ANY(?)` and pass in a string slice, instead of having to construct the comma-separated string arg by hand. Not sure if that works in sqlite though

ekerfelt: yeah, it doesn't unfortunately :(. There's a bunch of libs solving this in different ways, but I figured this was good enough.

eandre (approved):

PR #1283, opened 2024-07-03
Use dlv for debugging if available

ekerfelt:


tsparser/src/builder/transpiler.rs
```diff
@@ -44,7 +44,7 @@ pub struct TranspileParams<'a> {
     pub inputs: Vec<Input>,
 
     // If we should transpile for debug
-    pub debug: bool,
+    pub debug: u8,
```
eandre, line 47: This would be better represented with a rust enum type :)

tsparser/src/builder/mod.rs
```diff
@@ -182,3 +182,11 @@ impl std::fmt::Display for PlainError {
         write!(f, "{}", self.0)
     }
 }
+
+#[derive(Deserialize, Debug)]
```
eandre, line 186: If you also derive Clone and Copy you don’t need to take references to the value

tsparser/src/builder/compile.rs
```diff
@@ -22,7 +22,7 @@ pub struct CompileParams<'a> {
     pub working_dir: &'a Path,
     pub desc: &'a AppDesc,
     pub use_local_runtime: bool,
-    pub debug: u8,
+    pub debug: &'a DebugMode,
```
eandre, line 25: Make this just `DebugMode` without the reference (see below)

tsparser/src/bin/tsparser-encore.rs
```diff
@@ -149,7 +149,7 @@ fn main() -> Result<()> {
                                 working_dir: &cwd,
                                 desc: parse,
                                 use_local_runtime: input.use_local_runtime,
-                                debug: input.debug,
+                                debug: &input.debug,
```
eandre, line 152: Revert (see below)

tsparser/src/builder/transpiler.rs
```diff
@@ -116,8 +115,8 @@ impl OutputTranspiler for EsbuildCompiler<'_> {
                 ];
 
                 match p.debug {
-                    1 => command.push("--inspect".to_string()),
-                    2 => command.push("--inspect-brk".to_string()),
+                    DebugMode::Enabled => command.push("--inspect".to_string()),
+                    DebugMode::Break => command.push("--inspect-brk".to_string()),
                     _ => {}
```
eandre, line 120: Remove the catch-all in favor of only handling the Disabled case explicitly. That way we’ll get a compiler error here if we add more enum cases

eandre (approved):

PR #1289, opened 2024-07-08
Update go hello world example in readme

fredr:
Since it is under the "Example: Hello World" section. Make is a similar to the TS example as possible so it is easier to compare.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,10 +71,18 @@ Using the Go Backend SDK, it looks like so:
 ```go
 package hello
 
-//encore:api public
-func Ping(ctx context.Context, params *PingParams) (*PingResponse, error) {
-    msg := fmt.Sprintf("Hello, %s!", params.Name)
-    return &PingResponse{Message: msg}, nil
+import (
+	"context"
+)
+
+//encore:api public path=/hello/:name
+func World(ctx context.Context, name string) (*Response, error) {
+	msg := "Hello, " + name + "!"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,10 +71,18 @@ Using the Go Backend SDK, it looks like so:
 ```go
 package hello
 
-//encore:api public
-func Ping(ctx context.Context, params *PingParams) (*PingResponse, error) {
-    msg := fmt.Sprintf("Hello, %s!", params.Name)
-    return &PingResponse{Message: msg}, nil
+import (
+	"context"
+)
```
eandre, line 76: I'd say remove the import statement and use `fmt.Sprintf` below so it's more idiomatic Go

README.md
```diff
@@ -71,10 +71,14 @@ Using the Go Backend SDK, it looks like so:
 ```go
 package hello
 
-//encore:api public
-func Ping(ctx context.Context, params *PingParams) (*PingResponse, error) {
-    msg := fmt.Sprintf("Hello, %s!", params.Name)
-    return &PingResponse{Message: msg}, nil
+//encore:api public path=/hello/:name
+func World(ctx context.Context, name string) (*Response, error) {
+	msg := fmt.Sprintf("Hello, %s", name)
```
eandre, line 76:
```suggestion
	msg := fmt.Sprintf("Hello, %s!", name)
```

PR #1291, opened 2024-07-10
runtimes/core: Remove temporary bind of gateway listener

fredr:
The downside of this is that pingora will silently fail if the address in not bindable (or it will log out that it is trying to bind, and after 30 tries it will log out that is has failed, but we currently filter out those logs)

### DIFF
--- a/runtimes/core/src/api/gateway/mod.rs
+++ b/runtimes/core/src/api/gateway/mod.rs
@@ -1,7 +1,7 @@
 mod router;
 
 use std::borrow::Cow;
-use std::net::{SocketAddr, TcpListener};
+use std::net::SocketAddr;
 use std::sync::Arc;
 
 use anyhow::Context;
@@ -109,7 +109,7 @@ impl Gateway {
         self.inner.shared.auth.as_ref()
     }
 
-    pub async fn serve(self, listener: TcpListener) -> anyhow::Result<()> {
+    pub async fn serve(self, listen_addr: String) -> anyhow::Result<()> {
         let conf = Arc::new(
             ServerConf::new_with_opt_override(&Opt {
                 upgrade: false,
@@ -121,10 +121,7 @@ impl Gateway {
             .unwrap(),
         );
         let mut proxy = http_proxy_service(&conf, self);
-        let listen_addr = listener.local_addr().unwrap().to_string();
 
-        // unbind the address and let pingora re-bind it
-        drop(listener);
         proxy.add_tcp(&listen_addr);
 
         let (_tx, rx) = watch::channel(false);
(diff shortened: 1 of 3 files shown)


runtimes/core/src/api/gateway/mod.rs
```diff
@@ -109,7 +109,7 @@ impl Gateway {
         self.inner.shared.auth.as_ref()
     }
 
-    pub async fn serve(self, listener: TcpListener) -> anyhow::Result<()> {
+    pub async fn serve(self, listen_addr: String) -> anyhow::Result<()> {
```
eandre, line 112: Given that we pass it as a ref to `proxy.add_tcp`, can we make this take `&str`?

runtimes/core/src/api/manager.rs
```diff
@@ -42,7 +42,7 @@ pub struct ManagerConfig<'a> {
 }
 
 pub struct Manager {
-    gateway_listener: Mutex<Option<std::net::TcpListener>>,
+    gateway_listener: Option<String>,
```
eandre, line 45: It's no longer a listener, but the listen address

runtimes/core/src/api/manager.rs
```diff
@@ -58,9 +58,7 @@ impl ManagerConfig<'_> {
     pub fn build(mut self) -> anyhow::Result<Manager> {
         let gateway_listener = if !self.hosted_gateway_rids.is_empty() {
```
eandre, line 59: Same here

runtimes/core/src/api/manager.rs
```diff
@@ -337,7 +335,7 @@ impl Manager {
         let server = HttpServer::new(encore_routes, api, fallback);
 
         let api_listener = self.api_listener.lock().unwrap().take();
-        let gateway_listener = self.gateway_listener.lock().unwrap().take();
+        let gateway_listener = self.gateway_listener.clone();
```
eandre, line 338: This can be `.take()` instead

fredr: Then I would have to put it back into a mutex. I can do that if you prefer

fredr: or I can just do .as_ref actually!

eandre: Ah fair, can leave it as-is then. It was the case before that we would use `None` to detect that the server had already been started, but that's no longer the case so not much point in using `take()`

fredr: as_ref actually didnt work here, due to lifetimes (since we borrow self)

eandre (approved):

PR #1292, opened 2024-07-11
runtimes/core: handle lack of services and gateways

eandre:


runtimes/core/src/api/manager.rs
```diff
@@ -375,6 +375,10 @@ impl Manager {
                 res = async { api_fut.unwrap().await }, if api_fut.is_some() => {
                     res.context("serve api").inspect_err(|err| log::error!("api server failed: {:?}", err))?;
                 },
+                else => {
+                    // Nothing to serve.
+                    ::log::debug!("no api server or gateway to serve");
```
fredr, line 380: Maybe this should be a warn or error? Assuming that something would be wrong if we have nothing to serve?

eandre: It's actually expected in some situations that I encountered yesterday, so that's why I made it a debug statement

fredr (approved):
:ship:

PR #1315, opened 2024-08-09
TS: Add support for streaming apis

fredr:
Adds support for a first version of streaming apis to typescript apps.

It adds three different kinds of streams:
```
// bidirectional stream
api.streamInOut()

// from server outgoing stream
api.streamOut()

// to server incoming stream
api.streamIn()
```

It adds a new schema for the handshake, which is the websocket upgrade request from which path parameter, query parameters and headers can be extracted.

The stream handler can take either one or two parameters, the handshake data and a stream object. The handshake is optional and only needs to be added if data needs to be extracted from the upgrade request.

The stream object added as a parameter is different for the different stream types, for inout it can be used to receive and send messages, for in stream it can be used to receive only and for out streams it can be used to send data only.

A in stream can also return a value (optionally) that will be sent back to the client as a response (still sent over the socket).

The stream object implements async stream both on the server side and the client side to make it easy to iterate over incoming messages.

### Super simple ping pong example implementation
Server:
```typescript
interface Ping {
  payload: string;
}
interface Pong {
  count: number;
}

export const stream = api.streamInOut<Ping, Pong>(
  { expose: true, path: "/stream" },
  async (stream) => {
    let count = 1;

    for await (const ping of stream) {
      log.info(`received ping with payload ${ping.payload}`);
      await stream.send({ count: count++ });
    }
  },
);
```

Client:
```typescript
const pingPongStream = await client.pingPong.stream();

pingPongStream.send({ payload: "initial ping" });

for await (const pong of pingPongStream) {
  console.log(`receive pong ${pong.count}`);
  setTimeout(() => {
    pingPongStream.send({ payload: `ping after pong ${pong.count}` });
  }, 1000);
}
```


e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -770,6 +770,223 @@ function mustBeSet<A>(field: string, value: A | null | undefined): A {
...
+            // normal closure, no reconnect
+            if (event.code === 1005 || event.code === 1000) {
+                this.done = true;
+            }
+            if (!this.done) {
+                const delay = Math.min(this.minDelayMs * 2 ** this.retry, this.maxDelayMs);
+                console.log(`Reconnecting to ${this.url} in ${delay}ms`);
+                this.retry += 1;
+                setTimeout(() => {
+                    this.ws = this.connect();
```
eandre, line 838: Is automatic reconnect desirable?

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -770,6 +770,223 @@ function mustBeSet<A>(field: string, value: A | null | undefined): A {
...
+                yield this.buffer.shift() as Response;
+            } else {
+                if (this.connection.done) return;
+                await this.connection.hasUpdate();
+            }
+        }
+    }
+
+
+}
```
eandre, line 920: Newlines

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -830,6 +1047,97 @@ class BaseClient {
 
     }
 
+	async getAuthData(): Promise<CallParameters | undefined> {
+        let authData
```
eandre, line 1051: Indentation

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -830,6 +1047,97 @@ class BaseClient {
...
+            data.query["query"] = authData.Query.map((v) => String(v))
+            data.query["new-auth"] = String(authData.NewAuth)
+            data.headers = {};
+            data.headers["x-header"] = authData.Header
+            data.headers["x-auth-int"] = String(authData.AuthInt)
+            data.headers["authorization"] = authData.Authorization
+
+            return data;
+        }
+        return;
```
eandre, line 1076: Newline before this return

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -830,6 +1047,97 @@ class BaseClient {
...
+    }
+    // createBidiStream sets up a stream to a streaming api
+    async createBidiStream<Request, Response>(path: string, params?: CallParameters): Promise<BidiStream<Request, Response>> {
```
eandre, line 1079: Missing newline before comment, and rewrite the comment

runtimes/core/src/api/gateway/websocket.rs
```diff
@@ -0,0 +1,329 @@
...
+// hack to be able to have browsers send request headers when setting up a websocket
+// inspired by https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0
+pub fn update_headers_from_websocket_protocol(
+    upstream_request: &mut RequestHeader,
+) -> anyhow::Result<()> {
+    let headers = upstream_request
+        .headers
+        .get_all(SEC_WEBSOCKET_PROTOCOL)
+        .into_iter()
+        .cloned()
```
eandre, line 23: Do we need to clone this into a vec? Or can we just iterate?

runtimes/core/src/api/gateway/websocket.rs
```diff
@@ -0,0 +1,329 @@
...
+        | TE
+        | TRAILER
+        | TRANSFER_ENCODING
+        | UPGRADE
+        | VIA => true,
+        ref n if n == HeaderName::from_static("keep-alive") => true,
+        ref n if n == HeaderName::from_static("permissions-policy") => true,
+        ref n => {
+            let name = n.to_string();
+            name.strip_prefix("sec-").is_some() || name.strip_prefix("proxy-").is_some()
```
eandre, line 125: Is there not an `has_prefix` or similar?

runtimes/core/src/api/gateway/websocket.rs
```diff
@@ -0,0 +1,329 @@
...
+        | REFERER
+        | TE
+        | TRAILER
+        | TRANSFER_ENCODING
+        | UPGRADE
+        | VIA => true,
+        ref n if n == HeaderName::from_static("keep-alive") => true,
+        ref n if n == HeaderName::from_static("permissions-policy") => true,
+        ref n => {
+            let name = n.to_string();
```
eandre, line 124: Is there `as_str()` or similar to avoid the allocation?

runtimes/core/src/api/websocket.rs
```diff
@@ -0,0 +1,324 @@
...
+        let (incoming_messages_tx, incoming_message_rx) = mpsc::unbounded_channel();
+
+        let schema = SocketSchema { incoming, outgoing };
+        tokio::spawn({
+            async move {
+                loop {
+                    tokio::select! {
+                        msg = websocket.recv() => match msg {
+                            None => {
+                                log::debug!("websocket closed");
```
eandre, line 141: Downgrade to trace?

runtimes/core/src/api/websocket.rs
```diff
@@ -0,0 +1,324 @@
...
+                            Some(Err(e)) => {
+                                log::debug!("websocket receive failed: {e}");
+                                break;
+                            }
+                        },
+                        msg = outgoing_messages_rx.recv() => {
+                            match msg {
+                                None => {
+                                    _ = websocket.close().await;
+                                    log::debug!("websocket closed");
```
eandre, line 165: Trace?

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -96,5 +125,59 @@ api.raw = function raw(options: APIOptions, fn: RawHandler) {
...
 
+export interface StreamIn<Request> extends AsyncIterable<Request> {
+  recv: () => Promise<Request>;
+}
+
+export interface StreamOut<Response> {
+  send: (msg: Response) => Promise<void>;
+  close: () => Promise<void>;
+}
+export type StreamBidi<Request, Response> = StreamIn<Request> &
```
eandre, line 136:
```suggestion
}

export type StreamBidi<Request, Response> = StreamIn<Request> &
```

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -96,5 +125,59 @@ api.raw = function raw(options: APIOptions, fn: RawHandler) {
...
+  fn: (
+    data: HandshakeData,
+    stream: StreamBidi<Request, Response>
+  ) => Promise<void>
+): void;
+function streamBidi<Request, Response>(
+  options: StreamOptions,
+  fn: (stream: StreamBidi<Request, Response>) => Promise<void>
+): void;
+function streamBidi(options: APIOptions, fn: any): typeof fn {
```
eandre, line 150:
```suggestion
function streamBidi(options: StreamOptions, fn: any): typeof fn {
```

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -96,5 +125,59 @@ api.raw = function raw(options: APIOptions, fn: RawHandler) {
...
+
+function streamIn<Request, Response>(
+  options: StreamOptions,
+  fn: (stream: StreamIn<Request>) => Promise<Response>
+): void;
+function streamIn<HandshakeData, Request, Response>(
+  options: StreamOptions,
+  fn: (data: HandshakeData, stream: StreamIn<Request>) => Promise<Response>
+): void;
+function streamIn(options: APIOptions, fn: any): typeof fn {
```
eandre, line 162:
```suggestion
function streamIn(options: StreamOptions, fn: any): typeof fn {
```

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -96,5 +125,59 @@ api.raw = function raw(options: APIOptions, fn: RawHandler) {
...
+
+function streamOut<HandshakeData, Response>(
+  options: StreamOptions,
+  fn: (data: HandshakeData, stream: StreamOut<Response>) => Promise<void>
+): void;
+function streamOut<Response>(
+  options: StreamOptions,
+  fn: (stream: StreamOut<Response>) => Promise<void>
+): void;
+function streamOut(options: APIOptions, fn: any): typeof fn {
```
eandre, line 174:
```suggestion
function streamOut(options: StreamOptions, fn: any): typeof fn {
```

runtimes/js/src/websocket_api.rs
```diff
@@ -0,0 +1,186 @@
...
+                let status = self.handler.call(
+                    WsRequestMessage {
+                        tx,
+                        payload,
+                        req: Request::new(req),
+                    },
+                    ThreadsafeFunctionCallMode::Blocking,
+                );
+
+                log::trace!("js ws handler responded with status: {status}");
```
eandre, line 42: Probably remove?

tsparser/src/parser/resources/apis/api.rs
```diff
@@ -184,7 +189,11 @@ pub const ENDPOINT_PARSER: ResourceParser = ResourceParser {
             let methods = r.config.method.unwrap_or(Methods::Some(vec![Method::Post]));
 
             let raw = matches!(r.kind, EndpointKind::Raw);
```
eandre, line 191: Could arguably remove this variable now and use `true/false` since you moved the call to `describe_endpoint` into the `match`?

tsparser/src/parser/resources/apis/api.rs
```diff
@@ -194,14 +203,44 @@ pub const ENDPOINT_PARSER: ResourceParser = ResourceParser {
                         None => None,
                         Some(t) => Some(pass.type_checker.resolve_type(module.clone(), &t)),
                     };
-                    (request, response)
+
+                    describe_endpoint(pass.type_checker, methods, path, request, response, raw)?
+                }
+                EndpointKind::TypedStream {
```
eandre, line 209: Maybe better to move this match case below `EndpointKind::Raw`, since that one is simpler and more closely aligned with `EndpointKind::Typed`

runtimes/core/src/api/gateway/websocket.rs
fredr: We can't unfortunately have an iterator only here, as we then take a reference the headers map, and then later when we mutate the headers pingora takes a mutable reference.

I think this could be solved differently if pingora exposed a way to access the entry api to mutate in place, maybe. But since this only happens if the request is an upgrade request, and since 99.9999% of the time it will only be one header (e.g javascript/nodejs will send one comma separated header), it wont be too bad I think. I only added the "get_all" here in case several headers are sent instead of one comma separated header, which is supported according to the spec (they both mean the same thing), but ofc something we can also decide not to support and just parse one header.

e2e-tests/testdata/echo_client/ts/client.ts
fredr: In my mind it is the expected behavior. Not that this only reconnects on un-graceful terminations.

Otherwise if there is a deploy or a network issue, the connection will be lost. Alternatively we could ofc make it optional or make it possible to setup via some callbacks

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -770,6 +770,223 @@ function mustBeSet<A>(field: string, value: A | null | undefined): A {
...
+
+    connect(): WebSocket {
+        const ws = new WebSocket(this.url, this.protocols)
+
+        ws.addEventListener("open", (_event: any) => {
+            this.retry = 0;
+        });
+
+        ws.addEventListener("error", (event: any) => {
+          console.error(event.error);
```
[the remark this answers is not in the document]
fredr: I'll remove it and add a on error handle, it can be useful for debugging to have some logout on network errors etc

e2e-tests/testdata/echo_client/js/client.js
```diff
@@ -552,6 +552,211 @@ function mustBeSet(field, value) {
...
+
+}
+export class OutStream {
+    constructor(url, headers) {
+        let responseResolver;
+        this.responseValue = new Promise((resolve) => responseResolver = resolve);
+
+        this.connection = new WebSocketConnection(url, headers);
+        this.connection.setMsgHandler((event) => {
+            responseResolver(JSON.parse(event.data))
```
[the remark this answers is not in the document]
fredr: yes, we set it a few lines up by calling
```
new Promise((resolve) => responseResolver = resolve)
```

the callback there is called synchronously by the Promise constructor, so it should work from my understanding (and the ts version doesn't complain about that as an extra insurance)

I have seen similar patterns out in the wild, like here e.g https://github.com/nodejs/node/blob/e020dd8610978f752ff014b3d5558a3fadee69c9/lib/internal/util.js#L562-L571

.prettierrc.toml
```diff
@@ -0,0 +1 @@
+trailingComma = "none"
```
fredr, line 1: this I added because _most_ of our ts files are formatted this way. Not sure if we want it, or if we should format all ts files with the standard `prettier` settings?

tsparser/src/parser/resources/apis/api.rs
fredr: we actually need the raw value further down when we create the `Endpoint`. I switched to `true/from` for the describe calls, as that is maybe a bit cleaner

e2e-tests/testdata/echo_client/tsconfig.json
```diff
@@ -12,7 +12,7 @@
 
     /* Language and Environment */
     "target": "ESNext",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
-    "lib": [ "ES6", "DOM" ],                             /* Specify a set of bundled library declaration files that describe the target runtime environment. */
+    "lib": [ "ES6", "DOM", "ES2021.String" ],            /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
fredr, line 15: added this as `replaceAll` is part of this version. I assume it shouldn't be a problem since we target `ESNext`

runtimes/core/src/api/gateway/websocket.rs
fredr: there is a `starts_with`! fixed

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -885,16 +892,20 @@ export class BidiStream<Request, Response> {
         this.connection.close();
     }
 
+    onError(handler: (event: any) => void) {
```
eandre, line 895: Maybe this would be more idiomatic as `.on("error", handler)`

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -972,16 +986,20 @@ export class OutStream<Request, Response> {
...
```
eandre, line 989: Same here

e2e-tests/testdata/echo_client/ts/client.ts
fredr: do you think we should then also add `.on("errror|message|close", ..)`. Or do we only expose "error" for now, and then keep it open for more things in the future?

fredr: I've update to `on("error")` now, but didnt add any other events, we can add them later if we want an api for those along with the async iterator. 

But right now there is only one event handler, and maybe it is expected that several handlers can be added with this on-api?

e2e-tests/testdata/echo_client/ts/client.ts
eandre: How does a reconnect behave with a handshake? In my experience some streaming APIs are stateful and indicate an action being taken, and starting a new stream would amount to resubmitting the action, which seems undesirable.

I think it might be better to leave out auto-reconnect for now. If/when we have the ability to tag an endpoint as idempotent it would make more sense.

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -770,6 +770,246 @@ function mustBeSet<A>(field: string, value: A | null | undefined): A {
...
+        this.errorHandler = handler;
+    }
+
+    close() {
+        this.done = true;
+        this.ws.close();
+    }
+}
+
+export type HandlerType = "error";
```
eandre, line 880: Remove this

e2e-tests/testdata/echo_client/ts/client.ts
```diff
@@ -770,6 +770,246 @@ function mustBeSet<A>(field: string, value: A | null | undefined): A {
...
+        this.connection.setMsgHandler((event: any) => {
+            this.buffer.push(JSON.parse(event.data));
+        });
+    }
+
+    close() {
+        this.connection.close();
+    }
+
+    on(type: HandlerType, handler: (event: any) => void) {
```
eandre, line 897:
```suggestion
    on(type: "error", handler: (event: any) => void) {
```

e2e-tests/testdata/echo_client/ts/client.ts
fredr: That makes sense, I'll leave it out for now, and maybe add some callbacks so that the caller could potentially implement it themselves if needed (I think I just need to add a "close"-callback to the `on` api)

e2e-tests/testdata/echo_client/ts/client.ts
fredr: I've updated to just proxy `on` to the websockets `addEventListener` (and also added `off`). Let me know what you think

e2e-tests/testdata/echo_client/js/client.js
```diff
@@ -552,6 +552,178 @@ function mustBeSet(field, value) {
...
+    close() {
+        this.socket.close();
+    }
+
+    async send(msg) {
+        if (this.socket.ws.readyState === WebSocket.CONNECTING) {
+            // await that the socket is opened
+            await new Promise((resolve) => {
+                const handler = () => {
+                    this.socket.ws.removeEventListener("open", handler);
```
eandre, line 636: Looks like there's an option to register the event listener only once: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once

e2e-tests/testdata/echo_client/js/client.js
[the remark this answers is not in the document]
fredr: can be replaced with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers in the future, but I think that is still a bit too bleeding edge

e2e-tests/testdata/echo_client/js/client.js
fredr: fixed

eandre (approved):

Issue #1316, opened 2024-08-10
[ERROR]: `encore eject docker <DOCKER_IMAGE_TAG>`


ImBIOS:
### What I got

```sh
󰀵  󱑍 14:49  ﱮ encore-nextjs/backend    main !?+27 via 🥟 v1.1.22 via  v20.15.1   8s 
 ➜  docker context ls                
NAME            DESCRIPTION                               DOCKER ENDPOINT                                   ERROR
default *       Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                       
desktop-linux   Docker Desktop                            unix:///home/imbios/.docker/desktop/docker.sock   

󰀵  󱑍 14:49  ﱮ encore-nextjs/backend    main !?+27 via 🥟 v1.1.22 via  v20.15.1 
 ➜  encore eject docker encore-nextjs
2:49PM INF compiling Encore application for linux/amd64
2:49PM INF saving image to local docker daemon
2:49PM ERR unable to save docker image error="error loading image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?"
```

### What I expect

- Successfully build the docker image. 

### Encore Version

encore version v1.39.3

### Suspected Root Cause

The code which thrown the error:

https://github.com/encoredev/encore/blob/f8b785d73b213e01f25bbef44bccc5928b26d7e5/cli/daemon/export/export.go#L133-L147

Which from:

https://github.com/google/go-containerregistry/blob/b8e87ed57b8042c56cc4875aa99e79906b7b9ad5/pkg/v1/daemon/write.go#L66-L70

Suspected related issues:

- https://github.com/google/go-containerregistry/issues/205
- https://github.com/google/go-containerregistry/issues/552
- https://github.com/google/go-containerregistry/issues/627

### Command

```sh
encore eject docker encore-nextjs
```

### Tried Potential Solution

1. After identified the selected docker context is `dsktop-linux` and it's mismatch with the error given by encore, I switched it to `default`, and the error still the same.
2. After that, I thought it's similar to #1309, so I restarted my Ubunut machine, but error still there.

Let me know if there any other info you need.

ImBIOS:
By the way, it's out of topic, but... @encoredev why don't you have a bug/feature GitHub issue template in place?

eandre:
[...]

PR #1317, opened 2024-08-10
test: Add Docker export test with mock builder implementation

ImBIOS:
It introduces a new file `export_test.go` with unit tests for the export functionality. The tests use a mock implementation of the `builder.Impl` interface called `MockBuilder`.

The commit also includes an update to the `go.mod` file, adding the `github.com/lib/pq` and `github.com/stretchr/testify` dependencies.

Lastly, the `go.sum` file is updated with the checksum for the `github.com/stretchr/objx` v0.5.0 dependency.

Related:
- #1316


eandre:
I believe the original root cause has been fixed, and we're about to land some large improvements to the docker export functionality, so closing this as it's about to be superceded.

PR #1320, opened 2024-08-13
grpc-connect: Add warning about using encore decorators with Connect

echarrod:
### Changed
I had some problems when using the supplied Connect example with Encore decorators. Thought I'd add a warning to advise against it.

### DIFF
--- a/docs/how-to/grpc-connect.md
+++ b/docs/how-to/grpc-connect.md
@@ -127,6 +127,15 @@ there are no Encore specific changes required here.
 
 </Callout>
 
+
+<Callout type="warning">
+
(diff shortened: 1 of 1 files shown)


eandre:
Thanks for submitting a PR! After some deliberation I think that, on balance, it's a bit too specific to call out in the docs like this, since that approach isn't something the docs are advocating for.

echarrod:
> Thanks for submitting a PR! After some deliberation I think that, on balance, it's a bit too specific to call out in the docs like this, since that approach isn't something the docs are advocating for.

Sure no worries, I was thinking other people might be in a similar situation to me, but you'll be the expert in that, so not polluting the docs makes sense 😄

PR #1322, opened 2024-08-13
tsparser: Search for package manager in workspace

fredr:
If no package manager is found in the apps `package.json`, search for a workspace `package.json` and parse it.

### DIFF
--- a/tsparser/src/builder/package_mgmt.rs
+++ b/tsparser/src/builder/package_mgmt.rs
@@ -16,12 +16,29 @@ struct PackageJson {
     dependencies: HashMap<String, String>,
 }
 
+fn parse_package_json(package_json_path: &Path) -> Result<PackageJson> {
+    let package_json = std::fs::read_to_string(package_json_path)
+        .with_context(|| format!("failed to read {}", package_json_path.display()))?;
+
+    serde_json::from_str(&package_json)
+        .with_context(|| format!("failed to parse {}", package_json_path.display()))
+}
+
 pub(super) fn resolve_package_manager(package_dir: &Path) -> Result<Box<dyn PackageManager>> {
     let package_json_path = package_dir.join("package.json");
-    let package_json = std::fs::read_to_string(&package_json_path)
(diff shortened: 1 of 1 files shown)


tsparser/src/builder/package_mgmt.rs
```diff
@@ -16,12 +16,29 @@ struct PackageJson {
...
-        .with_context(|| format!("failed to parse {}", package_json_path.display()))?;
+    let mut package_json = parse_package_json(&package_json_path)?;
+
+    // if no package manager was set, search for a workspace package.json
+    if package_json.package_manager.is_none() {
+        let mut workspace_dir = package_dir.to_path_buf();
+        while workspace_dir.pop() {
+            let package_json_path = workspace_dir.join("package.json");
+            if package_json_path.exists() {
+                package_json = parse_package_json(&package_json_path)?;
```
eandre, line 37: Is it possible that this package.json file also contains no package manager, and that we should keep looking?

fredr: I didn't think there could be more that two levels of package.json (app root + workspace root), or is it possible to have workspaces in workspaces? :thinking:

tsparser/src/builder/package_mgmt.rs
```diff
@@ -16,14 +16,38 @@ struct PackageJson {
...
+    serde_json::from_str(&package_json)
+        .with_context(|| format!("failed to parse {}", package_json_path.display()))
+}
+
+fn find_workspace_package_manager(mut dir: PathBuf) -> Result<Option<String>> {
+    while dir.pop() {
+        let package_json_path = dir.join("package.json");
+        if package_json_path.exists() {
+            let package_json = parse_package_json(&package_json_path)?;
+            if package_json.package_manager.is_some() {
```
eandre, line 32: Should we check that it's not the empty string as well?

fredr: update to continue the searching if the package manager field is empty :heavy_check_mark:

eandre (approved):

PR #1333, opened 2024-08-21
runtimes/js: Re-use runtime in workers

fredr:
Store the runtime globally, and reference to the same runtime from workers (as an alternative to #1332)

This allows api calls from workers


eandre:
Closing this for now as it's not clear this is the right path forward.

PR #1344, opened 2024-08-28
daemon: set a default log level in tests

fredr:
Set a default log level for both runtime logs and encore log to error so that tests doesn't emit unnecessary logs

Can still be overridden by setting them when running the tests:
```
ENCORE_LOG=trace ENCORE_RUNTIME_LOG=trace encore test
```

### DIFF
--- a/cli/daemon/test.go
+++ b/cli/daemon/test.go
@@ -67,12 +67,14 @@ func (s *Server) Test(req *daemonpb.TestRequest, stream daemonpb.Daemon_TestServ
 			}
 		}()
 
+		testEnv := append([]string{"ENCORE_RUNTIME_LOG=error", "ENCORE_LOG=error"}, req.Environ...)
+
 		tp := run.TestParams{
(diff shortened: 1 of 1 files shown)


cli/daemon/test.go
```diff
@@ -67,12 +67,14 @@ func (s *Server) Test(req *daemonpb.TestRequest, stream daemonpb.Daemon_TestServ
 			}
 		}()
 
+		testEnv := append([]string{"ENCORE_RUNTIME_LOG=error", "ENCORE_LOG=error"}, req.Environ...)
```
eandre, line 70: Do we need to raise the application log level to error too? I think it would be sufficient to bump it for the runtime

fredr: Changed to only set the runtime log level. It might be a bit spammy still if you log from your handlers, would be nice to only show logs on failure or something, but maybe that is configurable in vitest

eandre (approved):

PR #1346, opened 2024-08-29
tsparser: error when auth handler have unsupported fields

fredr:
Adds an error if the authHandler parameter type consist of unsupported fields types. Errors out like:
```
error: authHandler parameter type can only consist of Query and Header fields
  --> /home/fredr/projects/encore-apps/hellots/hello/hello.ts:47:3
   |
47 |   userID: string;
   |   ^^^^^^^^^^^^^^^
```


eandre (approved):
Nice!

PR #1351, opened 2024-09-01
TS Database Docs: Fix example database query for single row

tuckers-tech:
The database docs call out using `queryRow` to query a single row, but the example uses `query`.  This led to confusion when I was getting up to speed with the database section since `query` didn't work for a single row.

This PR changes the docs to use `queryRow` in the example.

Also fixes a small typo I found a couple lines below! 😄


marcuskohlberg (approved):
Amazing, thank you!

PR #1385, opened 2024-09-11
Add a rust-based supervisor

ekerfelt:
Implementation of a supervisor responsible for launching encore service processes in docker images


supervisor/src/bin/supervisor-encore.rs
```diff
@@ -0,0 +1,278 @@
...
+                services, gateways
+            )
+        })?;
+
+    // Add proc-specific environment variables
+    env.extend(binary_config.env.iter().map(|e| {
+        let parts: Vec<&str> = e.splitn(2, '=').collect();
+        (
+            parts[0].to_string(),
+            parts.get(1).map_or("".to_string(), |&v| v.to_string()),
```
fredr, line 91: tiny nit to save some characters

```suggestion
            parts.get(1).unwrap_or(&"").to_string(),
```
alternatively 
```
parts.get(1).map_or("", |v| v).to_string();
```

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -0,0 +1,278 @@
...
+        ("PORT".to_string(), port.to_string()),
+        (
+            "ENCORE_PROCESS_CONFIG".to_string(),
+            base64::engine::general_purpose::STANDARD.encode(
+                serde_json::to_string(&ProcessConfig {
+                    hosted_gateways: gateways,
+                    hosted_services: services,
+                    local_service_ports: service_ports.clone(),
+                })
+                .unwrap(),
```
fredr, line 106: Not sure in which cases this can fail, maybe there are none? but if it can fail for some reason you could return an error here instead with `.map_err(|e| <format error somehow>)?` (can be made simpler by using `anyhow`)

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -0,0 +1,278 @@
...
+    let root_token = CancellationToken::new();
+    let supervisor_token = root_token.child_token();
+
+    // Spawn a task to listen for SIGINT or SIGTERM and cancel the root token
+    tokio::spawn(async move {
+        // Wait for SIGINT or SIGTERM
+        tokio::signal::ctrl_c()
+            .await
+            .expect("Failed to listen for ctrl+c");
+        println!("Received shutdown signal. Initiating graceful shutdown...");
```
fredr, line 271: you could use `log::info!("hello")` here to use the logger instead of just writing to stdout

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -0,0 +1,278 @@
...
+        root_token.cancel();
+    });
+
+    // Run the supervisor
+    sv.supervise(supervisor_token).await;
+    println!("All processes have exited. Shutting down.");
```
fredr, line 277: `log::info!` could be used here as well

fredr (approved):
Looks good! Left a few nits that you can decide on

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -0,0 +1,278 @@
...
+        ("PORT".to_string(), port.to_string()),
+        (
+            "ENCORE_PROCESS_CONFIG".to_string(),
+            base64::engine::general_purpose::STANDARD.encode(
+                serde_json::to_string(&ProcessConfig {
+                    hosted_gateways: gateways,
+                    hosted_services: services,
+                    local_service_ports: service_ports.clone(),
+                })
+                .map_err("Failed to serialize ProcessConfig"),
```
fredr, line 106:
```suggestion
                .map_err(|e| format!("Failed to serialize ProcessConfig: {}", e))?,
```
Should be more correct, map_err is like map, but you map the error of a result, so the parameter is a closure with the error as its paramter (and then add the `?` at the end to return if there was an error)

fredr (approved):
:ship:

PR #1393, opened 2024-09-13
runtimes/core: add string mapping for uuid fields

fredr:


runtimes/core/src/sqldb/val.rs
```diff
@@ -60,6 +62,10 @@ impl ToSql for RowValue {
                                 .map_err(Box::new)?;
                             val.to_sql(ty, out)
                         }
+                        Type::UUID => {
+                            let val = Uuid::parse_str(str).context("couldn't parse uuid")?;
```
eandre, line 66:
```suggestion
                            let val = Uuid::parse_str(str).context("unable to parse uuid")?;
```

eandre (approved):

PR #1394, opened 2024-09-14
 docs: update service name to a valid literal

AmanNegi:
## Changes

- Previously used `-` throws an error when the encore generates files as it acts as an import string.
- Changed to `_` which ensures that no error occurs when generating the file

## Screenshots

image
> Here it's prevalent that due to adding the `-` which is an invalid token while naming a variable we are facing an error.


eandre:
Thanks for the report! A better fix is to change the code generation to handle this :) that's now done in #1403, due to be released today or tomorrow.

PR #1396, opened 2024-09-16
supervisor: Add option to launch encore services with supervisor

ekerfelt:
This PR adds a new build config parameter to launch cohosted services as separate processes using the encore process supervisor.


cli/cmd/encore/eject.go
```diff
@@ -34,8 +34,20 @@ func init() {
 		},
 	}
 
+	lang, err := appfile.AppLang(p.AppRoot)
+	if err != nil {
+		lang = appfile.LangGo
+	}
+	p.BaseImg = "scratch"
+	if lang == appfile.LangTS {
+		p.BaseImg = "node"
```
eandre, line 43: Should this be `node:latest`?

cli/cmd/encore/eject.go
```diff
@@ -34,8 +34,20 @@ func init() {
...
+	}
+
 	dockerEjectCmd.Flags().BoolVarP(&p.Push, "push", "p", false, "push image to remote repository")
-	dockerEjectCmd.Flags().StringVar(&p.BaseImg, "base", "scratch", "base image to build from")
+	dockerEjectCmd.Flags().StringVar(&p.BaseImg, "base", p.BaseImg, "base image to build from")
+	dockerEjectCmd.Flags().StringVar(&p.Goos, "os", p.Goos, "target operating system. One of (linux, darwin, windows)")
+	dockerEjectCmd.Flags().StringVar(&p.Goarch, "arch", p.Goarch, "target architecture. One of (amd64, arm64)")
+	dockerEjectCmd.Flags().BoolVar(&p.CgoEnabled, "cgo", p.CgoEnabled, "enable cgo")
```
eandre, line 50: Should we read this from `encore.app` instead? I guess it's a bit annoying to do so :)

cli/daemon/export/download.go
```diff
@@ -0,0 +1,100 @@
...
+			oldVersionPath := filepath.Join(binDir.String(), entry.Name())
+			if err := os.RemoveAll(oldVersionPath); err != nil {
+				log.Warn().Msgf("failed to remove old version directory %s: %v", oldVersionPath, err)
+			}
+		}
+	}
+	return
+}
+
+func downloadFile(url, dest string) error {
```
eandre, line 78: Should we publish checksums alongside to verify it was successfully downloaded completely and correctly? Or rely on TCP for that for now?

runtimes/core/src/lib.rs
```diff
@@ -52,8 +53,61 @@ pub mod encore {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProcessConfig {
```
eandre, line 57: Maybe move this out of `lib.rs` into a separate file. Can use `pub use` to re-export from here

runtimes/core/src/lib.rs
```diff
@@ -52,8 +53,61 @@ pub mod encore {
...
+            .iter()
+            .map(|s| runtimepb::HostedService { name: s.clone() })
+            .collect();
+        deployment.hosted_gateways = self
+            .hosted_gateways
+            .iter()
+            .map(|s| {
+                cfg.infra
+                    .as_ref()
+                    .expect("infra not found in runtime config")
```
eandre, line 78: Do we want `unwrap_or_default()` instead to fall back to the default value if it's missing?

cli/cmd/encore/eject.go
ekerfelt: 👍  Latest is the default tag, but probably good to be explicit

cli/cmd/encore/eject.go
ekerfelt: Ah, for cgo that makes sense of course. I already read encore.app to find the language (to select default base image). Would you want to be able to specify arch and os in the encore.app too?

cli/daemon/export/download.go
ekerfelt: yeah. I guess that might save us some headaches

cli/cmd/encore/eject.go
eandre: Nah, arch/os seems more like the concern of a specific invocation of building an image, whereas using cgo is more a property of the app's requirements

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -223,41 +20,47 @@ pub async fn main() {
...
-                &supervisor_config,
+                &supervisor_config.binary_config,
             )
-            .expect("Failed to create process for service"),
+            .expect("Failed to create process for gateways"),
         );
     }
 
+    let proxy = proxy::GatewayProxy::new(
+        std::net::SocketAddr::new("127.0.0.1".parse().unwrap(), port),
```
eandre, line 61:
```suggestion
        std::net::SocketAddr::new(std::net::Ipv4Addr::new(127, 0, 0, 1), port)
```

supervisor/src/config.rs
```diff
@@ -0,0 +1,232 @@
...
+        .expect("Failed to read supervisor_config.json");
+    serde_json::from_str(&contents).expect("Failed to parse SupervisorConfig JSON")
+}
+
+// attempts to read the encore runtime config either as proto or json. Extracts and returns the
+// hosted services and gateways.
+fn load_hosted_processes() -> (Vec<String>, Vec<String>) {
+    // Read and decode the runtime config bytes from the environment variable
+    let runtime_config = env::var("ENCORE_RUNTIME_CONFIG")
+        .map_err(|e| format!("Failed to read ENCORE_RUNTIME_CONFIG: {}", e))
```
eandre, line 51: This would be better handled with `.context()` from the `anyhow` crate. And then make this function return `anyhow::Result<(Vec<String>, Vec<String>)>` and don't use `.expect()` but rather `?`

eandre (dismissed):

eandre (dismissed):

runtimes/core/src/lib.rs
```diff
@@ -266,6 +282,41 @@ impl Runtime {
...
+                let svc_name = (|| -> Result<String, anyhow::Error> {
+                    Ok(md
+                        .pubsub_topics
+                        .iter()
+                        .find(|t| t.name == s.topic_encore_name)
+                        .context("could not find topic")?
+                        .subscriptions
+                        .iter()
+                        .find(|ms| ms.name == s.subscription_encore_name)
+                        .context("could not find sub")?
```
eandre, line 301: Should we fail runtime startup in all these cases? Or is it better to fail when handling the push?

runtimes/core/src/api/gateway/mod.rs
```diff
@@ -192,20 +196,37 @@ impl ProxyHttp for Gateway {
...
-            .req_header()
-            .method
-            .as_ref()
-            .try_into()
-            .or_err(ErrorType::HTTPStatus(400), "invalid http method")?;
-
-        let service_name = self.inner.router.route_to_service(method, path)?;
+        let service_name = push_proxy_svc
+            .map(Ok)
+            .or_else(|| {
```
eandre, line 211: I think this can be `map_or_else(|| { ... }, Ok)` instead. Or a match statement, which might be clearer?

runtimes/core/src/api/gateway/mod.rs
```diff
@@ -192,20 +196,37 @@ impl ProxyHttp for Gateway {
...
+                        .and_then(|method| {
+                            self.inner
+                                .router
+                                .route_to_service(method, path)
+                                .context("couldn't find upstream")
+                        }),
+                )
+            })
+            .or_err(ErrorType::InternalError, "couldn't find upstream")?
+            .or_err(ErrorType::InternalError, "couldn't find upstream")?;
```
eandre, line 229: Huh?

runtimes/go/appruntime/apisdk/api/pubsub_push_remote.go
```diff
@@ -0,0 +1,76 @@
...
+type remotePubSubPushHandler struct {
+	server         *Server        // The server we're running in
+	hostingService config.Service // The service name of the remote subscription handler
+	pushURL        string         // The URL of the remote subscription handler
+	logger         zerolog.Logger
+}
+
+func (r *remotePubSubPushHandler) ForwardRequest(w http.ResponseWriter, req *http.Request) error {
+	// Create the remote push request
+	pushReq, err := http.NewRequestWithContext(req.Context(), req.Method, r.pushURL, req.Body)
```
eandre, line 28: Why not use `httputil.ReverseProxy`? It handles a lot more edge cases than this

runtimes/core/src/lib.rs
ekerfelt: Figured it would be a case of broken runtime config when there's a subscription defined which is not available in the metadata, so opted for failing. But I'm easily convinced so I'll change it :)

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -58,16 +59,27 @@ pub async fn main() {
...
     }
 
-    let proxy = proxy::GatewayProxy::new(
-        reqwest::Client::new(),
-        std::net::SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), port),
-        service_ports.clone(),
-    );
-    let sv = Supervisor::new(procs);
+    let mut handles = Vec::with_capacity(2);
+    let mut results = Vec::with_capacity(handles.len());
```
eandre, line 66: Isn't `handles.len()` always zero here? The capacity is 2 but length is 0. (Also seems like a highly unnecessary optimization ;))

supervisor/src/bin/supervisor-encore.rs
```diff
@@ -58,16 +59,27 @@ pub async fn main() {
...
     let root_token = CancellationToken::new();
+
+    let sv = Supervisor::new(procs);
     let supervisor_token = root_token.child_token();
+    handles.push(tokio::spawn(sv.supervise(supervisor_token)));
+
+    if use_proxy {
+        let proxy = proxy::GatewayProxy::new(
+            reqwest::Client::new(),
+            std::net::SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), port - 1),
```
eandre, line 76: Why `port-1`?

supervisor/src/bin/supervisor-encore.rs
ekerfelt: Haha, yeah. I might have had some extraterrestial help

supervisor/src/bin/supervisor-encore.rs
ekerfelt: The port is an incremental value increased after each assignment.

eandre: But isn't that port assigned to some other proc?

ekerfelt: Exactly. It's the proxy target. I.e. the port the proxy will forward all requests to. The serving port is assigned in the serve method

eandre (approved):

PR #1401, opened 2024-09-16
tsparser: fall back to junction points on windows

eandre:
Windows requires Developer Mode to be enabled to use
symlinks without admin permissions. This isn't always enabled,
so fall back to using junction points if symlink creation fails.

This is for example what rustup does already.
See rust-lang/rustup#3687 for example.

### DIFF
--- a/tsparser/src/builder/package_mgmt.rs
+++ b/tsparser/src/builder/package_mgmt.rs
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 use anyhow::{Context, Result};
 use duct::cmd;
 use serde::Deserialize;
+use std::io;
 
 use crate::builder::compile::CmdSpec;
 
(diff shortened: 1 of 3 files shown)


fredr (approved):
Nice!

tsparser/src/builder/package_mgmt.rs
```diff
@@ -304,9 +303,58 @@ fn symlink_encore_dev(app_root: &Path, encore_dev_path: &Path) -> Result<()> {
...
+fn read_symlink(src: &Path) -> io::Result<Option<PathBuf>> {
+    if let Ok(meta) = src.symlink_metadata() {
+        // Is this a symlink?
+        if meta.is_symlink() {
+            return std::fs::read_link(src).map(Some);
+        }
+    }
+
+    // Check if it's a junction.
+    if let Ok(target) = junction::exists(src) {
```
fredr, line 342:
```suggestion
    if let Ok(_target) = junction::exists(src) {
```
target isnt used

PR #1406, opened 2024-09-17
Tweaks to the streaming docs

simon-johansson:


docs/ts/primitives/streaming-apis.mdx
```diff
@@ -8,118 +8,269 @@ lang: ts
...
+
 ## Different kinds of stream
 
 Encore supports three types of streams, each designed for a specific data flow direction:
-- **StreamIn**: When you need to stream data into your service.
-- **StreamOut**: When you need to stream data out from your service.
-- **StreamInOut**: When you need to stream data into and out of service.
+- [**StreamIn**](#streamin): When you need to stream data into your service.
+- [**StreamOut**](#streamout): When you need to stream data out from your service.
+- [**StreamInOut**](#streaminout): When you need to stream data into and out of service.
```
fredr, line 21:
```suggestion
- [**StreamInOut**](#streaminout): When you need to stream data into and out of your service.
```
saw that I had missed adding a word here

docs/ts/primitives/streaming-apis.mdx
```diff
@@ -8,118 +8,269 @@ lang: ts
...
+// Returned when the stream is done, optional.
+interface Response {
+  success: boolean;
+}
+
+export const uploadStream = api.streamIn<
+  Handshake,
+  Message,
+  Response
+>({ path: "/upload", expose: true }, async (handshake, stream) => {
```
fredr, line 67:
```suggestion
>({ path: "/upload", expose: true }, async (handshake, stream): Promise<Response> => {
```
I think you can add the return type here

fredr (dismissed):
Really nice! Added a nit and a unrelated typo

Also I saw that you are using `console.log` and `console.error`, maybe it is better if we use `log.error()` and `log.info()` from `import log from "encore.dev/log";`?

docs/ts/primitives/streaming-apis.mdx
eandre: It's not needed since the generic parameters are passed in (`api.streamIn<A, B, C>`), so I think it's cleaner this way

docs/ts/primitives/streaming-apis.mdx
```diff
@@ -8,118 +8,269 @@ lang: ts
...
 
-Or if you want to have a stream from client to server, you can set it up with `api.streamIn`:
+```ts
+api.streamIn<Handshake, Incoming>(
+  {...}, async (handshake, stream) => {...})
+```
+
+```ts
+api.streamIn<Incoming, Outgoing>(
+  {...}, async (stream): Promise<Outgoing> => {...})
```
eandre, line 100: I guess we don't need the promise here either

simon-johansson: I thought it was clearer when it was explicit, no?

eandre: Seems like extra typing to me :) in most TypeScript code I've seen people prefer to use type inference when possible. But I don't feel strongly about it, keep it if you prefer

docs/ts/primitives/streaming-apis.mdx
simon-johansson: Thanks! I have pushed a new commit

fredr (approved):
:ship:

PR #1411, opened 2024-09-23
clientgen: handle optional auth params

### LINKED ISSUE #1131: Generated client has type error when using optional `AuthParams`
I have a system that authenticates via secure cookie _or_ api token (depending on if the client is the browser or CLI).  I currently have this set up on the encore side with:
```ts
interface AuthParams {
  cookie?: Header<'Cookie'>
  token?: Header<'x-api-token'>
}
```

This works great on the backend side, I didn't run into any issues until I started integrating the client.  I can generate the client but when I try and build (exec TS check) I'm running into an issue with the following generated lines in the `callAPI` method
```ts
if (authData) {
  init.headers['cookie'] = authData.cookie
  init.headers['x-api-token'] = authData.token
}
```

This throws the following error because both `authData.cookie` and `authData.token` are optional
```ts
Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

eandre:
Fixes #1131

### DIFF
--- a/e2e-tests/testdata/echo_client/js/client.js
+++ b/e2e-tests/testdata/echo_client/js/client.js
@@ -766,16 +766,19 @@ class BaseClient {
         if (authData) {
             const data = {};
 
-            data.query = {};
-            data.query["query"] = authData.Query.map((v) => String(v));
-            data.query["new-auth"] = String(authData.NewAuth);
-            data.headers = {};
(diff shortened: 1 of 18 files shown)


fredr (approved):
Looks good!

fredr (approved):

PR #1417, opened 2024-09-25
cli: Add alpha support for triggering deploys from cli

ekerfelt:


cli/cmd/encore/deploy.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				cmdutil.Fatalf("no app found. Run deploy inside an encore app directory or specify the app with --app")
+			}
+			appSlug, err = appfile.Slug(appRoot)
+			if err != nil {
+				cmdutil.Fatalf("no app found. Run deploy inside an encore app directory or specify the app with --app")
+			}
+		}
+		if envName == "" {
+			envName = "@primary"
```
eandre, line 49: Is this a good idea? Feels a bit risky to not be explicit about which environment to deploy to.

cli/cmd/encore/deploy.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+var deployAppCmd = &cobra.Command{
+	Use:                   "deploy --commit COMMIT_SHA | --branch BRANCH_NAME",
+	Short:                 "Deploy an Encore app to a cloud environment",
+	DisableFlagsInUseLine: true,
+	Hidden:                true,
+	Run: func(c *cobra.Command, args []string) {
+		if commit != "" {
+			hb, err := hex.DecodeString(commit)
```
eandre, line 33: Thoughts on using `git rev-parse` here? Even if we require the commit to be pushed it still seems like a nice convenience. Maybe not worth it?

cli/cmd/encore/deploy.go
```diff
@@ -0,0 +1,101 @@
...
+	},
+}
+
+func init() {
+	alphaCmd.AddCommand(deployAppCmd)
+	deployAppCmd.Flags().StringVar(&appSlug, "app", "", "app slug to deploy to (default current app)")
+	deployAppCmd.Flags().StringVarP(&envName, "env", "e", "", "environment to deploy to (default primary env)")
+	deployAppCmd.Flags().StringVarP(&commit, "commit", "c", "", "commit to deploy")
+	deployAppCmd.Flags().StringVarP(&branch, "branch", "b", "", "branch to deploy")
+	deployAppCmd.Flags().StringVarP(&format, "format", "f", "text", "format of output. One of (text, json)")
```
eandre, line 98: We have `cmdutil.Oneof` for this use case which does some niceties like autocomplete and validation

cli/cmd/encore/deploy.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	},
+}
+
+func init() {
+	alphaCmd.AddCommand(deployAppCmd)
+	deployAppCmd.Flags().StringVar(&appSlug, "app", "", "app slug to deploy to (default current app)")
+	deployAppCmd.Flags().StringVarP(&envName, "env", "e", "", "environment to deploy to (default primary env)")
+	deployAppCmd.Flags().StringVarP(&commit, "commit", "c", "", "commit to deploy")
+	deployAppCmd.Flags().StringVarP(&branch, "branch", "b", "", "branch to deploy")
```
eandre, line 97: The shorthands seem a bit esoteric; consider removing?

cli/cmd/encore/deploy.go
ekerfelt: Yeah, you're probably right. Was striving to be consistent with other commands.

cli/cmd/encore/deploy.go
ekerfelt: Yeah, maybe we should. Can't assume that git is installed or the command is run from within a local repository. But I suppose I could run it if --app is not set and git is available. On fail, I could parse the the revision specifier manually and submit it as a commit sha if it's a 20 byte hex, or a branch if it's not.

ekerfelt: Probably rename the flag to revision then?

eandre: I guess that's fair, let's just keep it as-is and not spend more time on it

eandre (approved):

PR #1419, opened 2024-09-25
Only include non-private endpoints in OpenAPI spec output

nitsanshai:
Currently, the OpenAPI spec includes private endpoints, which aren't accessible outside of the server. This change removes private endpoints from the spec.


eandre (approved):
Thanks!

PR #1431, opened 2024-09-26
Support filtering by tags in client generation

nitsanshai:
This change enables filtering by tags when generating API clients. If `tags` is provided, only endpoints with the provided tags will be included in the output. If `excluded-tags` is provided, endpoints that do not have the provided tags will be included. If both are provided, only endpoints that have one of the tags in the `tags` list and no tags in the `excluded-tags` list will be included. `excluded-tags` takes priority over the `tags` list.


internal/clientgen/clientgentypes/clientgentypes.go
```diff
@@ -63,3 +64,53 @@ func NewServiceSet(md *meta.Data, include, exclude []string) ServiceSet {
 func AllServices(md *meta.Data) ServiceSet {
 	return NewServiceSet(md, []string{"*"}, nil)
 }
+
+type TagSet struct {
+	includedTags []string
+	excludedTags []string
```
eandre, line 70: Make these `map[string]bool` to avoid running into O(N^2) performance in the `IsRPCIncluded` method

eandre (commented):
Nice! Just one nitpick.

internal/clientgen/clientgentypes/clientgentypes.go
nitsanshai: Done

cli/cmd/encore/gen.go
```diff
@@ -26,10 +26,11 @@ func init() {
 		envName          string
 		genServiceNames  []string
 		excludedServices []string
+		endpointTags     map[string]string
 	)
 
 	genClientCmd := &cobra.Command{
-		Use:   "client [<app-id>] [--env=<name>] [--services=foo,bar] [--excluded-services=baz,qux]",
+		Use:   "client [<app-id>] [--env=<name>] [--services=foo,bar] [--excluded-services=baz,qux] [--tags=cache=true,internal=false]",
```
eandre, line 33: I liked the previous semantics better; seems more consistent with `--services` and `--excluded-services`. Can you revert this and just keep the `map[string]bool` as an implementation detail of `clientgentypes.TagSet`?

nitsanshai: Gotcha, done

internal/clientgen/clientgentypes/clientgentypes.go
```diff
@@ -63,3 +64,56 @@ func NewServiceSet(md *meta.Data, include, exclude []string) ServiceSet {
 func AllServices(md *meta.Data) ServiceSet {
 	return NewServiceSet(md, []string{"*"}, nil)
 }
+
+type TagSet struct {
+	tagMap       map[string]bool
+	includedTags []string
```
eandre, line 70: I'd much rather model this as two separate maps:

```go
included map[string]bool
excluded map[string]bool
```

rather than having a single map represent both and then a separate list for the included tags. Seems much easier to reason about to me

nitsanshai: No problem, done

internal/clientgen/clientgentypes/clientgentypes.go
```diff
@@ -63,3 +64,56 @@ func NewServiceSet(md *meta.Data, include, exclude []string) ServiceSet {
...
+}
+
+func (t TagSet) IsRPCIncluded(rpc *meta.RPC) bool {
+	// First check if the RPC has any of the excluded tags.
+	for _, selector := range rpc.Tags {
+		if selector.Type != meta.Selector_TAG {
+			continue
+		}
+
+		if excluded, ok := t.excluded[selector.Value]; ok && excluded {
```
eandre, line 96:
```suggestion
		if t.excluded[selector.Value] {
```

internal/clientgen/clientgentypes/clientgentypes.go
```diff
@@ -63,3 +64,56 @@ func NewServiceSet(md *meta.Data, include, exclude []string) ServiceSet {
...
+		return true
+	}
+
+	// Check if the RPC has any of the included tags.
+	for _, selector := range rpc.Tags {
+		if selector.Type != meta.Selector_TAG {
+			continue
+		}
+
+		if included, ok := t.included[selector.Value]; ok && included {
```
eandre, line 112:
```suggestion
		if t.included[selector.Value] {
```

eandre (commented):
One more nitpick, sorry :)

nitsanshai:
@eandre could you take another look? Thanks!

proto/encore/daemon/daemon_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.3.0
-// - protoc             v4.23.4
+// - protoc-gen-go-grpc v1.5.1
+// - protoc             v5.28.1
```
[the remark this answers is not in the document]
nitsanshai: Sorry, almost right. I'll try again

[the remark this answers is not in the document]
nitsanshai: Okay, I believe this is correct now

nitsanshai:
@eandre if you could look again when you get a chance, thanks!

eandre (approved):
Thanks!

PR #1463, opened 2024-10-09
tsparser: better endpoint parser errors

fredr:
Report errors with span so its easier to understand them

### DIFF
--- a/tsparser/src/parser/resources/apis/encoding.rs
+++ b/tsparser/src/parser/resources/apis/encoding.rs
@@ -291,7 +291,7 @@ fn describe_req(
                 None,
             ));
         } else {
-            anyhow::bail!("missing request schema");
+            anyhow::bail!("request schema must be defined when having path parameters");
         }
     };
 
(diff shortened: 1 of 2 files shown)


tsparser/src/parser/resources/apis/encoding.rs
```diff
@@ -291,7 +291,7 @@ fn describe_req(
                 None,
             ));
         } else {
-            anyhow::bail!("missing request schema");
+            anyhow::bail!("request schema must be defined when having path parameters");
```
eandre, line 294: I think we should also ensure all path params are present in the schema, unless that's very hard

fredr: We do that, but thats a different error, looks like this:
```
error: path parameter "hello" not found in request schema
  --> /home/fredr/projects/encore-apps/teeeest/repro/repro.ts:29:27
   |
29 |   export const withParams = api(
   |  ___________________________^
30 | |   { expose: true, method: "POST", path: "/some/:hello" },
31 | |   async (req: Req) => {},
32 | | );
   | |_^
```

eandre: Ah great!

eandre (approved):

PR #1485, opened 2024-10-15
runtimes/core: propagate auth error from auth handler

fredr:
Propagate auth errors from the auth handler


fredr:
This also fixes a bug when we convert pingora errors to api errors in some cases :lady_beetle:

eandre (approved):

PR #1486, opened 2024-10-15
cli: Add role support for db shell

ekerfelt:


cli/cmd/encore/db.go
```diff
@@ -28,9 +29,31 @@ var (
 	resetAll bool
 	testDB   bool
 	shadowDB bool
-	nsName   string
+	role     = cmdutil.Oneof{
+		Value:   "reader",
+		Allowed: []string{"reader", "writer", "admin", "superuser"},
```
eandre, line 34: I think read/write/admin/superuser is better than reader/writer

eandre: But maybe `--write` etc instead of `--role write` etc?

ekerfelt: problem with --write is what to do with combinations. Seems cleaner to have a role?

eandre (approved):

fredr (approved):
:tophat:

fredr (approved):
:ship:

PR #1491, opened 2024-10-16
runtimes/core+js: add details to errors

fredr:


runtimes/js/encore.dev/api/error.ts
```diff
@@ -112,6 +114,8 @@ export class APIError extends Error {
   }
 }
 
+export type ErrDetails = object;
```
eandre, line 117: I think this should be `Record<string, any>`

runtimes/js/src/error.rs
```diff
@@ -0,0 +1,72 @@
+use crate::log::parse_js_stack;
+use encore_runtime_core::api;
+use napi::{Env, JsUnknown, Status};
+
+pub fn coerce_to_api_error(env: Env, val: napi::JsUnknown) -> Result<api::Error, api::Error> {
```
eandre, line 5: Should the return type just be `api::Error`? Is the Ok/Err distinction useful?

fredr: We can change it to that, its just nice to be able to know if the coerce succeeded or not imo, even though both success and fail is the same type.

As we use it today we dont care if it succeeded or not, so I dont have strong opinions about it

eandre: Fair, can keep it as-is

eandre (approved):

PR #1504, opened 2024-10-18
date handling

eandre:
- runtimes/{core,js}: refactor schema encode/decode
- Add date support


runtimes/core/src/pubsub/mod.rs
```diff
@@ -24,7 +24,7 @@ pub type MessageId = String;
 
 pub struct MessageData {
     pub attrs: HashMap<String, String>,
-    pub body: Option<serde_json::Value>,
+    // pub body: Option<serde_json::Value>,
```
fredr, line 27: delete this? or is it going to be re-added at some point?

fredr (dismissed):
looks great!

runtimes/js/src/pvalue.rs
```diff
@@ -0,0 +1,272 @@
...
+            let pv = unsafe { std::mem::transmute::<&PValue, &PVal>(v) };
+            obj.set(k, pv)?;
+        }
+
+        unsafe { JsObject::to_napi_value(raw_env, obj) }
+    }
+}
+
+impl FromNapiValue for PVals {
+    unsafe fn from_napi_value(env: sys::napi_env, napi_val: sys::napi_value) -> Result<Self> {
```
fredr, line 209: this code panics if the value is `undefined`. I had similar problems when parsing the error details, and had to parse it to JsUnknown first and then match on the type to check that it was an JsObject before coercing into that.

fredr (approved):
:clock1:

PR #1508, opened 2024-10-21
Improve self hosting documentation

ekerfelt:
* Improves structure of self hosting documentation
* Adds a guide for how to deploy to Digital Ocean using a docker image


docs/menu.cue
```diff
@@ -745,6 +746,28 @@
...
+	items: [
+		{
+			kind: "basic"
+			text: "Build Docker Images"
+			path: "/how-to/self-host"
+			file: "how-to/self-host"
+		}, {
+			kind: "basic"
+			text: "Configure Infrastructure"
+			path: "/selfhosting/configure-infra"
```
eandre, line 761: `/self-host/...` maybe is nicer?

docs/menu.cue
```diff
@@ -745,6 +746,28 @@
...
 }
 
+#SelfHostingSection: #SectionMenuItem & {
+	kind: "section"
+	text: "Self Hosting"
+	items: [
+		{
+			kind: "basic"
+			text: "Build Docker Images"
+			path: "/how-to/self-host"
```
eandre, line 756: Should we make this `/self-host/docker-build` or something? And redirect the old path?

docs/selfhosting/configure-infra.md
```diff
@@ -0,0 +1,429 @@
...
+If you are using infrastructure resources, such as SQL databases, Pub/Sub, or metrics, you will need to configure your Docker image with the necessary configuration.
+The `build` command lets you provide this by specifying a path to a config file using the `--config` flag.
+
+```bash
+encore build docker --config path/to/infra-config.json MY-IMAGE:TAG
+```
+
+The configuration file should be a JSON file using the [Encore Infra Config](https://encore.dev/schemas/infra.schema.json) schema.
+
+This includes configuring things like:
```
eandre, line 16: Maybe "supports" is better to indicate not everything is required?

```suggestion
This supports configuring things like:
```

docs/selfhosting/configure-infra.md
```diff
@@ -0,0 +1,429 @@
...
+        "my-db": {
+          "username": "my-db-owner",
+          "password": {"$env": "DB_PASSWORD"}
+        }
+      }
+    }
+  ],
+  "service_discovery": {
+    "myservice": {
+      "base_url": "https://my-service:8044"
```
eandre, line 54: Probably better to use the same hostname as the service name, since that's the most common setup?

```suggestion
      "base_url": "https://myservice:8044"
```

docs/selfhosting/configure-infra.md
```diff
@@ -0,0 +1,429 @@
...
+  }
+}
+```
+
+- `total`: The total time allowed for the shutdown process in seconds.
+- `shutdown_hooks`: The time allowed for executing shutdown hooks.
+- `handlers`: The time allocated for processing request handlers during the shutdown.
+
+### 3. Authentication Methods Configuration
+Private endpoints will not require authentication if no authentication methods are specified. To secure private endpoints, you can configure authentication methods.
```
eandre, line 145:
```suggestion
Private endpoints will not require authentication if no authentication methods are specified. This is typically fine when services are deployed on a private network such as a VPC. But sometimes you might need to connect to other services over the public internet, in which case you'll want to ensure private endpoints are only accessible to other backend services. To do that you can configure authentication methods.
```

docs/selfhosting/deploy-to-digital-ocean.md
```diff
@@ -0,0 +1,196 @@
...
+   - Open your Encore app’s codebase.
+   - Add `mydb` database to your app ([Encore Database Documentation](https://encore.dev/docs/ts/primitives/databases))
+   ```typescript
+      const mydb = new SQLDatabase("mydb", {
+         migrations: "./migrations",
+      });
+
+      export const getUser = api(
+        { expose: true, method: "GET", path: "/names/:id" },
+        async ({id}:{id:number}): Promise<{ id: number; name: string }> => {
```
eandre, line 138:
```suggestion
        async ({id}: {id:number}): Promise<{ id: number; name: string }> => {
```

eandre (approved):

PR #1511, opened 2024-10-22
tsparser: improve auth handler errors

fredr:


tsparser/src/legacymeta/mod.rs
```diff
@@ -275,9 +275,12 @@ impl<'a> MetaBuilder<'a> {
 
                 Dependent::Gateway((_b, gw)) => {
                     let auth_handler = if let Some(auth_handler) = &gw.auth_handler {
-                        let ah = auth_handlers
-                            .get(&auth_handler.id)
-                            .ok_or(anyhow::anyhow!("auth handler not found"))?;
+                        let Some(ah) = auth_handlers.get(&auth_handler.id) else {
+                            HANDLER.with(|handler| {
+                                handler.span_err(gw.range, "auth handler not found")
+                            });
```
eandre, line 281: `gw.range` is basically a wrapper around `Span`. We could implement `gw.range.err()` with the same trait we already have in `span_err` probably, or convert it to use a `Span` and start moving away from the `Range` type. But this is also fine for now, not critical to change it

fredr: I can add that implementation for range for now, always good with some convenience

eandre (approved):

PR #1516, opened 2024-10-23
propagate errors details between service calls

fredr:


runtimes/js/src/runtime.rs
```diff
@@ -242,4 +243,5 @@ impl Runtime {
 pub struct APICallError {
     pub code: String,
     pub message: String,
+    pub details: Option<serde_json::Map<String, serde_json::Value>>,
```
eandre, line 246: I guess we should wire this up with support for `Date` when that lands, too

eandre (approved):

PR #1542, opened 2024-11-01
runtimes/js: service middlewares

fredr:
Adds middlewares for services, example usage (`encore.service.ts`):

```ts
import { Service } from "encore.dev/service";
import { middleware } from "encore.dev/api";
import log from "encore.dev/log";

const myMiddleware = middleware(
  { target: { isRaw: false, isStream: false } },
  async (req, next) => {
    let resp = await next(req);
    resp.header.set("x-my-extra-header", ["value1", "value2", "value3"]);
    resp.header.set("x-other", "single value");
    return resp;
  },
);

const forAllEps = middleware(async (req, next) => {
  log.info("one middleware for all endpoints!");
  return await next(req);
});

export default new Service("hello", {
  middlewares: [
    myMiddleware,
    forAllEps,
  ],
});

```


runtimes/js/encore.dev/req_meta.ts
```diff
@@ -10,6 +10,12 @@ export interface APIDesc {
 
   /** Whether the endpoint is a raw endpoint. */
   raw: boolean;
+
+  /** Whether the endpoint requires auth. */
+  requiresAuth: boolean;
+
+  /** Whether the endpoint is a is exposed. */
+  exposed: boolean;
```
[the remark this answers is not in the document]
fredr: ah makes sense, I added these here for the previous implementation, so they are actually not needed, just keep them as it could be nice to have access to I thought. I'll remove this one, but should I keep auth/requiresAuth? or remove that as well?

runtimes/js/encore.dev/req_meta.ts
```diff
@@ -10,6 +10,12 @@ export interface APIDesc {
 
   /** Whether the endpoint is a raw endpoint. */
   raw: boolean;
+
+  /** Whether the endpoint requires auth. */
+  requiresAuth: boolean;
```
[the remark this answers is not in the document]
fredr: it is named requiresAuth in the meta: https://github.com/encoredev/encore/pull/1542/files#diff-7cf090c6d18624ee4515f89f3812b04dfe21d1b526434eccbd432f8b945ecad6R172

but I have no strong opinions, just that it was a bit more clear. I can rename it, or remove it for now, we dont actually need it here for this implementation

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -261,5 +265,113 @@ api.static = function staticAssets(options: StaticOptions) {
...
+    this._stream = stream;
+    this._rawReq = rawReq;
+    this._rawResp = rawResp;
+  }
+
+  /**
+   * requestMeta is set when the handler is a typed handler or a stream handler.
+   * for raw handlers, see rawRequest and rawResponse.
+   */
+  public get requestMeta() {
```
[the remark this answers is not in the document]
fredr: sure! they do get the type from the private definitions above, bit I'll add it here as well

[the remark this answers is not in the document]
eandre: Yeah, just easier for humans to parse when it's more explicit

runtimes/js/src/api.rs
```diff
@@ -87,15 +88,35 @@ impl Request {
...
+        let extra_headers = obj
+            .get_named_property::<napi::JsUnknown>("extraHeaders")
+            .map_err(|e| {
+                api::Error::invalid_argument("extraHeaders missing in handler response", e)
+            })?;
+
+        match parse_pvalues(payload) {
+            Ok(val) => Ok(HandlerResponseInner {
+                payload: val,
+                extra_headers: parse_header_map(extra_headers).unwrap(),
```
fredr, line 118: TODO: handle unwrap

runtimes/js/src/api.rs
```diff
@@ -87,15 +88,35 @@ impl Request {
...
+
+        let payload = obj
+            .get_named_property::<napi::JsUnknown>("payload")
+            .map_err(|e| api::Error::invalid_argument("payload missing in handler response", e))?;
+
+        let extra_headers = obj
+            .get_named_property::<napi::JsUnknown>("extraHeaders")
+            .map_err(|e| {
+                api::Error::invalid_argument("extraHeaders missing in handler response", e)
+            })?;
```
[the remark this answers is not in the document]
fredr: I'm not sure how we would know that here, since the middlewares are executed in js land? :thinking:

tsparser/src/builder/codegen.rs
```diff
@@ -159,6 +163,18 @@ impl Builder<'_> {
                 let mut endpoint_ctx = Vec::new();
                 let mut subscription_ctx = Vec::new();
 
+                let service_ctx = service.map(|service| {
```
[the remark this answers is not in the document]
fredr: I was assuming that we only wanted to support middewares from encore.service.ts? was that wrong? the `service_ctx` is only used so that we can import the service in the handlebars templates, so that we can access the middlewares from there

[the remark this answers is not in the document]
eandre: Yeah we will only have middleware when there is a service object in `encore.service.ts`. But I was just thinking logically there is still a service, even if there is no `encore.service.ts`. But can change that later if we need it

runtimes/js/encore.dev/req_meta.ts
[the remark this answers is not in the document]
eandre: Ah that's fine then :)

runtimes/js/encore.dev/req_meta.ts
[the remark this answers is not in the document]
eandre: Yeah let's remove `exposed`, but can keep `requiresAuth`

tsparser/src/builder/codegen.rs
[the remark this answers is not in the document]
fredr: ah, I misunderstood what you meant, I've changed to always set it, but only set `import_path` if there is a `encore.service.ts`

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -261,5 +265,133 @@ api.static = function staticAssets(options: StaticOptions) {
   return new StaticAssets(options);
 };
 
+export interface MiddlewareOptions {
+  /**
+   * Configuration for what endpoints that should be targeted by the middleware
+   */
+  target?: {
+    /**
```
fredr, line 273: are we happy with naming here (eg isRaw vs exposed, use a more unified naming)? 
anything that we should remove or add?

runtimes/js/encore.dev/req_meta.ts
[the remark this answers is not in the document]
eandre: Yeah, let's go with `auth` then

runtimes/js/encore.dev/api/mod.ts
fredr: and then I guess `expose` and not `exposed` as that is what it is called in the APIOptions?

eandre: Yeah, sounds good

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
+---
+seotitle: Using Middleware in your backend application
```
marcuskohlberg, line 2:
```suggestion
seotitle: Using Middleware in your Encore.ts application
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+seodesc: See how you can use middleware in your backend application to handle cross-cutting generic functionality, like request logging, auth, or tracing.
```
marcuskohlberg, line 3:
```suggestion
seodesc: See how you can use middleware in your Encore.ts application to handle cross-cutting generic functionality, like request logging, auth, or tracing.
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+title: Middleware
+subtitle: Handling cross-cutting, generic functionality
+lang: lang
+---
+
+Middleware is a way to write reusable code that runs before or after (or both)
+the handling of API requests, often across several (or all) API endpoints.
```
marcuskohlberg, line 10:
```suggestion
Middleware is a way to write reusable code that runs before, after, or both before and after
the handling of API requests, often across several (or all) API endpoints.
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+
+It's commonly used to implement cross-cutting concerns like
+[request logging](/docs/ts/observability/logging),
+[authentication](/docs/ts/develop/auth),
+[tracing](/docs/ts/observability/tracing),
+and so on. One of the benefits of Encore is that
+all of these use cases are already handled out-of-the-box, so there's no
+need to use middleware for those things.
```
marcuskohlberg, line 18:
```suggestion
Middleware is commonly used to implement cross-cutting concerns like
[request logging](/docs/ts/observability/logging),
[authentication](/docs/ts/develop/auth),
[tracing](/docs/ts/observability/tracing),
and so on. One of the benefits of Encore.ts is that
it handles these common use cases out-of-the-box, so there's no
need to write your own middleware.
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+
+Nonetheless, there are several use cases where it can be useful to write
+reusable functionality that applies to multiple API endpoints, and middleware
+is a good solution in those cases.
```
marcuskohlberg, line 22:
```suggestion
However, when developing applications there's often some use cases where it can be useful to write
reusable functionality that applies to multiple API endpoints, and middleware
is a good solution for this.
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+
+Encore provides built-in support for middlewares by adding functions to the
```
marcuskohlberg, line 24:
```suggestion
Encore provides built-in support for middleware by adding functions to the 
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+the incoming request before handing it off to the next middleware by calling the
+`next` function that's passed in as an argument. For the last middleware in the
+chain, calling `next` results in the actual API handler being called.
+
+The `req` parameter provides information about the incoming request, it has different fields
+depending on what kind of handler it is.
+
+Get information about the current request via `req.requestMeta` if the endpoint is a
+[typed API endpoint](/docs/ts/primitives/defining-apis) or a
+[Streaming API endpoint](/docs/ts/primitives/streaming-apis).
```
marcuskohlberg, line 60:
```suggestion
You can get information about the current request via `req.requestMeta` if the endpoint is a
[typed API endpoint](/docs/ts/primitives/defining-apis) or a
[Streaming API endpoint](/docs/ts/primitives/streaming-apis).
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+raw response via `req.rawRequest` and `req.rawResponse`.
+
+The `next` function returns a `HandlerResponse` object which contains the response from the API.
+Extra http headers can be added to the response objects `extraHeaders` fields if the endpoint is
+a [typed API endpoint](/docs/ts/primitives/defining-apis).
+
+## Middleware ordering
+
+Middlewares run in the order they are defined in the [Service definitions](/docs/ts/primitives/services)
+configuration, i.e:
```
marcuskohlberg, line 75:
```suggestion
Middleware runs in the order they are defined in the [Service definitions](/docs/ts/primitives/services)
configuration, i.e:
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+        third
+    ],
+});
+
+```
+
+## Targeting APIs
+
+The `target` option can be used to decide what endpoints within the service the middleware should run for.
+If the target option is not set, it will run for all endpoints.
```
marcuskohlberg, line 91:
```suggestion
The target option specifies which endpoints within the service the middleware should run on. If not set, the middleware will run for all endpoints by default.
```

docs/ts/develop/middleware.md
```diff
@@ -0,0 +1,95 @@
...
+
+It is preferable performance-wise to use the `target` option over filtering within the middleware function,
+as we calculate what middlewares should run for each endpoint during startup.
```
marcuskohlberg, line 94:
```suggestion
For better performance, use the `target` option instead of filtering within the middleware function. 
This enables calculating applicable middleware per endpoint during startup, reducing runtime overhead.
```

docs/ts/develop/middleware.md
marcuskohlberg: So when saying "Middleware runs in the order they are defined..." you are talking about individual middleware, i.e. they run individually in an order.

I believe grammatically this is correct, and gpt agrees (lol). The idea of the suggestion was to avoid using "middlewares" as this phrasing is quite uncommon as far as I can tell?

fredr: Ah right, that makes sense, my English is just not good enough 😄

runtimes/js/src/api.rs
```diff
@@ -87,15 +88,36 @@ impl Request {
...
+        let obj: JsObject = val
+            .try_into()
+            .map_err(|e| api::Error::invalid_argument("invalid handler response", e))?;
+
+        let payload = obj
+            .get_named_property::<napi::JsUnknown>("payload")
+            .map_err(|e| api::Error::invalid_argument("payload missing in handler response", e))?;
+
+        let extra_headers = obj
+            .get_named_property::<napi::JsUnknown>("extraHeaders")
```
eandre, line 110: Can we make `extraHeaders` optional, and treat it being missing as no headers? And if there is no middleware we don't send it?

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -261,5 +265,128 @@ api.static = function staticAssets(options: StaticOptions) {
...
+   * a typed handler or stream handler.
+   */
+  payload: any;
+
+  /**
+   * extraHeaders can be used by middlewares to add headers to the
+   * response. This only works for typed handler. For raw handlers
+   * see MiddlewareRequest.rawResponse.
+   */
+  extraHeaders: Record<string, string | string[]>;
```
eandre, line 355: I think API-wise it would be better to make this private and expose something like `addHeader` and `setHeader` or something to that effect. Rather than having the caller set the whole object. I think that would compose better with multiple middleware

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -261,5 +265,128 @@ api.static = function staticAssets(options: StaticOptions) {
...
+  options?: MiddlewareOptions;
+}
+
+export function middleware(m: MiddlewareFn): Middleware;
+export function middleware(
+  options: MiddlewareOptions,
+  fn: MiddlewareFn
+): Middleware;
+
+export function middleware(...args: unknown[]): Middleware {
```
eandre, line 380: I think this doesn't need to be varargs, but you can just do:

```
export function middleware(a: MiddlewareFn | MiddlewareOptions, b?: MiddlewareFn): Middleware {
```

runtimes/js/encore.dev/internal/appinit/mod.ts
```diff
@@ -90,36 +181,87 @@ function transformHandler(h: Handler): Handler {
...
+
+function toResponse(payload: any): any {
+  if (payload instanceof Promise) {
+    return (async () => {
+      return new HandlerResponse(await payload);
+    })();
+  } else {
+    return new HandlerResponse(payload);
+  }
+}
```
eandre, line 267: What's the purpose of this function?

fredr: If the handler returns a promise it converts `Promise<any>` into `Promise<HandlerResponse{ payload: any }>`, otherwise the payload field would be a promise in the case that the handler is async

runtimes/js/src/api.rs
fredr: its optiona in ts now, and `get_named_property` will not fail if it isnt set, but return something that is undefined, that we already handle in the parse functions, so I removed those error messages and replaced them with an internal error

PR #1547, opened 2024-11-04
Update graphql.mdx

grevory:
fix graphqls typo


marcuskohlberg:
Thanks, however these are graphql schema files where the use of `.graphqls` is intentional.

PR #1570, opened 2024-11-15
Object storage support

ekerfelt:
Adds native object storage support for Encore.ts and Encore.go


tsparser/src/legacymeta/mod.rs
```diff
@@ -371,6 +378,45 @@ impl<'a> MetaBuilder<'a> {
                     let idx = svc_index.get(&svc.name).unwrap();
                     self.data.svcs[*idx].databases.push(access.db.name.clone());
                 }
+
+                Usage::Bucket(access) => {
+                    let Some(svc) = self.service_for_range(&access.range) else {
+                        HANDLER.with(|h| {
+                            h.span_err(
```
fredr, line 385: can be written shorter like this:
```rust
access.range.err("unable to determine which service is accessing this bucket")
```

fredr (approved):
:bucket:

PR #1587, opened 2024-11-20
CORS fixes

fredr:
This ports the test cases from the go runtime to the rust runtime and makes them pass.

It also adds a check on OPTIONS requests on access-control-request-headers to check if its a request with credentials, as the authorization header is not passed on OPTIONS requests.


runtimes/core/src/api/cors/mod.rs
```diff
@@ -62,21 +75,38 @@ pub fn config(cfg: &pb::gateway::Cors, meta: MetaHeaders) -> anyhow::Result<Cors
...
+            if req.headers.contains_key(AUTHORIZATION) || req.headers.contains_key(COOKIE) {
+                return true;
+            }
+
+            if req.method == http::method::Method::OPTIONS {
+                return req
+                    .headers
+                    .get(ACCESS_CONTROL_REQUEST_HEADERS)
+                    .and_then(|val| val.to_str().ok())
+                    .map(|val| val.to_lowercase().contains("authorization"))
```
eandre, line 98: Should we check for `cookie` here, too?

runtimes/go/appruntime/apisdk/cors/cors.go
```diff
@@ -67,8 +67,20 @@ func Options(cfg *config.CORS, staticAllowedHeaders, staticExposedHeaders []stri
 		AllowOriginRequestFunc: func(r *http.Request, origin string) bool {
 			// If the request has credentials, look up origins in AllowOriginsWithCredentials.
 			// Credentials are cookies, authorization headers, or TLS client certificates.
-			hasCreds := len(r.Cookies()) > 0 || r.Header["Authorization"] != nil || (r.TLS != nil && len(r.TLS.PeerCertificates) > 0)
-			if hasCreds {
+			hasCreds := func(r *http.Request) bool {
+				if len(r.Cookies()) > 0 || r.Header["Authorization"] != nil || (r.TLS != nil && len(r.TLS.PeerCertificates) > 0) {
```
eandre, line 71:
```suggestion
				if len(r.Cookies()) > 0 || len(r.Header["Authorization"]) > 0 || (r.TLS != nil && len(r.TLS.PeerCertificates) > 0) {
```

runtimes/go/appruntime/apisdk/cors/cors.go
```diff
@@ -67,8 +67,20 @@ func Options(cfg *config.CORS, staticAllowedHeaders, staticExposedHeaders []stri
...
+					return true
+				}
+
+				if r.Method == http.MethodOptions {
+					if val := r.Header.Get("Access-Control-Request-Headers"); val != "" {
+						return strings.Contains(strings.ToLower(val), "authorization")
```
eandre, line 77: Should we check for `cookie` here too? Not sure it is present. Also, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Headers the header names are always lowercase, so can skip `strings.ToLower`.

Finally, we need to be more careful than just `strings.Contains`, otherwise someone could set a header like `x-dummy-authorization` and we would pick it up here. Possibly we could use `r.Header.Values` but I'm not sure

runtimes/core/src/api/cors/mod.rs
fredr: I guess so since we check it on the "real" request, added it

runtimes/go/appruntime/apisdk/cors/cors.go
fredr: changed to using `.Values()`, as accessing the map directly isnt case-sensitive afaict

runtimes/go/appruntime/apisdk/cors/cors.go
fredr: `.Values` only returns the value for each header, if it is set multiple times, so we need to split as well

runtimes/core/src/api/cors/mod.rs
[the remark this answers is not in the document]
fredr: fixed

eandre (approved):

runtimes/core/src/api/cors/mod.rs
```diff
@@ -62,21 +75,45 @@ pub fn config(cfg: &pb::gateway::Cors, meta: MetaHeaders) -> anyhow::Result<Cors
...
+            if req.method == http::method::Method::OPTIONS {
+                return req
+                    .headers
+                    .get_all(ACCESS_CONTROL_REQUEST_HEADERS)
+                    .iter()
+                    .any(|val| {
+                        val.to_str()
+                            .map(|val| {
+                                val.split(",")
+                                    .map(|val| val.trim())
```
fredr, line 102: not sure if the trim is actually needed, but I couldn't find any reference in the docs that says explicitly that it is invalid to have a space after the comma

PR #1588, opened 2024-11-21
docs: document SQLDatabase.named for Encore.ts

eandre:


docs/ts/primitives/databases.md
```diff
@@ -216,6 +216,22 @@ ORMs like [Prisma](/docs/ts/develop/orms/prisma) and [Drizzle](/docs/ts/develop/
...
+
+In order to share a database between services, you have two options.
+
+You can define the `SQLDatabase` object in a shared module as an exported variable, and reference this object
+from every service that needs to access the database.
+
+Or, alternatively, you can have define the `SQLDatabase` object inside one of the services using `new SQLDatabase("name", ...)`,
+and have the other services that need to access it create their own reference using `SQLDatabase.named("name")`.
+
+Both of these approaches have the same effect, but the latter is more explicit.
```
marcuskohlberg, line 233:
```suggestion
There are two primary ways of sharing a database between services:

- You can define the `SQLDatabase` object in a shared module as an exported variable, and reference this object
from every service that needs to access the database.
- You can define the `SQLDatabase` object in one service using `new SQLDatabase("name", ...)`, and have other services access it by creating a reference using `SQLDatabase.named("name")`.

Both approaches have the same effect, but the latter is more explicit.
```

marcuskohlberg (approved):

PR #1591, opened 2024-11-21
tsparser+runtimes/core: support root path for enpoints

fredr:


proto/encore/parser/meta/v1/meta.proto
```diff
@@ -290,6 +290,7 @@ message PathSegment {
     PARAM = 1;
     WILDCARD = 2;
     FALLBACK = 3;
+    ROOT = 4;
```
eandre, line 293: Thoughts on representing this instead as a LITERAL segment with value `""`?

fredr: would be nice not to have to update the protos, I just added ROOT so its more explicit, but I'm happy to change to that!

eandre: I think I'd rather do the LITERAL; it seems the logic is the same for literals everywhere. And otherwise we need to modify a lot of other code (dev dashboard, etc) to handle the new proto enum case

fredr:
@eandre I've tested generating a client (and added a clientgen testdata in the pr as well), tested importing the path via "~encore/clients", and tested implementing typed, raw and stream endpoints. Anything else you think should be added?

eandre (approved):

PR #1612, opened 2024-12-01
fix: use test mode as not production

rtritto:
Test mode should be used for `test` and `development` and not for `production`.

### DIFF
--- a/runtimes/js/encore.dev/internal/runtime/mod.ts
+++ b/runtimes/js/encore.dev/internal/runtime/mod.ts
@@ -3,7 +3,7 @@ import { Runtime } from "./napi/napi.cjs";
 
 export * from "./napi/napi.cjs";
 
-const testMode = process.env.NODE_ENV === "test";
+const testMode = process.env.NODE_ENV !== "production";
 
 export const RT = new Runtime({
   testMode,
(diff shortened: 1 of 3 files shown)


eandre:
What problem is this change solving for you? I'm not sure this is the right thing to do to in the general case.

rtritto:
In development, usually the `process.env.NODE_ENV` is undefined or `development`.
So when I started, I got:
```console
file:///C:/Users/<USER>/AppData/Local/Yarn/Berry/cache/encore.dev-npm-1.44.6-3d86b75c4d-10c0.zip/node_modules/encore.dev/dist/internal/runtime/mod.js:5
export const RT = new Runtime({
                  ^

Error: failed to initialize runtime: unable to parse app metadata file

Caused by:
    failed to read file: The system cannot find the path specified. (os error 3)
    at file:///C:/test-encore-win/Local/Yarn/Berry/cache/encore.dev-npm-1.44.6-3d86b75c4d-10c0.zip/node_modules/encore.dev/dist/internal/runtime/mod.js:5:19
    at ModuleJob.run (node:internal/modules/esm/module_job:262:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:483:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5) {
  code: 'GenericFailure'
}
```
because of:
https://github.com/encoredev/encore/blob/af07a53709ed9942f7fc6a1489cec3ff539aaa54/runtimes/js/encore.dev/internal/runtime/mod.ts#L6

PR #1631, opened 2024-12-06
runtimes/js: Added middleware functionality

fredr:
Adds a way to set custom status codes and an api for passing data from middlewares to the handler.

```ts
const mw = middleware(async (req, next) => {
  // add data that can be read from the handler
  req.data.myMiddlewareData = { some: "data" };

  const resp = await next(req);

  // change the status code on the response
  resp.status = 418;

  return resp
});

```

To access data in the handler
```ts
export const get = api(
  { expose: true, method: "GET", path: "/hello/:name" },
  async ({ name }: { name: string }): Promise<Response> => {
    const callMeta = currentRequest() as APICallMeta;
    const myData = callMeta.middlewareData?.myMiddlewareData;

    return { message: "Hello " + name };
  },
);

```

### DIFF
--- a/runtimes/js/src/request_meta.rs
+++ b/runtimes/js/src/request_meta.rs
@@ -5,30 +5,7 @@ use encore_runtime_core::model;
 
 use crate::pvalue::PVals;
 
-#[derive(thiserror::Error, Debug)]
-pub enum Error {
-    #[error("failed serializing json")]
-    SerializeJson(#[from] serde_json::Error),
-    #[error("mutex poisoned")]
-    MutexPoison,
-}
-
-impl<T> From<std::sync::PoisonError<T>> for Error {
-    fn from(_value: std::sync::PoisonError<T>) -> Self {
-        Self::MutexPoison
-    }
-}
-
-impl From<Error> for napi::Error {
-    fn from(err: Error) -> Self {
-        match err {
-            Error::SerializeJson(error) => napi::Error::from(error),
-            Error::MutexPoison => napi::Error::from(err),
-        }
-    }
-}
-
-pub fn meta(req: &model::Request) -> Result<RequestMeta, Error> {
+pub fn meta(req: &model::Request) -> Result<RequestMeta, serde_json::Error> {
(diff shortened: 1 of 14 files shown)


runtimes/js/src/request_meta.rs
```diff
@@ -32,6 +55,7 @@ pub fn meta(req: &model::Request) -> Result<RequestMeta, serde_json::Error> {
                     .map(serde_json::to_value)
                     .transpose()?,
                 headers: serialize_headers(&rpc.req_headers),
+                middleware_data: req.middleware_data.lock()?.clone().map(PVals),
```
fredr, line 58: I think this should be fine, as there shouldn't be any contention on this lock, but alternatively we can add a function that fetches this data on demand instead

solarsoft0:
any luck on this PR, cc @eandre, I actually need a way to set context/data from an handler too; after hook could benefit from this to determine behavior, or is there another alternative way to do this aside from raw? 

seems blocking my effort on this, and i'm avoiding forklift migration. https://github.com/solarsoft0/encore-better-auth/issues/1

solarsoft0:
> any luck on this PR, cc @fredr @eandre, I actually need a way to set context/data from an handler too; after hook could benefit from this to determine behavior, or is there another alternative way to do this aside from raw?
> 
> seems blocking my effort on this, and i'm avoiding forklift migration. [solarsoft0/encore-better-auth#1](https://github.com/solarsoft0/encore-better-auth/issues/1)

without this, i would need to resort to handling middleware manually like its currently being done, which makes it less of encore, if that makes sense.

fredr:
hey @solarsoft0, this PR stalled over Christmas and new years vacation, but I'm back now and we should soon get this merged. We are exploring some different approaches for the implementation first

solarsoft0:
> hey @solarsoft0, this PR stalled over Christmas and new years vacation, but I'm back now and we should soon get this merged. We are exploring some different approaches for the implementation first

thank you! I can see it's reaaaaady!

PR #1635, opened 2024-12-06
Add a PUT handler for the local bucket server.

erikcarlsson:


cli/daemon/objects/public.go
```diff
@@ -1,12 +1,17 @@
 package objects
```
eandre, line 2: This package implements essentially the “public CDN” for buckets. Do we want to serve the signed uploads/downloads from there, too? I would have thought it’s something implemented by the API server. Our local corresponding thing would be the `gcsemu` package.

PR #1647, opened 2024-12-11
dockerbuild: make a valid unix path when converting from windows path with volume

fredr:


eandre (approved):

pkg/dockerbuild/spec.go
```diff
@@ -174,6 +176,14 @@ func (h HostPath) JoinHost(p HostPath) HostPath {
 	return h.Join(string(p))
 }
 func (h HostPath) ToImage() ImagePath {
+	if runtime.GOOS == "windows" {
+		// convert windows path with volume to a unix path, i.e c:\some\path -> /c/some/path
+		volume := filepath.VolumeName(string(h))
+		if len(volume) == 2 && volume[1] == ':' {
+			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(strings.TrimPrefix(string(h), volume)))
```
eandre, line 183: Since we know `volume` is 2 bytes, how about:

```suggestion
			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(h[2:]))
```

fredr: clever!

pkg/dockerbuild/spec.go
```diff
@@ -174,6 +176,14 @@ func (h HostPath) JoinHost(p HostPath) HostPath {
 	return h.Join(string(p))
 }
 func (h HostPath) ToImage() ImagePath {
+	if runtime.GOOS == "windows" {
+		// convert windows path with volume to a unix path, i.e c:\some\path -> /c/some/path
+		volume := filepath.VolumeName(string(h))
+		if len(volume) == 2 && volume[1] == ':' {
+			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(h[2:]))
```
eandre, line 183: We might need:
```suggestion
			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(string(h[2:]))
```

pkg/dockerbuild/spec.go
```diff
@@ -174,6 +176,14 @@ func (h HostPath) JoinHost(p HostPath) HostPath {
 	return h.Join(string(p))
 }
 func (h HostPath) ToImage() ImagePath {
+	if runtime.GOOS == "windows" {
+		// convert windows path with volume to a unix path, i.e c:\some\path -> /c/some/path
+		volume := filepath.VolumeName(string(h))
+		if len(volume) == 2 && volume[1] == ':' {
+			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(string(h[2:]))
```
eandre, line 183:
```suggestion
			return ImagePath("/" + string(volume[0]) + filepath.ToSlash(string(h[2:])))
```

eandre (approved):

PR #1656, opened 2024-12-16
builder: make Stderr configurable

eandre:
This will make it easier to configure where build errors are redirected, for use in e.g. Encore Cloud CI pipelines.


fredr (approved):
nice

PR #1657, opened 2024-12-18
dockerbuild: Copy encore.dev instead of linking

fredr:
This also changes where we put the js runtime, instead of mirroring the host system we put it in /encore where we also store infra config etc.


fredr:
Closing this as there is a better solution on the way

Issue #416, opened 2022-10-08
Support watching go:embed files in `encore.run`


Minivera:
The `notify.Watch` call during the execution of `encore.run` is configured to only watch go files and the files that encore need to compile the go application, https://github.com/encoredev/encore/blob/628ce4ec1e00d8a86c64b2a7b5ef1e45386c6727/cli/daemon/run/watch.go#L55. 

If embedding files into the binary with `go:embed` for example, encore will not recompile when those file change unless you make changes to the go files directly. It would be great if Encore could detect the files I embed in my code and rebuild when those changes, like with JavaScript files for examples.

simon-johansson:
This isn't something we're planning to implement right now, but as a workaround you could set up a separate file watcher (using something like `fswatch`, `entr`, or `nodemon`) that monitors your embedded files and restarts `encore run` when they change.
If this is something you'd still like to see built into Encore, feel free to open a discussion under the Suggestions topic and we can gauge community interest!

PR #1667, opened 2024-12-20
cli/daemon: support watching embedded files

bn4t:
Regarding #416.

Added a very basic implementation to support watching files embedded using `//go:embed`.

Might make sense to track go files using watcher events instead of walking the whole project every time. Let me know if you want me to add that.

### DIFF
--- a/cli/daemon/run/watch.go
+++ b/cli/daemon/run/watch.go
@@ -1,6 +1,8 @@
 package run
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


eandre:
As much as this functionality would be desirable, the solution is not performant enough to read every single file that has a file modification event. I'm afraid it's a bit of a non-starter as is written.

bn4t:
@eandre Thanks for the feedback. I added now tracking of embedded files, so that embedding or unembedding a new file doesn't cause a rescan of the whole project directory tree.

What do you think of this implementation?

PR #1722, opened 2025-01-17
Add user configuration support

eandre:
This adds the `encore config` command for getting and setting configuration values.


cli/cmd/encore/config/config.go
```diff
@@ -0,0 +1,127 @@
...
+		} else if forceGlobal {
+			appScope = false
+		}
+
+		if appScope && appRoot == "" {
+			// If the user specified --app, error if there is no app.
+			cmdutil.Fatal(cmdutil.ErrNoEncoreApp)
+		}
+
+		if len(args) == 2 {
```
leofmarciano, line 59: #### Suggestion: Refactor Decision Structure for Readability

The command validates `cobra.RangeArgs(0, 2)`, but the method body uses different checks for the case `len(args) == 2` and other cases, all within a large `if/else` block. 

For improved readability, consider refactoring the logic into a `switch` statement based on `len(args)`. For example:

```go
switch len(args) {
case 2:
    // Set config value
case 1:
    // Get config value
case 0:
    // Used only with --all; otherwise, display usage and exit
}
```

This makes the flow more explicit and easier to follow. While the current implementation is not incorrect, this change enhances clarity and maintainability.

eandre: Thanks for the feedback @leofmarciano! I agree this would be a bit easier to follow, but probably not enough to warrant refactoring it now. The code is not very likely to change, so probably not worth it for now.

ekerfelt (approved):

PR #1726, opened 2025-01-20
runtimes/js: Add endpoint tags

fredr:


runtimes/js/encore.dev/internal/appinit/mod.ts
```diff
@@ -91,42 +92,19 @@ function calculateMiddlewareChain(
...
+    const target = m.options?.target;
+    if (!target) return true;
+    const { auth, expose, isRaw, isStream, tags } = target;
+    return (
+      (auth === undefined || auth === endpointOptions.auth) &&
+      (expose === undefined || expose === endpointOptions.expose) &&
+      (isRaw === undefined || isRaw === endpointOptions.isRaw) &&
+      (isStream === undefined || isStream === endpointOptions.isStream) &&
+      (tags === undefined ||
+        tags.some((tag) => endpointOptions.tags.includes(tag)))
```
eandre, line 105: This causes the middleware to match if any tag matches. Is that the behavior we want? Or do we want all tags to match?

fredr: that is how it works in go if I understand it correctly, thats why I choose that logic

eandre (approved):

eandre (approved):

PR #1729, opened 2025-01-21
tsparser: make tags non-nullable

eandre:
This makes the `tags` field always set, in line with what the runtime expects. Otherwise we can end up with type errors when tags are undefined.

Thanks Daan Kets for reporting.


daankets (approved):
For what it's worth @eandre ;-)

fredr (approved):

PR #1737, opened 2025-01-24
Add DB migration status in the dev dashboard.

erikcarlsson:


cli/daemon/dash/dash.go
```diff
@@ -264,7 +265,29 @@ func (h *handler) Handle(ctx context.Context, reply jsonrpc2.Replier, r jsonrpc2
...
+		}
+
+		// Find the latest app by platform ID or local ID.
+		app, err := h.apps.FindLatestByPlatformOrLocalID(params.AppID)
+		if err != nil {
+			return reply(ctx, nil, err)
+		}
 
+		// Now find the running instance(s)
+		runInstance := h.run.FindRunByAppID(params.AppID)
```
eandre, line 283: We should be able to respond with this without a running instance; the database can be running even if the app is not. We can fall back to the cached metadata using `app.CachedMetadata()`

cli/daemon/dash/dash.go
```diff
@@ -1015,6 +1038,16 @@ type appStatus struct {
 	CompileError string                `json:"compileError,omitempty"`
 }
 
+type dbMigrationHistory struct {
+	DatabaseName string              `json:"databaseName"`
+	Migrations   []dbMigrationStatus `json:"migrations"`
+}
+
+type dbMigrationStatus struct {
+	Migration *meta.DBMigration `json:"migration"`
```
eandre, line 1047: Let's define our own struct instead of using this. Protobuf JSON encoding is different from regular Go json encoding, so I'm a bit uneasy with reusing protobuf structs in a non-protobuf setting

cli/daemon/dash/dash.go
```diff
@@ -1064,3 +1097,66 @@ func buildAppStatus(app *apps.Instance, runInstance *run.Run) (s appStatus, err
...
+	clusterType := sqldb.Run
+	cluster, ok := run.Mgr.ClusterMgr.Get(sqldb.GetClusterID(app, clusterType, run.NS))
+	if !ok {
+		return nil, fmt.Errorf("failed to get database cluster of type %s", clusterType)
+	}
+	var statuses []dbMigrationHistory
+	for _, dbMeta := range md.SqlDatabases {
+		db, ok := cluster.GetDB(dbMeta.Name)
+		if !ok {
+			return nil, fmt.Errorf("failed to get database %s", dbMeta.Name)
```
eandre, line 1122: Maybe better to skip the database than erroring out completely?

cli/daemon/sqldb/db.go
```diff
@@ -477,3 +486,25 @@ func (db *DB) connectSuperuser(ctx context.Context) (*pgx.Conn, error) {
...
+// Connects as a superuser or admin to the database. Fails fast if the cluster
+// is not running yet.
+// On success the returned conn must be closed by the caller.
+func (db *DB) connectToDB(ctx context.Context) (*sql.Conn, error) {
+	info, err := db.Cluster.Info(ctx)
+	if err != nil {
+		return nil, err
+	}
+	uri := info.ConnURI(db.EncoreName, info.Config.Superuser)
+	pool, err := sql.Open("pgx", uri)
```
eandre, line 499: Can we open a single pgx conn instead of setting up a connection pool?

cli/daemon/dash/dash.go
erikcarlsson: Done

cli/daemon/dash/dash.go
erikcarlsson: Yes, good point. Forgot to JSON convert. Fixed like you proposed instead.

cli/daemon/dash/dash.go
erikcarlsson: Done.

eandre (approved):

cli/daemon/sqldb/db.go
erikcarlsson: As discussed out-of-PR: it's not trivial if we want to reuse the existing lookup code for migrations. I agree opening a new pool repeatedly isn't a great usage pattern. Let's rewire this if we see any issues.

PR #1777, opened 2025-02-23
Increase max message size

Willyham:
Generating debug metadata for large apps fails due to default gRPC message size limits:

```
git:(main) encore debug meta --format json
error: grpc: received message larger than max (6181042 vs. 4194304)
```

This increases the limit from 4mb to 16mb.


eandre (approved):
Thanks!

PR #1782, opened 2025-02-25
runtimes/core: decode path before verifying seal

fredr:
When we create the sign in go, we use the decoded path, so make sure we decode it before verify, otherwise paths like `/åäö` will fail verification.

### DIFF
--- a/runtimes/core/src/api/reqauth/platform.rs
+++ b/runtimes/core/src/api/reqauth/platform.rs
@@ -4,6 +4,7 @@ use base64::engine::general_purpose;
 use base64::Engine;
 use encore::runtime::v1 as pb;
 use hmac::Mac;
+use percent_encoding::percent_decode_str;
 use std::fmt::Display;
 use std::time::SystemTime;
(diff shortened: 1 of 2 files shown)


runtimes/core/src/api/reqauth/platform.rs
```diff
@@ -69,6 +70,10 @@ impl RequestValidator {
     }
 
     pub fn sign_outgoing_request(&self, req: &mut reqwest::Request) -> anyhow::Result<()> {
+        let path = percent_decode_str(req.url().path())
+            .decode_utf8_lossy()
+            .to_string();
```
eandre, line 75: I don't think we need `.to_string()` here? `Cow<'_, str>` should have `as_bytes()` too.

fredr: This is to battle the borrow checker, `req` is mutually borrowed below when we insert the date header. So we can't also have a non-mutual reference at the same time

PR #1791, opened 2025-03-04
object storage: Apply list prefix from options

ekerfelt:


runtimes/core/src/objects/s3/bucket.rs
```diff
@@ -97,6 +97,12 @@ impl objects::BucketImpl for Bucket {
                     req = req.prefix(key_prefix);
                 }
 
+                if let Some(key_prefix) = &options.prefix {
+                    let current_prefix = req.get_prefix().as_ref().unwrap_or(& String::new()).to_owned();
```
fredr, line 101: nit, I would probably do something like this instead, but feel free to ignore:

```rust
                    let current_prefix = req.get_prefix().as_deref().unwrap_or_default();
                    let new_prefix = format!("{}{}", current_prefix, key_prefix);
                    req = req.prefix(new_prefix);
```

fredr (dismissed):

runtimes/core/src/objects/s3/bucket.rs
ekerfelt: 👍

fredr (approved):

PR #1794, opened 2025-03-05
Better support for ts workspaces

fredr:


pkg/dockerbuild/spec.go
```diff
@@ -469,8 +490,76 @@ func (b *imageSpecBuilder) allocArtifactDir(out builder.BuildOutput) *imageArtif
...
 }
+
+// DetermineIncludes determines what paths within the workspace should be included in the image.
+func DetermineIncludes(appLang appfile.Lang, bundleSource bool, workspaceRoot string, appRoot string) ([]RelPath, error) {
+	// if the actual setting is set, include all files from the workspace
+	if bundleSource {
+		return []RelPath{"."}, nil
+	}
+
+	// app root is always includede
```
fredr, line 507:
```suggestion
	// app root is always included
```

pkg/dockerbuild/tarcopy.go
```diff
@@ -93,13 +96,45 @@ func (tc *tarCopier) CopyData(spec *ImageSpec) error {
 	return nil
 }
 
+// shouldInclude returns true if the path should be included in the tar.
+func shouldInclude(desc *dirCopyDesc, path HostPath) bool {
+	if len(desc.IncludeSrcPaths) == 0 {
+		return true
+	}
```
eandre, line 103: This behavior is a bit strange to me; why do we always return `true` if the list is empty?

fredr: good catch, I think this is a remnant from some refactoring, let me have a look, I think we can just remove that

pkg/dockerbuild/spec_test.go
```diff
@@ -44,28 +46,30 @@ func TestBuild_Node(t *testing.T) {
...
+		}),
 		Supervisor:      option.None[SupervisorSpec](),
 		BundledServices: []string{"bar", "foo"},
 		BundledGateways: []string{"baz", "qux"},
 		DockerBaseImage: "scratch",
 		FeatureFlags:    map[FeatureFlag]bool{NewRuntimeConfig: true},
 		StargzPrioritizedFiles: []ImagePath{
-			"/artifacts/0/package.json",
-			"/artifacts/0/build/entrypoint",
+			"/image/artifacts/entrypoint",
```
eandre, line 72: Are we still producing this `artifacts` directory? I believe it's not really needed anymore since the entrypoint is under `/workspace`?

fredr: no, this is just from what we set the entrypoint to in this test, the /artifacts/0/... are removed from typescript images

eandre (approved):

PR #1806, opened 2025-03-11
gcsemu: trim os path separator

fredr:
on windows the separator in \\, so it will not be stripped


fredr:
also added fix upstream https://github.com/fullstorydev/emulators/pull/291

erikcarlsson (approved):

eandre (approved):

PR #1808, opened 2025-03-12
[errs] use errors.As instead of type assertions

twindebank:
I'd like to be able to embed the `errs.Error` struct without losing the error metadata when the handlers in `errs` are used. Currently, if I try to use an error struct with an embedded `*errs.Error` struct, the type assertions fail and the error metadata is lost. 

The use case is a bit niche, but if this change is controversial I can go into more details.

### DIFF
--- a/runtimes/go/beta/errs/error.go
+++ b/runtimes/go/beta/errs/error.go
@@ -111,15 +111,15 @@ func Convert(err error) error {
 		return nil
 	}
 
+	if e, ok := err.(*Error); ok {
+		// If the parent is already an *Error we can return it directly.
+		return e
+	}
+
 	var e *Error
 	if errors.As(err, &e) {
-		if directErr, ok := err.(*Error); ok {
-			// If the parent is already an *Error, return it directly
-			return directErr
-		}
-
-		// The error itself isn't an *Error, but it wraps one somewhere in the chain
-		// Create a new *Error that preserves the outer error but takes properties from inner *Error
+		// The error itself isn't an *Error, but it wraps one somewhere in the chain. Create a new *Error that preserves
+		// the outer error but takes properties from inner *Error
 		return &Error{
 			Code:       e.Code,
 			Message:    err.Error(),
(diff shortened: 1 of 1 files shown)


runtimes/go/beta/errs/error.go
```diff
@@ -106,9 +109,13 @@ func WrapCode(err error, code ErrCode, msg string, metaPairs ...interface{}) err
 func Convert(err error) error {
 	if err == nil {
 		return nil
-	} else if e, ok := err.(*Error); ok {
+	}
+
+	var e *Error
+	if errors.As(err, &e) {
```
eandre, line 115: I don't think it makes sense to use `errors.As` here; we don't want to return an inner error as that would drop any outer errors that wrap it.

eandre: Yeah I guess it's fine to copy the fields (metadata, code, etc) from the inner error if it exists, while still preserving the full error chain

eandre: Note I would still want to keep the original special case of the type assertion, as a fast path.

runtimes/go/beta/errs/error.go
```diff
@@ -167,9 +193,10 @@ func (e *Error) ErrorMessage() string {
 	var next error = e.underlying
 	for next != nil {
 		var msg string
-		if e, ok := next.(*Error); ok {
-			msg = e.Message
-			next = e.underlying
+		var ee *Error
+		if errors.As(next, &ee) {
+			msg = ee.Message
+			next = ee.underlying
```
eandre, line 199: This will skip the error message of any intermediate errors in the chain between this error and the next wrapped `*Error`. I think that's undesirable, and that we should probably revert this part of the PR

twindebank: good catch! have updated

twindebank:
@eandre is this good to go?

eandre (approved):

PR #1814, opened 2025-03-14
tsparser: duplicate api name error

fredr:
emit error when two endpoints with the same name are defined within the same service

```
error: api endpoints with conflicting names defined within the same service
  --> /home/fredr/projects/encore-apps/ts-tags/other/otherfile.ts:3:14
   |
3  | export const get1 = api(
   |              ^^^^
   |
note: previously defined here
  --> /home/fredr/projects/encore-apps/ts-tags/other/other.ts:18:14
   |
18 | export const get1 = api(
   |              ^^^^

```

### DIFF
--- a/tsparser/src/app/mod.rs
+++ b/tsparser/src/app/mod.rs
@@ -7,7 +7,6 @@ use swc_common::Span;
 use crate::encore::parser::meta::v1;
 use crate::legacymeta::compute_meta;
 use crate::parser::parser::{ParseContext, ParseResult};
-use crate::parser::resourceparser::bind::Bind;
 use crate::parser::resources::apis::api::{Endpoint, Method, Methods};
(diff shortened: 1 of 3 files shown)


tsparser/src/app/mod.rs
```diff
@@ -92,6 +93,29 @@ impl AppValidator<'_> {
     }
 
     fn validate_apis(&self) {
+        let mut seen = std::collections::HashMap::new();
+        for bind in &self.parse.binds {
```
eandre, line 97: I think we should do this based on resources, not binds

fredr: we can, but then we dont have a range that point to the identifier. We could add that ofc

eandre (approved):

eandre (approved):

PR #1815, opened 2025-03-14
clientgen: Add support for sharing types between ts backend and frontend

ekerfelt:


pkg/clientgen/testdata/tsapp/expected_shared.ts
```diff
@@ -0,0 +1,852 @@
...
+     * a function which returns a new object for each request.
+     */
+    auth?: RequestType<typeof auth_Auth> | AuthDataGenerator
+}
+
+/**
+ * Import the endpoint handlers to derive the types for the client.
+ */
+import {
+    dummy as api_Svc_Svc_Dummy,
```
eandre, line 79: Why are all the names CamelCased? Doesn't look very idiomatic. Or is that just the input endpoint name?

ekerfelt: escape bug. fixed

PR #1827, opened 2025-03-21
Add option to redirect tsparser stderr output.

erikcarlsson:


eandre:
Merged https://github.com/encoredev/encore/pull/1656 instead, so closing this

PR #1837, opened 2025-03-27
refactor: use the built-in max/min to simplify the code

evenevent:
Use the built-in max/min from the standard library in Go 1.21 to simplify the code.


luisnquin (changes requested):
I don't think that the complexity is that high to change that snippet. In other hand, the diff could be despicable for certain cases but you're evaluating the result by calling first `min` and then `max`, this will make the flow unnecessarily more long, sacrificing performance.

evenevent:
> I don't think that the complexity is that high to change that snippet. In other hand, the diff could be despicable for certain cases but you're evaluating the result by calling first `min` and then `max`, this will make the flow unnecessarily more long, sacrificing performance.

Thank you for the reply. It seems to be the case. I will close this PR.

PR #1847, opened 2025-04-02
database transaction support in typescript

fredr:
Adds support for transactions in runtimes/core and runtimes/js

We should probably merge https://github.com/encoredev/rust-postgres/pull/1 and update Cargo.toml again. (suggested that change upstream as well https://github.com/sfackler/rust-postgres/pull/1230)


runtimes/js/encore.dev/storage/sqldb/database.ts
```diff
@@ -329,15 +331,13 @@ export class Transaction extends BaseQueryExecutor {
   async rollback() {
     const source = getCurrentRequest();
     await this.impl.rollback(source);
+    this.done = true;
```
eandre, line 334: I wonder if it's more correct to set this before starting the rollback than when it completes?

eandre (approved):

PR #1861, opened 2025-04-08
fix/make RPC functions and parameters accessible in generated Go client

janceesvansenden:
# Changes
This PR addresses a problem with Go client generation from a Typescript API. Currently naming of RPC functions and parameters are taken straight from the Typescript codebase. This is a problem because in Typescript these typically start with lowercase characters, making them private and unaccessible in Go.

This PR fixes that by making sure that the first letter is always capitalized for:
- RPC function names
- RPC parameters

# Note for potential further improvement of client generation (Go -> Typescript)
Ideally client generation the other way around (from Go code to Typescript) would also follow Typescript naming conventions of using `camelCase`. Currently it just takes the casing from Go, which means it usually is `PascalCase`. 

I would be happy to submit a PR for this, but also realize this could have backwards compatibility issues so wanted to get your thoughts on the matter first.

### DIFF
--- a/pkg/clientgen/golang.go
+++ b/pkg/clientgen/golang.go
@@ -356,7 +356,7 @@ func (g *golang) generateServiceClient(file *File, service *meta.Service, tags c
 		}
 
 		interfaceMethods = append(interfaceMethods,
-			Id(strings.Title(rpc.Name)).Add(g.rpcParams(rpc)).Add(g.rpcReturnType(rpc, false)),
+			Id(idents.Convert(rpc.Name, idents.PascalCase)).Add(g.rpcParams(rpc)).Add(g.rpcReturnType(rpc, false)),
 		)
 	}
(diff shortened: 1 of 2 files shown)


eandre:
Thanks for doing this! We actually have a package designed for doing identifier conversions in a way that's more idiomatic: https://pkg.go.dev/encr.dev/pkg/idents

Can you change this to use `idents.Convert(input, idents.PascalCase)` instead?

janceesvansenden:
> Thanks for doing this! We actually have a package designed for doing identifier conversions in a way that's more idiomatic: https://pkg.go.dev/encr.dev/pkg/idents
> 
> Can you change this to use `idents.Convert(input, idents.PascalCase)` instead?

Sure, thanks for sharing. Will update that soon!

janceesvansenden:
@eandre Any chance you could review and merge this one soon? We could really use this to avoid our current workarounds.

PR #1863, opened 2025-04-09
clientgen: Add option to add auth handler to an existing client

ekerfelt:


pkg/clientgen/testdata/tsapp/expected_typescript.ts
```diff
@@ -43,6 +45,19 @@ export default class Client {
...
         this.svc = new svc.ServiceClient(base)
     }
+
+    /**
+     * Creates a new Encore client with the given auth handler.
+     *
+     * @param authHandler Authentication data to be used for each request. Either a static  
+     *                    object or a function which returns a new object for each request.
+     **/
+    public withAuth(authHandler: svc.AuthParams | AuthDataGenerator): Client {
```
eandre, line 55: Should the argument also support `| undefined` to unset the auth generator?

We should probably not call it `authHandler` since that is another Encore concept

pkg/clientgen/typescript.go
```diff
@@ -1083,6 +1091,7 @@ if (typeof options === "string") {
 			w := w.Indent()
 
 			w.WriteString("const base = new BaseClient(target, options ?? {})\n")
+			w.WriteString("this.target = target\n")
```
ekerfelt, line 1094: target is only defined for auth services right now.

PR #1870, opened 2025-04-14
add port config to cli docs

marcuskohlberg:


docs/go/cli/cli-reference.md
```diff
@@ -34,6 +34,13 @@ Checks your application for compile-time errors using Encore's compiler.
 $ encore check
 ```
 
+#### Port
+
+Specify the port on which the local application runs
+
+```shell
+$ encore run --port PORT_NUMBER
+```
```
eandre, line 43: This shouldn't be a separate entry, but added to the list of options under `encore run` on line 16: `encore run [--debug] [--port NUMBER] ...`

eandre (approved):

PR #1877, opened 2025-04-17
docs: remove consecutive duplicate words

zeroprooff:
hey team! I found duplicate words and fixed them.

### DIFF
--- a/tsparser/src/parser/resources/apis/service.rs
+++ b/tsparser/src/parser/resources/apis/service.rs
@@ -51,7 +51,7 @@ pub static SERVICE_PARSER: ResourceParser = ResourceParser {
                 }
 
                 // This resource is only allowed to be defined in a module named "encore.service.ts".
-                // Check that that is the case.
+                // Check that is the case.
                 match &pass.module.file_path {
                     FilePath::Real(buf) if buf.ends_with("encore.service.ts") => {}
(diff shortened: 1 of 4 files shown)


tsparser/src/parser/resources/apis/service.rs
```diff
@@ -51,7 +51,7 @@ pub static SERVICE_PARSER: ResourceParser = ResourceParser {
                 }
 
                 // This resource is only allowed to be defined in a module named "encore.service.ts".
-                // Check that that is the case.
+                // Check that is the case.
```
eandre, line 54: The previous version was correct.

v2/app/testdata/pubsub_err_subscriber_different_service.txt
```diff
@@ -1,5 +1,5 @@
 ! parse
-err 'The function passed to `pubsub.NewSubscription` must be declared in the the same service as the'
+err 'The function passed to `pubsub.NewSubscription` must be declared in the same service as the'
```
eandre, line 2: This is testing that the parser returns a specific error. You'll need to update the code that returns that error, too, otherwise the test will fail

eandre (commented):
You still haven't addressed where this error message is actually generated. You're just changing the tests.

PR #1892, opened 2025-04-30
runtimes/{js,core}: add call options to api calls

fredr:
Adds call options for api calls that can be used to set options when calling apis. For now only adds a way of overriding auth data, mainly for, but not limited to, mocking when calling via ~encore/clients in tests.

### DIFF
--- a/runtimes/core/src/api/call.rs
+++ b/runtimes/core/src/api/call.rs
@@ -120,7 +120,7 @@ impl ServiceRegistry {
         target: EndpointName,
         data: JSONPayload,
         source: Option<Arc<model::Request>>,
-        opts: Option<Arc<api::CallOpts>>,
+        opts: Option<api::CallOpts>,
     ) -> impl Future<Output = APIResult<ResponsePayload>> + 'static {
         let tracer = self.tracer.clone();
         let call = model::APICall { source, target };
@@ -131,7 +131,7 @@ impl ServiceRegistry {
             data,
             call.source.as_deref(),
             start_event_id,
-            opts,
+            opts.as_ref(),
         );
         async move {
             let result = fut.await;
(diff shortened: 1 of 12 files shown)


runtimes/core/src/api/call.rs
```diff
@@ -120,12 +120,19 @@ impl ServiceRegistry {
         target: EndpointName,
         data: JSONPayload,
         source: Option<Arc<model::Request>>,
+        opts: Option<Arc<api::CallOpts>>,
```
eandre, line 123: Does it need to be wrapped in an `Arc`?

runtimes/js/src/runtime.rs
```diff
@@ -322,6 +327,25 @@ impl Runtime {
...
+    fn from(value: CallOpts) -> Self {
+        api::CallOpts {
+            auth_data: value.auth_data.as_ref().map(|ad| ad.0.clone()),
+            auth_user_id: value.auth_data.as_ref().and_then(|ad| {
+                ad.0.get("userID")
+                    .and_then(|pv| pv.as_str().map(|s| s.to_string()))
+            }),
+        }
+    }
+}
```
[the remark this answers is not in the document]
fredr: The problem is that in the js runtime the userID is part of the authdata schema, so I think that maybe separating them there will cause some confusion, e.g we could do something like
```
import { svc } from "~encore/clients";

svc.endpoint(parms, { userID: "id1", authData: { email: "..." } });
```
and then merge the userID into the authData, but wouldn't that be a bit confusing? or did you have something else in mind?

runtimes/core/src/api/call.rs
fredr: good question, it feels like I added it there for a reason, but I cant remember why, will remove

runtimes/core/src/api/manager.rs
```diff
@@ -436,3 +438,9 @@ fn listen_addr() -> String {
     }
     "0.0.0.0:8080".to_string()
 }
+
+#[derive(Debug)]
+pub struct CallOpts {
+    pub auth_data: Option<PValues>,
+    pub auth_user_id: Option<String>,
```
eandre, line 445: Thoughts on making this into:

```suggestion
    pub auth: Option<(String, Option<PValues>)>,
```

fredr: switched to a struct so the fields are named :eyes:

eandre (approved):

PR #1896, opened 2025-05-05
Upgrade pgproto

ekerfelt:


cli/daemon/sqldb/proxy.go
```diff
@@ -420,7 +420,11 @@ func (cm *ClusterManager) cancelRequest(client io.Writer, req *pgproxy.CancelDat
 			Code:     "08006",
 			Message:  "database cluster not running",
 		}
-		_, _ = client.Write(msg.Encode(nil))
+		encode, err := msg.Encode(nil)
+		if err != nil {
+			return
+		}
+		_, _ = client.Write(encode)
```
eandre, line 427: Since this is in the error path this could arguably just be
```suggestion
		encode, _ := msg.Encode(nil)
		_, _ = client.Write(encode)
```

eandre (dismissed):

PR #1917, opened 2025-05-19
runtimes/js: Add support for cookies

fredr:
Adds support for adding cookies in auth handler params and in api request/response schema.

Generated client will ignore cookies and hand over the responsibility to the browser.


pkg/clientgen/javascript.go
```diff
@@ -106,6 +108,34 @@ func (js *javascript) Generate(p clientgentypes.GenerateParams) (err error) {
...
+	case *schema.Type_Named:
+		decl := js.md.Decls[typ.GetNamed().Id]
+		return js.getFields(decl.Type)
+	default:
+		return nil
+	}
+}
+
+func (js *javascript) isAuthCookiesOnly() bool {
+	fields := js.getFields(js.md.AuthHandler.Params)
```
eandre, line 127: Might be good to check if `js.md.AuthHandler` is nil here, just to be safe

pkg/clientgen/testdata/tsapp/expected_golang.go
```diff
@@ -91,31 +91,50 @@ func WithAuthFunc(authGenerator func(ctx context.Context) (SvcAuthParams, error)
...
+	Baz        string  // Baz is better
+	QueryFoo   bool    `encore:"optional" query:"foo,optional"`
+	QueryBar   string  `encore:"optional" query:"bar,optional"`
+	HeaderBaz  string  `encore:"optional" header:"baz,optional"`
+	HeaderNum  float64 `encore:"optional" header:"num,optional"`
+	CookieQux  string  `cookie:"qux,optional" encore:"optional"`
+	CookieQuux float64 `cookie:"quux,optional" encore:"optional"`
+}
+
+type SvcRequest struct {
```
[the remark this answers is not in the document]
fredr: This is indeed a problem, but I will address this and several ts->go generation issues (ENC-2928)

runtimes/core/src/api/pvalue.rs
```diff
@@ -50,6 +50,18 @@ impl PValue {
             _ => None,
         }
     }
+
+    pub fn type_name(&self) -> &'static str {
+        match self {
```
eandre, line 55: I would have expected we would introduce a new `PValue::Cookie` enum variant for cookies, so we don't need to treat it as an `Object` with string key lookups like `"maxAge"`

fredr: I've updated to add a cookie variant!

pkg/clientgen/javascript.go
```diff
@@ -489,12 +522,23 @@ func (js *javascript) rpcCallSite(w *indentWriter, rpc *meta.RPC, rpcPath string
 	w.WriteString("\n//Populate the return object from the JSON body and received headers\nconst rtn = await resp.json()\n")
 
 	for _, headerField := range respEnc.HeaderParameters {
+		isSetCookie := strings.ToLower(headerField.WireFormat) == "set-cookie"
+		if isSetCookie {
+			w.WriteString("// Skip set-cookie header in browser context as browsers doesn't have access to read it\n")
+			w.WriteString("if (!BROWSER) {\n")
+			w = w.Indent()
```
fredr, line 529: maybe it would be better to not have the set-cookie fields at all in the type? and then add it back when we create a nodejs server target for the clientgen?

runtimes/core/src/api/auth/mod.rs
```diff
@@ -161,6 +164,30 @@ impl Authenticator {
             }
         };
 
+        // Cookies.
+        if let Some(schema) = &self.schema.cookie {
+            let mut inbound_cookies = cookie::CookieJar::new();
+            for raw in inbound.headers().get_all(http::header::COOKIE) {
+                if let Ok(raw) = String::from_utf8(raw.as_bytes().to_vec()) {
```
eandre, line 171: Is `raw` not a string already? What  type is it? A bit strange to need to first convert to bytes and then to string

runtimes/core/src/api/schema/cookie.rs
```diff
@@ -0,0 +1,343 @@
...
+            let name = field.name_override.as_deref().unwrap_or(key);
+            let c = to_cookie(name, value)?;
+            resp = resp.header(SET_COOKIE, c.to_string());
+        }
+
+        Ok(resp)
+    }
+}
+
+struct RawCookie<'a> {
```
eandre, line 127: How is this used? I was hoping we could keep the JavaScript-specific encoding in the JS runtime crate, if possible

pkg/clientgen/testdata/tsapp/expected_list_of_union_shared.ts
```diff
@@ -4,6 +4,7 @@
 /* eslint-disable */
 /* jshint ignore:start */
 /*jslint-disable*/
+import { CookieWithOptions } from "encore.dev/api";
```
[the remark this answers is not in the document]
fredr: This is for the generated client with shared types enabled, it already imports things from `encore.dev/api` so I think it should be fine?

runtimes/core/src/api/auth/mod.rs
fredr: it is a HeaderValue which is a bit annoying to deal with, we can use `to_str` and then unwrap_or_default. I can see if I can make it a bit less code

runtimes/core/src/api/schema/cookie.rs
fredr: Ah, forgot about that when I was refactoring, will have a look if I can remove that, It was used internally when parsing a cookie out of a PValue

pkg/clientgen/testdata/tsapp/expected_list_of_union_shared.ts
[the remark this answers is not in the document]
eandre: Ah in that case it's fine :)

runtimes/core/src/api/auth/mod.rs
eandre: Mainly trying to avoid unnecessary allocations; I guess if `to_str` fails we wouldn't be able to parse it as a string using the current code either

fredr: ah wait, I thought I had used from_utf8_lossy here to avoid errors, but I see that I havent. I'll have a look and see if I can clean that up a bit

runtimes/core/src/api/schema/cookie.rs
fredr: I've removed this now. The problem is that when we parse pvalues from the js handler we dont have the schema, so there is no way of knowing if it should be a PValue::Object or PValue::Cookie. So I added that we pass the response schema down, and then we transform the parsed PValues based on that schema. Let me know if you think this is a good approach or if you have something else in mind

pkg/clientgen/testdata/tsapp/expected_list_of_union_shared.ts
```diff
@@ -105,10 +108,14 @@ export namespace svc {
 
 type PickMethods<Type> = Omit<CallParameters, "method"> & {method?: Type}
 
+type OmitCookie<T> = {
```
eandre, line 111: If all keys are cookies, this will yield the type `{}` which suggests it's still a value and not `void`. Is that how the backend behaves, or should the type evaluate to `void` in that case? (Annoying questions, I know)

fredr: good question :thinking: when we have shared types turned off the types get rendered as `{}` as well, and I think that works actually, but maybe it would be cleaner to do it the same way as if there where no type, I'll see what I can come up with for both shared and not shared types :eyes:

pkg/clientgen/javascript.go
fredr: will address this when adding cli flags for generating for browser or server target

pkg/clientgen/testdata/tsapp/expected_list_of_union_shared.ts
```diff
@@ -108,14 +108,25 @@ export namespace svc {
...
 
+// Helper type to omit all fields that are cookies.
 type OmitCookie<T> = {
   [K in keyof T as T[K] extends CookieWithOptions<any> ? never : K]: T[K];
 };
 
-type RequestType<Type extends (...args: any[]) => any> =
-  Parameters<Type> extends [infer H, ...any[]] ? OmitCookie<H> : void;
+// Helper type to check if an object type is empty (has no properties)
+type IsEmptyObject<T> = [keyof T] extends [never] ? true : false;
```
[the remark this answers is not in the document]
fredr: yes! that is cleaner

PR #1939, opened 2025-05-27
docs/ts: document cookie support

fredr:
relates to #1917

### DIFF
--- a/docs/ts/primitives/defining-apis.mdx
+++ b/docs/ts/primitives/defining-apis.mdx
@@ -181,7 +181,7 @@ export const getBlogPost = api(
 
 ### Cookie parameters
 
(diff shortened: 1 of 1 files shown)


docs/ts/primitives/cookies.mdx
```diff
@@ -0,0 +1,206 @@
...
+type ImplicitStringCookie = Cookie<"sessionId">;
+
+// Number cookie
+type NumberCookie = Cookie<number, "userId">;
+
+// Boolean cookie
+type BooleanCookie = Cookie<boolean, "darkMode">;
+
+// Date cookie
+type DateCookie = Cookie<Date, "lastVisit">;
```
eandre, line 140: Do we need these type aliases? I think we could get the point across using just the example `UserPreferences` example below

fredr: yeah, its a bit weird, removed!

PR #1944, opened 2025-05-28
runtime/core: encode cookies

fredr:
This adds percent encoding/decoding to cookies. With out it we cant use a lot of characters, like `;`. It seems to be common practice to encode cookies in cookie libs, e.g in universal-cookie etc.


fredr:
Lets not build in encoding into core. If a user want to encode/decode cookies they'll have to manage it themselves

PR #1948, opened 2025-05-28
runtime/core: refactor streaming reporting

fredr:


runtimes/core/src/trace/log.rs
```diff
@@ -57,190 +55,176 @@ pub(super) struct TraceEvent {
...
-                    return Poll::Ready(None);
-                }
-                self.num_events_this_tick = 0;
-
-                // Call the ticker again to schedule a wake-up for the next tick.
-                _ = self.no_data_ticker.poll_tick(cx);
-            }
-            Poll::Pending => {}
+        // Send data
+        let data = streaming_event.event.data.clone();
```
eandre: Would be very nice to not have to clone the data here; that's what I was trying to avoid

runtimes/core/src/trace/log.rs
```diff
@@ -57,190 +55,176 @@ pub(super) struct TraceEvent {
...
-                if self.num_events_this_tick == 0 {
-                    return Poll::Ready(None);
-                }
-                self.num_events_this_tick = 0;
-
-                // Call the ticker again to schedule a wake-up for the next tick.
-                _ = self.no_data_ticker.poll_tick(cx);
-            }
-            Poll::Pending => {}
+        // Send data
```
eandre: I don't think we can break this up into two `sender.send()` calls, since I believe they could be interleaved from different threads? So we could get `[header1, header2, data1, data2]` instead of `[header1, data1, header2, data2]`?

fredr: good point! yeah, I think you are right, I changed to sending the TraceEvent over the channel instead, and then handle it in the body stream

runtimes/core/src/trace/log.rs
fredr: this is a bytes::Bytes, so the clone is cheap, we actually did it in the previous solution as well, but I accidentally removed the comment that it is cheap.
however, with the latest changes I no longer need to clone it, so it have been removed

PR #1958, opened 2025-06-04
tsparser: handle default type params

fredr:
for req/res types like

```ts
export interface MyType<A, B = string> {
   field1: A;
   field2: B;
}
```

### DIFF
--- a/tsparser/src/parser/types/type_resolve.rs
+++ b/tsparser/src/parser/types/type_resolve.rs
@@ -683,8 +683,8 @@ impl Ctx<'_> {
             return Type::Basic(Basic::Date);
         }
 
-        let mut type_arguments =
-            Vec::with_capacity(typ.type_params.as_ref().map_or(0, |p| p.params.len()));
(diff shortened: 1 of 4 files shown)


tsparser/src/parser/types/type_resolve.rs
```diff
@@ -1964,7 +1964,24 @@ impl Ctx<'_> {
...
+            // For each parameter that wasn't provided, try to use its default
+            for param in type_params.iter().skip(args.len()) {
+                if let Some(default) = param.default.as_ref() {
+                    args.push(self.typ(default));
+                }
+            }
+
+            args
+        } else {
+            named.type_arguments.clone()
```
eandre: Can we avoid this clone? We didn't need it before

fredr: its a bit finicky because we create a new vec in the other if branch, but if we break things out to a different function it should be fine. I've pushed a change :rocket:

PR #1963, opened 2025-06-06
Remove redaction section from ts tracing docs

asiFarran:
It appears that there is no support for marking fields for redaction in ts yet.

The docs section looks like it was just copied from the `go` one, and is at the moment incorrect and created a false assumption.

This section of the docs can be restored when redaction support actually exists.


marcuskohlberg:
Thanks!

PR #1970, opened 2025-06-17
tsparser: usage of default exported resources

fredr:
the resolved bind of a default export will have no name, but we still need to add it or it wont be tracked

### DIFF
--- a/tsparser/src/parser/resources/apis/service.rs
+++ b/tsparser/src/parser/resources/apis/service.rs
@@ -86,7 +86,6 @@ struct ServiceLiteral {
     pub doc_comment: Option<String>,
     pub resource_name: String,
     pub config: Option<DecodedServiceConfig>,
-    pub expr: ast::NewExpr,
 }
 
 impl ReferenceParser for ServiceLiteral {
@@ -130,7 +129,6 @@ impl ReferenceParser for ServiceLiteral {
                     doc_comment,
                     resource_name: resource_name.to_string(),
                     config,
-                    expr: (*expr).to_owned(),
                 }));
             }
         }


tsparser/src/parser/usageparser/mod.rs
```diff
@@ -186,13 +186,11 @@ impl<'a> UsageResolver<'a> {
                         };
 
                         for bind in resolved_binds.into_iter().flatten() {
-                            if let Some(name) = &bind.name {
```
eandre, line 189: I think we should map the `None` case to `"default"`, as I believe that is how JavaScript models default exports.

fredr: :heavy_check_mark:

tsparser/src/parser/usageparser/mod.rs
```diff
@@ -186,13 +186,11 @@ impl<'a> UsageResolver<'a> {
...
-                                external.push(BindToScan {
-                                    bound_name: local_name.to_id(),
-                                    selector: Some(name),
-                                    bind: bind.to_owned(),
-                                });
-                            }
+                            external.push(BindToScan {
+                                bound_name: local_name.to_id(),
+                                selector: Some(bind.name.as_deref().unwrap_or("default")),
```
eandre: I think we need to handle the `"default"` case earlier. There are multiple reasons why there might not be a name, like `const _ = new Subscription(...)`.

fredr: ah right right, I'll fix

tsparser/src/parser/resources/apis/gateway.rs
```diff
@@ -39,7 +39,7 @@ pub const GATEWAY_PARSER: ResourceParser = ResourceParser {
         for r in iter_references::<Res>(&module, &names) {
             let r = report_and_continue!(r);
 
-            let object = match &r.bind_name {
+            let object = match r.bind_name.ident() {
                 None => None,
```
eandre: I guess this could resolve the default export, but not sure how difficult that would be to do. If it's tricky maybe add a TODO here?

tsparser/src/parser/resources/infra/cron.rs
```diff
@@ -61,7 +61,7 @@ fn parse_cron_job(
     pass: &mut ResourceParseContext,
     r: NamedClassResource<DecodedCronJobConfig>,
 ) -> ParseResult<()> {
-    let object = match &r.bind_name {
+    let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/objects.rs
```diff
@@ -46,7 +46,7 @@ pub const OBJECTS_PARSER: ResourceParser = ResourceParser {
                 let r = report_and_continue!(r);
                 let cfg = r.config.unwrap_or_default();
 
-                let object = match &r.bind_name {
+                let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/objects.rs
```diff
@@ -67,15 +67,14 @@ pub const OBJECTS_PARSER: ResourceParser = ResourceParser {
...
-                    is_default_export: r.is_default_export,
                 });
             }
         }
 
         {
             for r in iter_references::<NamedStaticMethod>(&module, &names) {
                 let r = report_and_continue!(r);
-                let object = match &r.bind_name {
+                let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/pubsub_subscription.rs
```diff
@@ -71,7 +71,7 @@ pub const SUBSCRIPTION_PARSER: ResourceParser = ResourceParser {
                 spread.err("cannot use ... for PubSub topic reference");
                 continue;
             }
-            let object = match &r.bind_name {
+            let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/pubsub_topic.rs
```diff
@@ -63,7 +63,7 @@ pub const TOPIC_PARSER: ResourceParser = ResourceParser {
 
         for r in iter_references::<PubSubTopicDefinition>(&module, &names) {
             let r = report_and_continue!(r);
-            let object = match &r.bind_name {
+            let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/sqldb.rs
```diff
@@ -152,7 +152,7 @@ pub const SQLDB_PARSER: ResourceParser = ResourceParser {
                     }
                 };
 
-                let object = match &r.bind_name {
+                let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/infra/sqldb.rs
```diff
@@ -172,15 +172,14 @@ pub const SQLDB_PARSER: ResourceParser = ResourceParser {
...
-                    is_default_export: r.is_default_export,
                 });
             }
         }
 
         {
             for r in iter_references::<NamedStaticMethod>(&module, &names) {
                 let r = report_and_continue!(r);
-                let object = match &r.bind_name {
+                let object = match r.bind_name.ident() {
```
eandre: Same here

tsparser/src/parser/resources/apis/gateway.rs
fredr: I'm not entirely sure how object is used, but I've given it a stab, it resolves to something, not sure exactly what we want it to resolve to?

tsparser/src/parser/resources/infra/pubsub_topic.rs
```diff
@@ -63,9 +63,12 @@ pub const TOPIC_PARSER: ResourceParser = ResourceParser {
 
         for r in iter_references::<PubSubTopicDefinition>(&module, &names) {
             let r = report_and_continue!(r);
-            let object = match r.bind_name.ident() {
-                None => None,
-                Some(id) => pass
+            let object = match r.bind_name {
+                BindName::Anonymous => None,
+                BindName::DefaultExport(ref expr) => {
```
eandre: I don't think we should be resolving this by passing in an expression to `resolve_obj`.

`resolve_obj` goes into `self.expr(...)` which goes into a big match statement where we end up at:

```rs
            ast::Expr::Ident(expr) => {
                let Some(obj) = self.ident_obj(expr) else {
                    HANDLER.with(|handler| handler.span_err(expr.span, "unknown identifier"));
                    return Type::Basic(Basic::Never);
                };

                let named = Named::new(obj, vec![]);
                Type::Named(named)
            }
```

This goes into `ident_obj(expr)`:

```rs
    fn ident_obj(&self, ident: &ast::Ident) -> Option<Rc<Object>> {
        // Does this represent a type parameter?
        self.state.resolve_module_ident(self.module, ident)
    }
}
```

And `resolve_module_ident` checks:
```rs
        // Is it a top-level object in this module?
        let ast_id = AstId::from(ident);
        if let Some(obj) = module.data.top_level.get(&ast_id) {
            return Some(obj.clone());
        }
```

I believe the correct way to resolve default exports is to do the same thing, but check:
```rs
        if let Some(obj) = module.data.default_export {
            return Some(obj.clone());
        }
```

(sorry for the trouble)

tsparser/src/parser/resources/apis/gateway.rs
fredr: I've moved all these to a separate function and are now resolving it via `module.data`. The object will still resolve to `None` for default export expressions, I've added a todo for that for now, maybe that is fine to start with?

eandre (approved):

PR #1977, opened 2025-06-19
Detect duplicate migrations in metadata migrator

ekerfelt:


cli/daemon/sqldb/migrate.go
```diff
@@ -80,17 +80,30 @@ var _ io.ReadCloser = (*multiReadCloser)(nil)
...
 	}
+	src.validate()
+	return src
+}
+
+func (src *MetadataSource) validate() {
+	seen := make(map[uint64]struct{})
+	for _, m := range src.migrations {
+		if _, ok := seen[m.Number]; ok {
+			src.err = fmt.Errorf("duplicate migration identifier %q", m.Filename)
```
eandre, line 98: Maybe only set this if it's nil? That way we report the first error, not the last

ekerfelt: 👍

PR #1987, opened 2025-07-01
Try to mitigate cue race bug

ekerfelt:


cli/daemon/secret/secret.go
```diff
@@ -79,6 +81,8 @@ func (mgr *Manager) Load(app *apps.Instance) *LoadResult {
 func (lr *LoadResult) Get(ctx context.Context, expSet *experiments.Set) (data *Data, err error) {
 	defer func() {
 		if err == nil {
+			lr.localSecretMu.Lock()
```
eandre, line 86: Add a comment referencing the bug

PR #1997, opened 2025-07-08
Fix type resolution with nested generic type parameters

fredr:
Previous to this change `MyOmit` in the test data below would resolve to the same thing as `MyPick`, resulting in the invalid type in certain situations.


fredr:
Fixed in https://github.com/encoredev/encore/pull/2014

PR #2001, opened 2025-07-10
runtimes/core: write logs in batches

fredr:


runtimes/core/src/log/writers.rs
```diff
@@ -53,25 +54,63 @@ impl ActorWriter {
...
-    fn write_with_retry<W: Write>(writer: &mut W, bytes: &[u8]) {
+    fn recv_batch(recv: &Receiver<Vec<u8>>) -> Result<Vec<Vec<u8>>, RecvError> {
+        const MAX_BATCH_SIZE: usize = 256;
+        const TIMEOUT: Duration = Duration::from_millis(10);
+
+        let mut bufs = Vec::with_capacity(MAX_BATCH_SIZE);
+        // wait for a log message
+        bufs.push(recv.recv()?);
+
+        let instant = Instant::now();
```
eandre: I was thinking instead of receiving with a timeout we could just run try_recv() in a loop until it fails (up to a max of N messages).

fredr: ah thats nicer, updated

runtimes/core/src/log/writers.rs
```diff
@@ -53,25 +54,63 @@ impl ActorWriter {
...
+                Self::write_batch_with_retry(&mut writer, &bytes);
             }
         });
         Self { sender }
     }
 
-    fn write_with_retry<W: Write>(writer: &mut W, bytes: &[u8]) {
+    fn recv_batch(recv: &Receiver<Vec<u8>>) -> Result<Vec<Vec<u8>>, RecvError> {
+        const MAX_BATCH_SIZE: usize = 256;
+        let mut bufs = Vec::with_capacity(MAX_BATCH_SIZE);
```
eandre: I would make `Vec::new()` instead; the most common scenario is that we'll get a single log message so this will over-allocate

fredr: sure! it will then ofc cause a lot of re-allocations for apps with large request volumes, but probably not a problem. We could do something smarter in the future if we want also, by re-using the buffer or something

PR #2020, opened 2025-07-29
Update insert-test-data-db.md

rmcsharry:
Make it clear that operations using the db at startup must still happen inside a service


marcuskohlberg (dismissed):

marcuskohlberg:
Thanks!

PR #2045, opened 2025-08-19
runtimes/core+js: add support for setting custom http status codes

fredr:
This pr depends on #2042 (it needs to be merged first), it implements the same thing but for the core and js runtime.

Example of what the api looks like:
```ts
import { HttpStatus, api } from "encore.dev/api";

export const get = api(
  { expose: true, method: "GET", path: "/hello/:name" },
  async ({ name }: { name: string }): Promise<Response> => {
    const msg = `Hello ${name}!`;
    return { message: msg, status: 201 };
  },
);

interface Response {
  message: string;
  status: HttpStatus;
}

```


tsparser/src/parser/resources/apis/encoding.rs
```diff
@@ -118,6 +121,9 @@ impl RequestEncoding {
                 ParamData::Query { .. } => by_loc.query.push(p),
                 ParamData::Body => by_loc.body.push(p),
                 ParamData::Cookie => by_loc.cookie.push(p),
+                ParamData::HTTPStatus => {
+                    // HTTPStatus can only be set in responses
```
eandre: How does this behave if you try to use it in a request?

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -19,6 +19,71 @@ export type Method =
   | "TRACE"
   | "CONNECT";
 
+export type HttpStatus =
```
eandre: Can we move this to the bottom of the file? :)

tsparser/src/parser/resources/apis/encoding.rs
```diff
@@ -86,6 +87,8 @@ pub struct RequestEncoding {
 pub struct ResponseEncoding {
     /// Parsed params.
     pub params: Vec<Param>,
+    /// HTTP status param
+    pub http_status: Option<Param>,
```
eandre: Why isn't this handled by `params`? It seems like it should be able to be?

runtimes/js/encore.dev/api/mod.ts
fredr: yeah sure! I was contemplating moving it to a separate file even

tsparser/src/parser/resources/apis/encoding.rs
fredr: it was just because it was annoying to have to deal with a vec, since it can only be one, but I can change to params and add some helpers for dealing with that

tsparser/src/parser/resources/apis/encoding.rs
fredr: it is currently ignored :thinking: not sure how we want to handle it?

fredr: I guess it should work like any other type, will update

fredr: its now handled like any other field if the encoding doesnt support http statuses (so it will use the default location)

runtimes/js/encore.dev/api/mod.ts
fredr: I moved it to a separate file so that where I defined all codes with their names, so this will work:

```ts
export const get = api(
  { expose: true, method: "GET", path: "/hello/:name" },
  async (req: Request): Promise<Response> => {
    // ...
    return { message: msg, status: HttpStatus.Created };
  },
);
interface Response {
  message: string;
  status: HttpStatus;
}

```

wdyt?

PR #2048, opened 2025-08-20
docs: add docs for custom http statuses

fredr:
Docs for https://github.com/encoredev/encore/pull/2042 and https://github.com/encoredev/encore/pull/2045

### DIFF
--- a/docs/ts/primitives/defining-apis.mdx
+++ b/docs/ts/primitives/defining-apis.mdx
@@ -216,6 +216,30 @@ export const fallback = api.raw(
 )
 ```
 
+## Custom HTTP status codes
+
+By default, Encore automatically sets appropriate HTTP status codes for your API responses. We recommend using these default status codes, but there are situations where you might need to set a custom HTTP status code, such as when porting an existing API that clients depend on for specific status codes.
+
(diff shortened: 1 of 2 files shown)


eandre (approved):

docs/ts/primitives/defining-apis.mdx
```diff
@@ -216,6 +216,30 @@ export const fallback = api.raw(
...
+
+interface Response {
+  msg: string;
+  status: HttpStatus;
+}
+
+export const endpoint = api(
+  { expose: true, method: "GET", path: "/path" },
+  async (): Promise<Response> => {
+    return { msg: "Hello", status: HttpStatus.Created };
```
shimonenator, line 236: @fredr what would the actual API response look like? 

```
{
  "msg": "Hello"
}
```

What if I wanted to have `status` field in the API response that's unrelated to the http status, would I be able to?

fredr: yes exactly, this type works the same way the Header, Query and Cookie types work, so the http status will not be part of the response body. 

If you want to have a http status code in your response body you can just use a different type for that, e.g your own HttpStatus type that you define yourself, or just number

PR #2055, opened 2025-09-08
runtimes/core: add support for pgvector type

fredr:


runtimes/core/src/sqldb/val.rs
```diff
@@ -253,6 +261,9 @@ impl ToSql for PValue {
     }
 
     fn accepts(ty: &Type) -> bool {
+        if is_pgvector(ty) {
```
eandre: Maybe move this to the bottom? Seems unnecessary to check for this before the others that are more common

runtimes/core/src/sqldb/val.rs
```diff
@@ -410,6 +436,9 @@ impl<'a> FromSql<'a> for PValue {
...
```
eandre: Same

runtimes/core/src/sqldb/val.rs
```diff
@@ -392,7 +403,22 @@ impl<'a> FromSql<'a> for PValue {
             }
 
             _ => {
-                if let Kind::Array(_) = ty.kind() {
+                if is_pgvector(ty) {
```
eandre: Same

runtimes/core/src/sqldb/val.rs
```diff
@@ -116,7 +120,11 @@ impl ToSql for PValue {
                         val.to_sql(ty, out)
                     }
                     _ => {
-                        if let Kind::Enum(_) = ty.kind() {
+                        if is_pgvector(ty) {
```
eandre: Can this be after the Enum case?

fredr: ah, sorry, thought I fixed them all, fixed

eandre (approved):

PR #2067, opened 2025-09-15
tsparser: validate sqldb query syntax

eandre:


tsparser/src/parser/resources/infra/sqldb.rs
```diff
@@ -389,3 +408,17 @@ pub struct AccessDatabaseUsage {
...
+    let mut query = String::new();
+    for (i, q) in tpl.tpl.quasis.iter().enumerate() {
+        query.push_str(&q.raw);
+        if !q.tail {
+            query.push_str(&format!("${}", i + 1));
+        }
+    }
+
+    let res = pg_query::parse(&query);
+    log::info!("Parsed query: {} -> {:?}", query, res);
```
fredr: was this for debugging? just checking in case you didnt mean to commit it

eandre: Good catch, fixed

PR #2083, opened 2025-09-24
Add support for arbitrary precition numbers

fredr:


runtimes/core/src/sqldb/val.rs
```diff
@@ -119,6 +122,11 @@ impl ToSql for PValue {
                         let val = cidr::IpInet::from_str(str).context("unable to parse inet")?;
                         val.to_sql(ty, out)
                     }
+                    Type::NUMERIC => {
+                        let n = Numeric::from_str(str)?;
+                        numeric_to_sql(n, out);
```
eandre: Why not `n.to_sql(ty, out)` like how other types are implemented?

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -543,3 +543,15 @@ export type HttpStatus =
...
+(runtime.Decimal.prototype as any)[Symbol.toPrimitive] = function (
+  hint: string
+) {
+  if (hint === "number") {
+    return +this.value;
+  }
+
+  return this.value;
+};
+export class Decimal extends runtime.Decimal {}
```
eandre: For all other types we've deliberately hidden the internal implementation and designed our own, hand-written public API. It makes the API more discoverable for end-users and reduces the risk of accidentally changing the API without noticing. I think we should do the same here

runtimes/js/src/runtime.rs
```diff
@@ -388,3 +389,38 @@ impl From<api::Error> for APICallError {
         }
     }
 }
+
+#[napi]
+pub struct Decimal {
+    value: String,
```
eandre: Why not store the value as an already-parsed `encore_runtime_core::api::Decimal`?

runtimes/js/encore.dev/api/mod.ts
```diff
@@ -543,3 +543,15 @@ export type HttpStatus =
 export { APIError, ErrCode } from "./error";
 export { Gateway, type GatewayConfig } from "./gateway";
 export { IterableSocket, IterableStream, Sink } from "./stream";
+
```
eandre: I don't think we should put the Decimal types in `encore.dev/api`; it's not specific to APIs in general. For Encore.go we have `encore.dev/types/uuid` as a package, so maybe this should live in `encore.dev/types/decimal`?

tsparser/src/parser/types/type_resolve.rs
```diff
@@ -718,6 +719,13 @@ impl Ctx<'_> {
             }
         }
 
+        // Is this a encore Decimal?
+        if obj.name.as_ref().is_some_and(|s| s == "Decimal")
+            && self.state.is_module_path(obj.module_id, "encore.dev/api")
```
eandre: Update this to the final import path

proto/encore/parser/schema/v1/schema.proto
```diff
@@ -253,4 +253,6 @@ enum Builtin {
 
   INT     = 18;
   UINT    = 19;
+  
+  DECIMAL = 20;
```
eandre, line 257: I guess if we add it here we need to consider how to generate clients etc for these types

runtimes/core/src/sqldb/val.rs
fredr: Numeric here is the postgres protocol type that represents how the type is stored in postgres and not the "resulting" type which in this case would be our own Decimal type, so to keep it similar to that I've now implemeted to/from sql for Decimal. Wdyt?

runtimes/js/encore.dev/api/mod.ts
fredr: wrapped the runtime type in a plain class now, was that how you meant?

runtimes/js/src/runtime.rs
fredr: yeah, sounds good, changed

runtimes/js/encore.dev/types/mod.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A decimal type that can hold values with arbitrary precision.
+ * Unlike JavaScript's native number type, this can accurately represent
+ * decimal values without floating-point precision errors.
+ */
+export class Decimal {
+  private impl: runtime.Decimal;
+
+  constructor(value: string) {
+    this.impl = new runtime.Decimal(value);
```
eandre: Should we support numbers for the constructor too?

runtimes/js/encore.dev/types/mod.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  toJSON(): string {
+    return this.impl.toJSON();
+  }
+
+  get value(): string {
+    return this.impl.value;
+  }
+
+  get __encore_decimal(): boolean {
```
eandre: Can this be made private? It should still be accessible to Rust

runtimes/js/src/runtime.rs
```diff
@@ -388,3 +389,37 @@ impl From<api::Error> for APICallError {
...
+    pub fn to_json(&self) -> String {
+        self.inner.to_string()
+    }
+
+    #[napi(getter)]
+    pub fn value(&self) -> String {
+        self.inner.to_string()
+    }
+
+    #[napi(getter, js_name = "__encore_decimal")]
```
eandre: This is presumably unnecessary now?

runtimes/js/src/runtime.rs
```diff
@@ -388,3 +389,37 @@ impl From<api::Error> for APICallError {
...
+
+    #[napi(js_name = "toJSON")]
+    pub fn to_json(&self) -> String {
+        self.inner.to_string()
+    }
+
+    #[napi(getter)]
+    pub fn value(&self) -> String {
+        self.inner.to_string()
+    }
```
eandre: Instead of these two should we just have a single `to_string` method?

proto/encore/parser/schema/v1/schema.proto
fredr: I updated the clients to handle it as string. I think that is easiest since we dont want to add dependencies.

runtimes/js/src/runtime.rs
fredr: both this one and the type defined in typescript must be equal, otherwise code will behave differently if the value comes from user code or from rust (e.g if we return a decimal from a db query or if a user created it) or do we have some other way of converting the rust type to the "typescript defined" type?

eandre: Ah, hmm. I don't like the idea of supporting both interchangeably. That seems very error-prone. I guess one way could be to register the constructor with the runtime when it starts up?

fredr: yeah, thats why I used the rust type directly and just had a extended version in js earlier.

hmm, register the constructor at startup, and then create an instance of that instead of the runtime type from the rust side? I can give that a try

eandre: Yeah, I understand now why you did it that way :) try registering the constructor and see if that works

fredr: added a way of registering the constructors, let me know what you think of the api/implementation

runtimes/js/encore.dev/internal/runtime/mod.ts
```diff
@@ -5,5 +5,5 @@ export * from "./napi/napi.cjs";
 const testMode = process.env.NODE_ENV === "test";
 
 export const RT = new Runtime({
-  testMode,
+  testMode
```
eandre: Could we register the type constructor here instead? Then we don't need to worry about the case where the constructor hasn't been registered?

fredr: yes! I've added it the runtime constructor now

runtimes/js/encore.dev/types/mod.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A decimal type that can hold values with arbitrary precision.
+ * Unlike JavaScript's native number type, this can accurately represent
+ * decimal values without floating-point precision errors.
+ */
+export class Decimal {
+  private impl: runtime.Decimal;
+
+  constructor(value: string | number) {
```
eandre: Should we also support bigint here?

fredr: yes! added

eandre (approved):

PR #2094, opened 2025-10-03
runtimes/js: add per-env type constructors

eandre:


runtimes/js/src/napi_util.rs
```diff
@@ -1,4 +1,6 @@
 use napi::{Either, Env, JsFunction, JsObject, JsUnknown};
+use std::collections::HashMap;
```
fredr:
```suggestion
```

This one is unused

fredr:
Nice! Tested and it works well :rocket:

fredr (approved):

PR #2106, opened 2025-10-10
Add encore secret check command to validate environment secrets before deployment

Mugesh2024rj:
This PR introduces a new CLI command encore secret check [envs...] that helps detect missing secrets across specified environments (e.g., prod, dev) before deployment.

The main goal is to prevent late feedback loops caused by missing secrets during deployment — for example, when a secret is set in the development environment but not in production.

By integrating this command into GitHub Workflows, it becomes possible to automatically verify that all required secrets are configured for each environment before merging a PR or triggering a deployment.

added new CLI command:
encore secret check [envs...]

Example Usage:
encore secret check prod dev

Output:

Secret Key     Production   Development
MySecret       ✗            ✓

Error: There is 1 secret missing.

Successfully updated secret value for MySecretKey.

C:\Users\muges\myfolder\devday\encore\encore>encore secret list
Secret Key    Production   Development   Local   Preview   Specific Envs   
MySecretKey   ✓            ✓             ✓       ✓

successfully solved
image


fredr:
This pr contains 250 commits, I think there might be something wrong, you probably want to rebase on latest main

PR #2109, opened 2025-10-10
runtimes/core: Add metrics support

fredr:
Adds the ground work for adding metrics to encore.ts, adds the e_requests_total counter for apps running in gcp/encore cloud that export metrics to gcp cloud monitoring. 

~Will add support for aws, datadog and prometheus in a follow up.~
~This also doesnt expose any sys metrics, will add that in a follow up pr as well.~

### DIFF
--- a/runtimes/core/src/lib.rs
+++ b/runtimes/core/src/lib.rs
@@ -342,12 +342,7 @@ impl Runtime {
             .collect::<Result<HashMap<_, _>, anyhow::Error>>()
             .context("failed to resolve gateway push subscriptions")?;
 
-        let pubsub = pubsub::Manager::new(
-            tracer.clone(),
-            resources.pubsub_clusters,
-            &md,
-            Arc::clone(metrics_manager.registry()),
-        )?;
+        let pubsub = pubsub::Manager::new(tracer.clone(), resources.pubsub_clusters, &md)?;
         let objects =
             objects::Manager::new(&secrets, tracer.clone(), resources.bucket_clusters, &md);
         let sqldb = sqldb::ManagerConfig {
--- a/runtimes/core/src/pubsub/manager.rs
+++ b/runtimes/core/src/pubsub/manager.rs
@@ -15,7 +15,6 @@ use crate::encore::parser::meta::v1 as meta;
 use crate::encore::parser::schema::v1 as schema;
 use crate::encore::runtime::v1 as pb;
 use crate::log::LogFromRust;
-use crate::metrics::{self, counter};
 use crate::model::{PubSubRequestData, RequestData, ResponseData, SpanId, SpanKey, TraceId};
 use crate::names::EncoreName;
 use crate::pubsub::noop::NoopCluster;
@@ -37,7 +36,6 @@ pub struct Manager {
     topics: Arc<RwLock<HashMap<EncoreName, Arc<TopicInner>>>>,
     subs: Arc<RwLock<HashMap<SubName, Arc<SubscriptionObj>>>>,
     push_registry: PushHandlerRegistry,
-    metrics_registry: Arc<metrics::Registry>,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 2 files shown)


fredr:
actually, I will refactor this a bit since metrics-rs isn't flexible enough with the different types that we support. Drafting this for now

fredr:
refactored and added system metrics to this ps

runtimes/core/src/metrics/exporter/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod gcp;
+pub use gcp::Gcp;
+
+#[async_trait::async_trait]
```
eandre, line 10: Do we still need this? I think newer versions of Rust supported async traits natively for some use cases

runtimes/core/src/metrics/exporter/gcp.rs
```diff
@@ -0,0 +1,222 @@
...
+                        metric.key.name()
+                    );
+                    continue;
+                }
+            };
+
+            // Pre-allocate exact capacity for labels
+            let mut labels: HashMap<String, String> =
+                HashMap::with_capacity(container_labels_len + metric.key.labels().len());
+            labels.extend(container_labels.clone());
```
eandre: Unnecessary `.clone()`?

runtimes/core/src/metrics/registry.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct CollectedMetric {
+    pub key: Key,
+    pub value: MetricValue,
+    pub registered_at: SystemTime,
+}
+
+#[derive(Debug, Clone)]
```
eandre: It's a bit unexpected to me that the registry itself is clonable. Would it be better to instead use `Arc<Registry>` in callers?

runtimes/core/src/metrics/manager.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn create_exporter(
+        &self,
+        env: &Environment,
+        http_client: &reqwest::Client,
+        runtime_handle: tokio::runtime::Handle,
+    ) -> Arc<dyn Exporter + Send + Sync> {
+        match self {
+            Self::Gcp(config) | Self::EncoreCloud(config) => {
+                runtime_handle.block_on(Self::create_gcp_exporter(config, env, http_client))
```
eandre: A bit annoying to need to block here. I guess the only reason we need to do that is because the container metadata needs to be collected? Can we do that lazily/in the background, similar to the lazy gcp client?

runtimes/core/src/lib.rs
```diff
@@ -330,7 +341,12 @@ impl Runtime {
             .collect::<Result<HashMap<_, _>, anyhow::Error>>()
             .context("failed to resolve gateway push subscriptions")?;
 
-        let pubsub = pubsub::Manager::new(tracer.clone(), resources.pubsub_clusters, &md)?;
+        let pubsub = pubsub::Manager::new(
+            tracer.clone(),
+            resources.pubsub_clusters,
+            &md,
+            Arc::new(metrics_manager.registry().clone()),
```
eandre: For example, wouldn't it be easier to have `metrics_manager.registry()` return `Arc<Registry>` instead of this?

runtimes/core/src/metrics/exporter/mod.rs
fredr: it seems we can use dynamic dispatch if we use the stabilized one, so cant store it as a `dyn Exporter`. We could change it from being a trait to a enum and that way avoid using async_trait, but its a bit less ergonomic. Wdyt?

runtimes/core/src/metrics/manager.rs
fredr: I refactored to using a lazy cached client and avoiding the block_on by fetching the container data in the collect step

runtimes/core/src/metrics/exporter/mod.rs
eandre: Probably fine to keep, in that case

runtimes/core/src/metrics/registry.rs
fredr: yeah, that is nicer, updated to that now

fredr:
updated the pr with the other exporters as well

PR #2111, opened 2025-10-12
feat: enable []string as a valid set-cookie type, fixes #1548

### LINKED ISSUE #1548: Proposal: Adding multiple values for Set-Cookie and allowing *http.Cookie type
Hey, recently I came across an interesting problem.

One of the endpoints in my service needs to return two or more Set-Cookie cookies at the same time.
Currently there is no way to create multiple Set-Cookies at the same time, and any attempt to change the type to an array will raise an error such as: 
`unsupported type in header: []string`

I have two suggestions:
1. Let's add the possibility to apply an array type for the Set-Cookie type header
2. Allow for passing *http.Cookie (and the array) the same as the request comes with *http.Cookie, this would unify the types coming in and out

[Example response from documentation](https://encore.dev/docs/primitives/apis#cookies)
```go
type LoginResponse struct {
	SessionID string `header:"Set-Cookie"`
}
```

[Example request from documentation](https://encore.dev/docs/develop/auth#accepting-structured-auth-information)
```go
type MyAuthParams struct {
	SessionCookie *http.Cookie `cookie:"session"`
}
```

Example of how I would see the proposal to be integrated:
```go
type Response struct {
	// example param with `header:"Set-Cookie"` as a string
	Cookie string `header:"Set-Cookie"`
	
	// example param with `header:"Set-Cookie"` as a []string
	Cookies []string `header:"Set-Cookie"`
	
	// example param with `header:"Set-Cookie"` as a *http.Cookie
	Cookie *http.Cookie `header:"Set-Cookie"`

        // example param with `header:"Set-Cookie"` as a []*http.Cookie
	 Cookies []*http.Cookie `header:"Set-Cookie"`
}
```
(obviously only one parameter with `header:"Set-Cookie"`, just an example of multiple different types)

TheGB0077:
Reasoning: Setting multiple Cookies as part of one request is often needed, but current limitations on encore go don't allow this without using raw api endpoints. This PR adds []string as a valid type for the struct tag "Set-Cookie" in order to provide a clean api for setting up multiple Cookies from a list.

Out of scope on this PR: Adding *http.Cookie type to allow list

fixes #1548

### DIFF
--- a/v2/codegen/apigen/endpointgen/testdata/set_cookie_array.txt
+++ b/v2/codegen/apigen/endpointgen/testdata/set_cookie_array.txt
@@ -101,7 +101,10 @@ var EncoreInternal_api_APIDesc_TestCookies = &__api.Desc[*EncoreInternal_TestCoo
 		if resp != nil {
 
 			// Encode headers
-			headers = map[string][]string{"x-other": []string{__etype.MarshalOne(__etype.MarshalString, resp.OtherHeader)}}
+			headers = map[string][]string{
+				"set-cookie": __etype.MarshalList(__etype.MarshalString, resp.Cookies),
+				"x-other":    []string{__etype.MarshalOne(__etype.MarshalString, resp.OtherHeader)},
+			}
 		}
 
 		// Set response headers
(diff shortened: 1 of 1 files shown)


TheGB0077:
Closing PR, will implement this on my personal fork.

Lenstack:
i need the same

eandre:
I believe this is already supported on the latest release of Encore.

Lenstack:
> I believe this is already supported on the latest release of Encore.

nice i will to try thanks

PR #2125, opened 2025-10-27
Implement TOUCH for nsq subs

ekerfelt:


runtimes/core/src/pubsub/nsq/sub.rs
```diff
@@ -107,7 +107,40 @@ impl Subscription for NsqSubscription {
...
+
+        // Return the message so we can finish or requeue it
+        msg
+    });
+
+    let result = handle_message(body, timestamp, attempt, handler).await;
+
+    // Signal the touch task to stop and return the message
+    let _ = stop_tx.send(());
+    let msg = touch_handle.await.expect("touch task panicked");
```
fredr: this will cause a panic (expect is pretty much same as `unwrap` just that is has a description). Maybe better to just log something on error here?

runtimes/core/src/pubsub/nsq/sub.rs
```diff
@@ -107,7 +107,40 @@ impl Subscription for NsqSubscription {
...
+    let timestamp = msg.timestamp;
+    let attempt = msg.attempt;
+
+    // Create a channel to signal when to stop sending touch messages
+    let (stop_tx, mut stop_rx) = tokio::sync::oneshot::channel::<()>();
+
+    // Spawn a background task to send touch messages every 30 seconds
+    let touch_handle = tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_secs(30));
+        // Skip the first tick (immediate)
```
fredr, line 119: is this true? 🤔 wont this cause a 30 sec wait? so the first touch will be sent after 1 min? or am I mistaken?

fredr: ah I was mistaken!

ekerfelt: /// Creates new [`Interval`] that yields with interval of `period`. The first
/// tick completes immediately.

PR #2128, opened 2025-10-30
runtimes/js: add support for custom metrics

fredr:
Adds support for creating custom metrics in typescript apps. Depends on #2109

Inspired by the implementation in the go runtime.
Metrics needs to be defined within a service, and will automatically add the service label.

This implementation is backed by a SharedArrayBuffer, where each u64 slot is a metric value, created from js by the main thread. From js we do atomic operation to update values on this buffer, and from rust we do atomic load when collecting the metrics. The rust runtime is use to coordinate what slot should be used for a label set, so that we can support multiple workers updating the same metric.

### DIFF
--- a/docs/ts/observability/metrics.md
+++ b/docs/ts/observability/metrics.md
@@ -183,6 +183,32 @@ function processJob(jobType: string, priority: number) {
 }
 ```
 
+## Metric references
+
+Encore uses static analysis to determine which services are using each metric, and what operations each service is performing.
+
+This means metric objects can't be passed around however you like, as it makes static analysis impossible in many cases. To simplify your workflow, given these restrictions, Encore supports defining a "reference" to a metric that can be passed around any way you want.
+
+To create a reference, call the `.ref()` method on any metric:
+
+```typescript
+import { Counter } from "encore.dev/metrics";
+
+export const ordersProcessed = new Counter("orders_processed");
+
+// Create a reference that can be passed around
+const metricRef = ordersProcessed.ref();
+
+// Pass the reference to other functions
+function logMetric(metric: Counter) {
+    metric.increment();
+}
+
+logMetric(metricRef);
+```
+
+This works for all metric types (`Counter`, `CounterGroup`, `Gauge`, and `GaugeGroup`).
+
 <Callout type="important">
 
 Each combination of label values creates a unique time series tracked in memory and stored by the monitoring system.


runtimes/js/encore.dev/internal/appinit/mod.ts
```diff
@@ -28,16 +33,26 @@ export function registerGateways(gateways: Gateway[]) {
 
 export async function run(entrypoint: string) {
   if (isMainThread) {
+    const metricsBuffer = __internalInitGlobalMetricsBuffer();
     const extraWorkers = runtime.RT.numWorkerThreads() - 1;
     if (extraWorkers > 0) {
+      log.debug(`Starting ${extraWorkers} worker threads`);
```
eandre: Intended? Remove?

runtimes/js/encore.dev/metrics/mod.ts
```diff
@@ -0,0 +1,466 @@
...
+ * ```typescript
+ * import { CounterGroup } from 'encore.dev/metrics';
+ *
+ * interface Labels {
+ *   success: boolean;
+ * }
+ *
+ * export const OrdersProcessed = new CounterGroup<Labels>("orders_processed");
+ *
+ * OrdersProcessed.with({ success: true }).increment();
```
eandre:
```suggestion
 * export const ordersProcessed = new CounterGroup<Labels>("orders_processed");
 *
 * ordersProcessed.with({ success: true }).increment();
```

runtimes/js/encore.dev/metrics/mod.ts
```diff
@@ -0,0 +1,466 @@
...
+ * This module provides counters and gauges that can be statically analyzed
+ * by the Encore compiler and automatically exported to observability backends.
+ *
+ * @example Simple counter
+ * ```typescript
+ * import { Counter } from 'encore.dev/metrics';
+ *
+ * export const OrdersProcessed = new Counter("orders_processed");
+ *
+ * OrdersProcessed.increment();
```
eandre:
```suggestion
 * export const ordersProcessed = new Counter("orders_processed");
 *
 * ordersProcessed.increment();
```

runtimes/js/encore.dev/metrics/mod.ts
```diff
@@ -0,0 +1,466 @@
...
+let globalBuffer: SharedArrayBuffer | undefined;
+let initErrorLogged = false;
+
+const testMode = process.env.NODE_ENV === "test";
+
+/**
+ * Called during encores initialization
+ * @internal
+ */
+export function __internalSetGlobalMetricsBuffer(
```
eandre: Do we need to export this here? Could we put it in an internal/ folder?

runtimes/js/encore.dev/metrics/mod.ts
```diff
@@ -0,0 +1,466 @@
...
+  private get buffer(): SharedArrayBuffer | undefined {
+    return getBuffer();
+  }
+
+  /**
+   * Get a counter for the given label values.
+   *
+   * Note: Number values in labels are converted to integers using Math.floor().
+   */
+  with(labels: L): AtomicCounter | NoOpCounter {
```
eandre: Do we need this return type? Or could we always return a `Counter` so the user doesn't need to care about the difference? For other resources we've separated the underlying implementation into an `impl` field so we could do like:

```typescript
const impl = new NoOpCounterImpl();
return new Counter(impl);
```

or something like that. Maybe a bit tricky to do since it would require a private constructor?

runtimes/js/encore.dev/internal/appinit/mod.ts
fredr: I added it for debugging, but thought it was quite nice to have it, so its possible to see how many workers where actually spun up. But I can remove it

runtimes/js/encore.dev/metrics/mod.ts
fredr: moved to internal! needs to be exported so it can be accessed from appinit, if you dont have a better suggestion for how to set it up?

runtimes/js/encore.dev/metrics/mod.ts
fredr: yeah, makes sense to not introduce these extra types to the public api, I refactored a bit, let me know what you think and I'll do an extra round of testing

eandre (dismissed):

PR #2138, opened 2025-11-06
Update CONTRIBUTING.md to refer to the `encore` official binary

eleijonmarck:
I found that it would be better to note that the users should use the official `encore` binary for building encore


CONTRIBUTING.md
```diff
@@ -49,6 +49,8 @@ To do that, run `encore daemon env` and grab the value of `ENCORE_GOROOT`. For e
 export ENCORE_GOROOT=/opt/homebrew/Cellar/encore/0.16.2/libexec/encore-go
 ```
 
+Note: It is easiest to use the office `encore` binary.
```
eandre, line 52: Isn't this already listed?

> Unless you want to make changes to the Go runtime it's easiest to point this to an existing Encore installation.

Maybe worth clarifying that sentence instead?

Issue #2143, opened 2025-11-12
Encore fails to build Docker image on Docker v29.0.0


shimonenator:
Encore fails to build Docker image on the newly released [Docker Engine v29.0.0](https://github.com/moby/moby/releases/tag/docker-v29.0.0) with `MinAPIVersion: 1.44`

The following command running on Ubuntu with `Docker version v29.0.0, build 3d4129b`

```
encore build docker 44303 --config infra-config.json --arch amd64
```

Fails with the following message:

```
Note: This CLI tool collects usage data to help us improve Encore.
      You can disable this by running 'encore telemetry disable'.

3:52PM INF compiling Encore application for linux/amd64
3:52PM INF resolving base image node:slim
3:52PM INF could not get image from local daemon, fetching it remotely
3:52PM INF saving remote image node:slim to local docker daemon
3:52PM WRN unable to save remote image to local docker daemon, skipping error="error loading image: Error response from daemon: client version 1.43 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version"
3:53PM INF saving image to local docker daemon
3:53PM ERR unable to save docker image error="error loading image: Error response from daemon: client version 1.43 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version"
```

Perhaps we can update the following package in [go.mod](https://github.com/encoredev/encore/blob/main/go.mod)? v26.1.3 would match `docker/cli` version used.

```
-github.com/docker/docker v24.0.7+incompatible // indirect
+github.com/docker/docker v26.1.3+incompatible // indirect
```

OR use `moby/moby` which `docker/docker` redirects to anyway

```
-github.com/docker/docker v24.0.7+incompatible // indirect
+github.com/moby/moby v26.1.3+incompatible // indirect
```

Qwerty-0101:
hi. i have same problem and i dont want downgrade my docker version.did u solve the problem ?

shimonenator:
> hi. i have same problem and i dont want downgrade my docker version.did u solve the problem ?

Downgraded

fredr:
should be fixed in #2152, I'll have a release out tomorrow

shimonenator:
Thank you, @fredr

fredr:
Released as part of v1.51.10, let me know if there is still any issues

Qwerty-0101:
> Released as part of v1.51.10, let me know if there is still any issues

i tried now its working. thanks for the fix.

PR #2152, opened 2025-11-17
Update go-containerregistry

fredr:
To support newer versions of docker (https://www.docker.com/blog/docker-engine-version-29/)

Should fix https://github.com/encoredev/encore/issues/2143

### DIFF
--- a/cli/daemon/run.go
+++ b/cli/daemon/run.go
@@ -104,14 +104,14 @@ func (s *Server) Run(req *daemonpb.RunRequest, stream daemonpb.Daemon_RunServer)
 
 	// If force upgrade has been enabled, we force the upgrade now before we try and run the app
 	if newVer != nil && newVer.ForceUpgrade {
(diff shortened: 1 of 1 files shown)


cli/daemon/run.go
```diff
@@ -104,14 +104,14 @@ func (s *Server) Run(req *daemonpb.RunRequest, stream daemonpb.Daemon_RunServer)
 
 	// If force upgrade has been enabled, we force the upgrade now before we try and run the app
 	if newVer != nil && newVer.ForceUpgrade {
-		_, _ = fmt.Fprintf(stderr, aurora.Red("An urgent security update for Encore is available.").String()+"\n")
+		_, _ = fmt.Fprintf(stderr, "%s", aurora.Red("An urgent security update for Encore is available.").String()+"\n")
```
eandre:
```suggestion
		_, _ = fmt.Fprint(stderr, aurora.Red("An urgent security update for Encore is available.").String()+"\n")
```

fredr: good point! updated!

PR #2162, opened 2025-11-21
cli: generate llm rules for apps

fredr:
Adds a bunch of tools to generate llm rules for, add a selector in the app create form, and add a separate command to generate the files in existing apps.

### DIFF
--- a/cli/cmd/encore/llm_rules/llm_rules.go
+++ b/cli/cmd/encore/llm_rules/llm_rules.go
@@ -7,8 +7,8 @@ import (
 )
 
 var llmRulesCmd = &cobra.Command{
-	Use:   "llm_rules",
-	Short: "Commands to create llm rules for apps",
+	Use:   "llm-rules",
+	Short: "Commands to create LLM rules for apps",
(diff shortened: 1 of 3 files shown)


cli/cmd/encore/llm_rules/llm_rules.go
```diff
@@ -0,0 +1,16 @@
+package llm_rules
+
+import (
+	"github.com/spf13/cobra"
+
+	"encr.dev/cli/cmd/encore/root"
+)
+
+var llmRulesCmd = &cobra.Command{
+	Use:   "llm_rules",
```
eandre: Can we call this `llm-rules` instead?

cli/cmd/encore/llm_rules/llm_rules.go
```diff
@@ -0,0 +1,16 @@
...
+	Short: "Commands to create llm rules for apps",
```
eandre:
```suggestion
	Short: "Commands to create LLM rules for apps",
```

PR #2165, opened 2025-11-25
docs: document the different env variables to configure encore

fredr:


docs/menu.cue
```diff
@@ -385,6 +385,11 @@
 				text: "Validation"
 				path: "/go/develop/validation"
 				file: "go/develop/validation"
+			}, {
+				kind: "basic"
+				text: "Environment"
```
eandre:
```suggestion
				text: "Environment Variables"
```

docs/menu.cue
```diff
@@ -385,6 +385,11 @@
...
+				path: "/go/develop/environment"
```
eandre: Can we rename this to `env-vars` (and the file name too)? Otherwise it's a bit confusing since we also have "cloud environments"

```suggestion
				path: "/go/develop/env-vars"
```

docs/menu.cue
```diff
@@ -838,11 +843,16 @@
...
 				file: "ts/develop/multithreading"
-			},{
+			}, {
 				kind: "basic"
 				text: "Running Scripts"
 				path: "/ts/develop/running-scripts"
 				file: "ts/develop/running-scripts"
+			}, {
+				kind: "basic"
+				text: "Environment"
```
eandre:
```suggestion
				text: "Environment Variables"
```

docs/menu.cue
```diff
@@ -838,11 +843,16 @@
...
+				path: "/ts/develop/environment"
```
eandre: Same here (`env-vars`)

fredr: fixed!

PR #2166, opened 2025-11-25
clientgen/typescript: render pointers as unions of type and null

fredr:


eandre (changes requested):
This is already implemented in the PR for option.Option[T], sorry!

fredr:
ah right, no worries. Closing this one!

PR #2167, opened 2025-11-26
runtimes/go: add Exposed, AuthRequired to (encore.dev).APIDesc

eandre:


runtimes/go/appruntime/apisdk/api/auth.go
```diff
@@ -175,6 +175,9 @@ func (d *AuthHandlerDesc[Params]) rpcDesc() *model.RPCDesc {
 			RequestType:  nil,
 			ResponseType: nil,
 			Tags:         nil,
+
+			Exposed:      true,
+			AuthRequired: false,
```
fredr, line 180: shouldn't this be true since it is the auth handler? just out of curiosity

eandre: Was thinking about it, but my logic was that calling the auth handler doesn't require authentication

fredr: makes sense!

fredr (dismissed):
looks good! (but you have a failing test)

PR #2168, opened 2025-11-26
Add config option to disable request logging

djeer:
We are flooded by request logging requests when developing locally, it's impossible to find useful information between noisy log lines, and setting WARN level is not an option because our app emits mist of the logs on the INFO level. Let's add an option to disable request logging.

image


eandre:
Maybe a better option is to just decrease the log level of these to Trace, and add a better way to set the minimum log level?

djeer:
@eandre sure that's another good option if you don't consider it a breaking change, let's do it! https://github.com/encoredev/encore/pull/2169
I agree they should be trace ideally

djeer:
Closing because alternative PR #2169 merged

PR #2178, opened 2025-11-29
exposing NSQ ports to external serivces messaging #issue-2177

### LINKED ISSUE #2177: Ability to expose/override nsq port
I had my encore service running in my local connected to NSQ,
 the problem I had is that I had an external service(not implemented in encore) that needs to send events over NSQ so encore can process them,
 I checked on the source code and there is not environment variable that allow me to override this and expose the port to my local machine so I can send events.

eramosr25:
Fixes #2177


fredr (changes requested):
Thanks for adding this, and sorry for the slow response. 

Please change the suffix from _ADDRESS to _LISTEN_ADDR so they match all the other environment variables

PR #2182, opened 2025-12-02
should not validate streaming req/resp the same way as regular req/resp

### LINKED ISSUE #2181: API schema validation incorrectly rejects `StreamInOut` messages
## Issue
As of [#2149](https://github.com/encoredev/encore/pull/2149), the `Message` types of `StreamIn` and `StreamInOut` endpoints are being validated as if they were query params.

This means nested objects now cause compilation to error out with the `error: type not supported for query parameters` message.

## Cause
I suspect [these three lines](https://github.com/encoredev/encore/pull/2149/files#diff-2cb01c9d99a465327fe2c7175ece7590a20530674d66dc8ae0d78b0c08367a41R136-R138) should _not_ be run when the `Endpoint` is `StreamIn` or `StreamInOut`.

It's unclear to me if the `Endpoint` struct is aware of when it's `StreamIn` or `StreamInOut` (other than during handshake) otherwise I would take a stab at PRing the fix.

## Repro
```js
import { api } from "encore.dev/api";
export const repro = api.streamIn<{ a: { b: number } }, void>(
    { path: "/api/repro" },
    async (stream) => {},
);

// error: type not supported for query parameters
// export const repro = api.streamIn<{ a: { b: number } }, void>(
//                                      ^^^^^^^^^^^^^
```

fredr:
Fixes #2181

### DIFF
--- a/tsparser/src/app/mod.rs
+++ b/tsparser/src/app/mod.rs
@@ -133,10 +133,14 @@ impl AppValidator<'_> {
         if let Some(req_enc) = &ep.encoding.handshake {
             self.validate_req_params(&req_enc.params);
         }
-        for req_enc in &ep.encoding.req {
-            self.validate_req_params(&req_enc.params);
+        if !ep.streaming_request {
+            for req_enc in &ep.encoding.req {
+                self.validate_req_params(&req_enc.params);
+            }
         }
-        self.validate_resp_params(&ep.encoding.resp.params);
+        if !ep.streaming_response {
+            self.validate_resp_params(&ep.encoding.resp.params);
+        }
         if let Some(schema) = &ep.encoding.raw_handshake_schema {
             self.validate_schema_type(schema);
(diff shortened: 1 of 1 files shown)


eandre (commented):
I mean they are essentially req/resp params, right? Would it be better to call the validation method still, but have it skip the parts that don't apply to streams? Or put differently, we want to validate the request as if it were a body payload, not a GET payload, right?

fredr:
> I mean they are essentially req/resp params, right? Would it be better to call the validation method still, but have it skip the parts that don't apply to streams? Or put differently, we want to validate the request as if it were a body payload, not a GET payload, right?

yeah, probably, can have a look at improving that, they are handled a bit special, but this at least unblocks anyone that have that problem right now, so I think it makes sense to release this now

fredr:
And they will probably need to be validated a bit differently anyways, like http status etc

PR #2183, opened 2025-12-02
tsparser: parse stream request as post

fredr:


fredr:
Will implement this differently

PR #2188, opened 2025-12-05
Add `parent_span_id` to protobuf def of SpanSummary for Traceview

eleijonmarck:
We need `parent_span_id` to build traceview from only a trace with a specific span. This enables us to write `parent_span_id` and consume for spans in the frontend where we want to view the spans to build the spantree.

This adds
- `parent_span_id` to protobuf on trace2.proto
- adds column to database
- adds write to column of the database
- `emitCompleteSpanToListeners()`, WebSocket events will have NULL parent_span_id (acceptable as it is not used by frontend)
- does not add any backfill for `Get`. Get uses events for populating
- does not add any backfill for `List`, see questions answered

**Questions answered:**
why do we not need to backFill `parent_span_id` on the spans returned by List?
It is because list filters on `is_root` , is_root is always the root span there the parent_id is "". Therefore we do not need to backfill the spans returned in **List**.

### DIFF
--- a/cli/daemon/engine/trace2/sqlite/read.go
+++ b/cli/daemon/engine/trace2/sqlite/read.go
@@ -58,7 +58,6 @@ func (s *Store) List(ctx context.Context, q *trace2.Query, iter trace2.ListEntry
 	}
 
 	defer fns.CloseIgnore(rows)
-
 	n := 0
 	for rows.Next() {
 		if n >= limit {
@@ -68,7 +67,6 @@ func (s *Store) List(ctx context.Context, q *trace2.Query, iter trace2.ListEntry
 
 		var t tracepb2.SpanSummary
 		var startedAt int64
-
 		err := rows.Scan(
 			&t.TraceId, &t.SpanId, &startedAt, &t.Type, &t.IsRoot, &t.ServiceName, &t.EndpointName,
 			&t.TopicName, &t.SubscriptionName, &t.MessageId, &t.IsError, &t.TestSkipped,
@@ -84,7 +82,7 @@ func (s *Store) List(ctx context.Context, q *trace2.Query, iter trace2.ListEntry
 		}
 	}
 
-	return nil
+	return errors.Wrap(rows.Err(), "iterate traces")
 }
 
 // emitCompleteSpanToListeners emits the given trace/span to all listeners
@@ -92,7 +90,6 @@ func (s *Store) List(ctx context.Context, q *trace2.Query, iter trace2.ListEntry
 func (s *Store) emitCompleteSpanToListeners(ctx context.Context, appID, traceID, spanID string) {
 	var t tracepb2.SpanSummary
 	var startedAt int64
-
 	err := s.db.QueryRowContext(ctx, `
 		SELECT
 			trace_id, span_id, started_at, span_type, is_root, service_name, endpoint_name,
@@ -114,7 +111,6 @@ func (s *Store) emitCompleteSpanToListeners(ctx context.Context, appID, traceID,
 
 	ts := time.Unix(0, startedAt)
 	t.StartedAt = timestamppb.New(ts)
-
 	for _, ln := range s.listeners {
 		ln <- trace2.NewSpanEvent{
 			AppID:     appID,
@@ -156,7 +152,7 @@ func (s *Store) Get(ctx context.Context, appID, traceID string, iter trace2.Even
 	if err := rows.Err(); err != nil {
 		return errors.Wrap(err, "iterate events")
 	} else if !hasRows {
-		return errors.Wrap(rows.Err(), "iterate traces")
+		return trace2.ErrNotFound
 	}
 	return nil
 }


cli/daemon/engine/trace2/sqlite/write.go
```diff
@@ -177,49 +177,65 @@ func (s *Store) updateSpanStartIndex(ctx context.Context, meta *trace2.Meta, ev
...
+			encodedParentSpanID := encodeSpanID(*start.ParentSpanId)
+			parentSpanID = &encodedParentSpanID
+		}
 		_, err := s.db.ExecContext(ctx, `
 			INSERT INTO trace_span_index (
 				app_id, trace_id, span_id, span_type, started_at, is_root, service_name,
-				endpoint_name, has_response, test_skipped
-			) VALUES (?, ?, ?, ?, ?, ?, ?, ?, false, false)
+				endpoint_name, has_response, test_skipped, parent_span_id
+			) VALUES (?, ?, ?, ?, ?, ?, ?, ?, false, false, ?)
```
eandre: Nitpick: I generally would put the `parent_span_id` after `span_id` in the list, rather than at the end, since that makes more sense semantically. I would also try to avoid putting a placeholder after the two `false` constants.

eleijonmarck: reviewed this change and it makes sense for sure.

cli/daemon/engine/trace2/sqlite/read.go
```diff
@@ -80,24 +84,27 @@ func (s *Store) List(ctx context.Context, q *trace2.Query, iter trace2.ListEntry
 		}
 	}
 
-	return errors.Wrap(rows.Err(), "iterate traces")
```
eandre, line 83: Why was this removed?

cli/daemon/engine/trace2/sqlite/read.go
```diff
@@ -148,7 +156,7 @@ func (s *Store) Get(ctx context.Context, appID, traceID string, iter trace2.Even
 	if err := rows.Err(); err != nil {
 		return errors.Wrap(err, "iterate events")
 	} else if !hasRows {
-		return trace2.ErrNotFound
+		return errors.Wrap(rows.Err(), "iterate traces")
```
eandre: Why was this changed? `rows.Err()` is already guaranteed to return `nil` from line 156, so this seems like a mistake

cli/daemon/engine/trace2/sqlite/read.go
eleijonmarck: We collected rows in memory and i did a bad replace of error collections, when i removed the backfill logic i missed this.

cli/daemon/engine/trace2/sqlite/read.go
eleijonmarck: See comment https://github.com/encoredev/encore/pull/2188/changes#r2620207239

cli/daemon/engine/trace2/sqlite/write.go
```diff
@@ -177,77 +177,103 @@ func (s *Store) updateSpanStartIndex(ctx context.Context, meta *trace2.Meta, ev
...
+		if start.ParentSpanId != nil {
+			encodedParentSpanID := encodeSpanID(*start.ParentSpanId)
+			parentSpanID = &encodedParentSpanID
+		}
 		_, err := s.db.ExecContext(ctx, `
 			INSERT INTO trace_span_index (
-				app_id, trace_id, span_id, span_type, started_at, is_root, service_name, endpoint_name, external_request_id, has_response, test_skipped
-			) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, false, false)
+				app_id, trace_id, span_id, span_type, started_at, is_root, service_name, endpoint_name, external_request_id, parent_span_id,
+				has_response, test_skipped
```
eleijonmarck, line 188: made boolean flags, as a separate line.

cli/daemon/engine/trace2/sqlite/write.go
```diff
@@ -177,77 +177,103 @@ func (s *Store) updateSpanStartIndex(ctx context.Context, meta *trace2.Meta, ev
...
+			encodedParentSpanID := encodeSpanID(*start.ParentSpanId)
+			parentSpanID = &encodedParentSpanID
+		}
 		_, err := s.db.ExecContext(ctx, `
 			INSERT INTO trace_span_index (
-				app_id, trace_id, span_id, span_type, started_at, is_root, service_name,
-				endpoint_name, has_response, test_skipped
-			) VALUES (?, ?, ?, ?, ?, ?, ?, ?, false, false)
+				app_id, trace_id, span_id, span_type, started_at, is_root, service_name, endpoint_name, parent_span_id,
+				has_response, test_skipped
```
eleijonmarck, line 214: made boolean flags, as a separate line.

eandre (approved):

PR #2193, opened 2025-12-09
tsparser: support key remapping via as

fredr:
https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as


eandre (approved):
Nice!

Issue #1867, opened 2025-04-12
Unable to generate OpenAPI schema due to "Internal Compiler Error"


L8D:
Hello. I tried to generate an OpenAPI schema for the first time. This is what happened:

```
phnx% encore gen client phoenix-api-kwr2 --lang=openapi --output=./openapi.json

error: 
── Internal compiler error ────────────────────────────────────────────────────────────────[E0001]──

An unhandled panic occurred in the Encore compiler: runtime error: invalid memory address or nil
pointer dereference

This is a bug in Encore and should not have occurred. Please report this issue to the Encore team
either on Github at https://github.com/encoredev/encore/issues/new and include this error.

╭─[Stack Trace]
├─▶ clientgen.Client.func1              encr.dev/pkg/clientgen/client.go:65
├─▶ openapi.(*Generator).Generate.func1 encr.dev/pkg/clientgen/openapi/openapi.go:54
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ openapi3.SchemaRef.MarshalJSON      github.com/getkin/kin-openapi@v0.115.0/openapi3/refs.go:568
╰─[... remaining frames omitted ...]

```

other details:

this is a typescript api
this has not been deployed to Encore Cloud yet

marcuskohlberg:
For visibility, this issue is being discussed on Discord: https://discord.com/channels/814482502336905216/1360749478390530230/1360749478390530230

fredr:
Looking at the stack trace, this looks like an issue that was solved in v1.52.5, please update and check if it solves the problem for you. Otherwise it would be great if you could update the issue with a way of reproducing the issue, e.g what type in your schema is causing the panic?

simon-johansson:
The compiler panic in the OpenAPI generation was fixed in v1.52.5. Since there's been no further response, we're going to close this out. If you're still running into this on a recent version, please reopen (or open a new issue) with details on the schema type causing the issue and we'll take another look!

Issue #1757, opened 2025-02-07
Bug when generating openapi spec


daankets:
```bash
encore gen client --lang openapi
```
```
error: 
── Internal compiler error ────────────────────────────────────────────────────────────────[E0001]──

An unhandled panic occurred in the Encore compiler: runtime error: invalid memory address or nil
pointer dereference

This is a bug in Encore and should not have occurred. Please report this issue to the Encore team
either on Github at https://github.com/encoredev/encore/issues/new and include this error.

╭─[Stack Trace]
├─▶ clientgen.Client.func1              encr.dev/internal/clientgen/client.go:65
├─▶ openapi.(*Generator).Generate.func1 encr.dev/internal/clientgen/openapi/openapi.go:54
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:294
╰─[... remaining frames omitted ...]
```

fredr:
This might have been solved in v1.52.5, please update and check if it solves your problem. Otherwise if you could provide a way of reproducing the problem would be great (e.g what type in your schema is causing this error)

simon-johansson:
Hi @daankets, since there's been no further response and this compiler panic should be resolved in v1.52.5+, we're going to close this issue. If you're still experiencing the problem on a recent version, feel free to reopen (or open a new issue) and we'll look into it!

daankets:
@simon-johansson I'll check it out!

Issue #1521, opened 2024-10-25
[Typescript] Openapi client genaration fails


darkstarac:
interface A {}
interface B {}
interface Response {
    data: A | B
}

export const api_call = api(
    { expose: true, auth: true, method: "GET", path: "/api/creatives" },  async (): Promise<Response> => {
        return {} as Response
    })

error:
── Internal compiler error ────────────────────────────────────────────────────────────────[E0001]──

An unhandled panic occurred in the Encore compiler: runtime error: invalid memory address or nil
pointer dereference

This is a bug in Encore and should not have occurred. Please report this issue to the Encore team
either on Github at https://github.com/encoredev/encore/issues/new and include this error.

╭─[Stack Trace]
├─▶ clientgen.Client.func1              encr.dev/internal/clientgen/client.go:65
├─▶ openapi.(*Generator).Generate.func1 encr.dev/internal/clientgen/openapi/openapi.go:54
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:293
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:293
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:293
├─▶ json.(*encodeState).marshal.func1   encoding/json/encode.go:293
╰─[... remaining frames omitted ...]

eandre:
Does this work when you run `encore run`? The syntax looks wrong to me. You want `async (): Promise<Response>`, no?

darkstarac:
Yes you are right. My fault it was a typo.

interface A {}
interface B {}
interface Response {
    data: A | B
}

export const api_call = api(
    { expose: true, auth: true, method: "GET", path: "/api/api_call" }, async (): Promise<Response> => {
        return {} as Response
    })

Works with: encore run.
Fails with:  encore gen client app-namei --lang=openapi --output=./openapi.json --env=local

compelinnovations:
Getting the same error, worked fine initially but after running it a couple of times this pops up.

error: 
── Internal compiler error ────────────────────────────────────────────────────────────────[E0001]──

An unhandled panic occurred in the Encore compiler: runtime error: invalid memory address or nil
pointer dereference

This is a bug in Encore and should not have occurred. Please report this issue to the Encore team
either on Github at https://github.com/encoredev/encore/issues/new and include this error.

╭─[Stack Trace]
├─▶ clientgen.Client.func1              encr.dev/internal/clientgen/client.go:65
[...]

PR #2198, opened 2025-12-11
clientgen: fix panic when generation openapi with named unions

fredr:
Reported in https://discord.com/channels/814482502336905216/1445923767015571599/1448428947607130183

```ts
type FooRequest = {
  foo: string;
};

type ApiRequest = {
  blob: FooRequest | string;
};

export const myApi = api<ApiRequest, void>(
  { expose: true, method: "POST" },
  (req) => {},
);
```

Now yields:
```json
{
  "schema": {
    "properties": {
      "blob": {
        "anyOf": [
          {
            "$ref": "#/components/schemas/hello.FooRequest"
          },
          {
            "type": "string"
          }
        ]
      }
    },
    "required": [
      "blob"
    ],
    "type": "object"
  }
}

```

Possibly also solves #1757 #1521 #1867

### DIFF
--- a/pkg/clientgen/openapi/schema.go
+++ b/pkg/clientgen/openapi/schema.go
@@ -186,12 +186,13 @@ func (g *Generator) schemaType(typ *schema.Type) *openapi3.SchemaRef {
 		}
 
 		// Otherwise, we have to represent this as an anyOf schema.
-		schemas := make([]*openapi3.Schema, 0, len(t.Union.Types))
+		schemaRefs := make([]*openapi3.SchemaRef, 0, len(t.Union.Types))
 		for _, tt := range t.Union.Types {
-			schemas = append(schemas, g.schemaType(tt).Value)
+			schemaRefs = append(schemaRefs, g.schemaType(tt))
 		}
 
-		s := openapi3.NewAnyOfSchema(schemas...)
+		s := openapi3.NewSchema()
+		s.AnyOf = schemaRefs
 		s.Nullable = haveLiteralNull
 		return s.NewRef()
 


eleijonmarck (approved):

eleijonmarck:
looks valid to me

pkg/clientgen/openapi/schema.go
```diff
@@ -186,12 +186,13 @@ func (g *Generator) schemaType(typ *schema.Type) *openapi3.SchemaRef {
 		}
 
 		// Otherwise, we have to represent this as an anyOf schema.
-		schemas := make([]*openapi3.Schema, 0, len(t.Union.Types))
+		schemaRefs := make([]*openapi3.SchemaRef, 0, len(t.Union.Types))
```
eleijonmarck, line 189: https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3#SchemaRef

definition. so we are nowing sending the ful schemaRefs

fredr:
> Does this change how the schema looks like? Does everything get nested as refs now?

No, I tried to show that in the example, e.g string is represented as  `{"type": "string"}` as before the change, but named will be represented as ref. Or where there any other types you where thinking about?

eandre (approved):

pkg/clientgen/openapi/schema.go
fredr: yes, the `NewAnyOfSchema` actually re-creates `SchemaRef` objects from the `Schema` objects, see https://github.com/getkin/kin-openapi/blob/v0.133.0/openapi3/schema.go#L596-L604

PR #2200, opened 2025-12-11
tsparser: add support for method signatures

fredr:
This doesn't add any new schema features, but makes it possible for the parser to parse method signatures, so that types like these are possible:

```ts
type AType = {
  a: string;
  b: number;
  c(x: string, z: number): boolean;
};

type MyType = Pick<AType, "a" | "b">;
```


eandre (approved):
Nice!

PR #2203, opened 2025-12-16
cli/daemon: handle invalid nsq names

fredr:
If nsq names are invalid (max 64 char and passes their regex), hash the name so that they are unique

### DIFF
--- a/runtimes/core/src/pubsub/nsq/topic.rs
+++ b/runtimes/core/src/pubsub/nsq/topic.rs
@@ -25,8 +25,8 @@ impl NsqTopic {
         let cloud_name = cfg.cloud_name.clone();
         let (tx, mut rx) = mpsc::channel::<PublishRequest>(32);
         tokio::spawn(async move {
-            let topic = NSQTopic::new(&cloud_name)
-                .expect("cloud_name should be valid NSQ topic name");
+            let topic =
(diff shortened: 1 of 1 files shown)


runtimes/core/src/pubsub/nsq/sub.rs
```diff
@@ -37,8 +38,15 @@ impl NsqSubscription {
         cfg: &pb::PubSubSubscription,
         meta: &meta::pub_sub_topic::Subscription,
     ) -> Self {
-        let topic = NSQTopic::new(cfg.topic_cloud_name.clone()).unwrap();
-        let channel = NSQChannel::new(cfg.subscription_cloud_name.clone()).unwrap();
+        let topic = NSQTopic::new(&cfg.topic_cloud_name).unwrap_or_else(|| {
+            let topic_name = hash_name(&cfg.topic_cloud_name);
```
eandre: We shouldn't handle this in the runtime, but when we generate NSQ names in the runtime config gen

fredr: Makes sense! Wasn't entirely sure where though? I moved it to when we create the runtime config in the daemon, since nsq is only for running locally, I'm guessing that should be enough?

PR #2206, opened 2025-12-18
encore test: fix conflict with pre-installed go versions

fredr:
In some corner case where a package didnt have any tests, running `encore test ./... -cover` caused errors like:
```
compile: version "go1.25.4-encore X:cacheprog" does not match go tool version "go1.25.4"
```

this is due a version conflict with a pre-installed go version. This makes sure we set the encore-go bin as the first item in the PATH so that wont happen

### DIFF
--- a/v2/compiler/build/tests.go
+++ b/v2/compiler/build/tests.go
@@ -189,15 +189,25 @@ func (b *builder) generateTestSpec(testCfg *GenerateTestSpecConfig) *TestSpec {
 		env = append(env, "CGO_ENABLED=0")
 	}
 
+	environ := append(os.Environ(), env...)
+
+	// Get the last PATH= item
+	var originalPath string
+	for _, e := range environ {
+		if path, ok := strings.CutPrefix(e, "PATH="); ok {
+			originalPath = path
+		}
+	}
+
 	// prefix PATH with encore-go, so it doesnt conflict with other installed go versions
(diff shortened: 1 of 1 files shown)


v2/compiler/build/tests.go
```diff
@@ -188,6 +189,11 @@ func (b *builder) generateTestSpec(testCfg *GenerateTestSpecConfig) *TestSpec {
 		env = append(env, "CGO_ENABLED=0")
 	}
 
+	// prefix PATH with encore-go, so it doesnt conflict with other installed go versions
+	// if not set causes problems when running cover tests in go
+	path := goroot.Join("bin").ToIO() + string(filepath.ListSeparator) + os.Getenv("PATH")
```
eandre: We should probably prefer using `PATH` from `b.cfg.Env` if it is provided there, by looping through the list looking for `PATH=` as a prefix (can use `strings.CutPrefix`). And falling back to os.Getenv("PATH") otherwise.

fredr: we do send in the whole os.Environ there it seems (maybe we shouldnt?), so then this fix wont help as that PATH will take prescience then.

fredr: ah sorry, now I understood what you meant!

fredr: fixed!

eandre (approved):

PR #2209, opened 2025-12-19
tsparser: add support for call expressions

fredr:
Add support for call expessions by parsing type annotations, not via body parsing

Add support for getting types like these:

```ts
export type Ret = {
  id: number;
  name: string;
  description: string;
};

function HelloWorld(a: string, b: number): Ret {
  return {
    id: 1,
    name: "Hello",
    description: "Hello description",
  };
}

export const X = HelloWorld("hello", 5);
export type Y = typeof X;

```

### DIFF
--- a/tsparser/src/parser/types/type_resolve.rs
+++ b/tsparser/src/parser/types/type_resolve.rs
@@ -701,6 +701,12 @@ impl Ctx<'_> {
     }
 
     fn parse_method_signature(&self, method: &ast::TsMethodSignature) -> FunctionType {
+        if let Some(type_params) = &method.type_params {
+            HANDLER.with(|handler| {
+                handler.span_err(type_params.span, "generic methods are not yet supported")
+            });
+        }
+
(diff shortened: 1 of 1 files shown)


tsparser/src/parser/types/type_resolve.rs
```diff
@@ -722,6 +728,108 @@ impl Ctx<'_> {
...
+                    "function must have explicit return type annotation (function body parsing not yet supported)",
+                )
+            });
+            Box::new(Type::Basic(Basic::Never))
+        };
+
+        FunctionType {
+            params,
+            return_type,
+            type_params: None, // Generic functions not yet fully supported
```
eandre: Should we error if there are type params?

fredr: yeah, that is probably better! will fix

fredr: updated

PR #2215, opened 2026-01-02
Adds `caller_event_id` to spansummaries same as `parent_span_id`

eleijonmarck:
We need to add the `caller_event_id` from startSpan to populate spans represented in the spanSummaries to have a full picture of the flow of the trace.

Context: CallerEventId We receive the CallerEventId as part of the SpanStart event. 

It represents the unique ID of the specific event in the parent span (e.g., the specific API call site) that triggered the creation of this new span. Capturing this allows the trace viewer to link the new span not just to a parent span, but to the exact causal event within that parent.

**Bug fixes**
We have seen bugs on not processing messages from pubsub messages in the localdev

image

PR #2216, opened 2026-01-02
bug: Fix argument list having duplicate reads 

eleijonmarck:
in pr https://github.com/encoredev/encore/pull/2215 we added too many argument to the sql scan listing


eleijonmarck:
follow up on - https://github.com/encoredev/encore/pull/2188

PR #2223, opened 2026-01-05
feat(traces): Adds `caller_event_id` to spanEnd message

eleijonmarck:
Contex:
We need to add `caller_event_id` to `requestSpanEnd` message in order for us to link spans and their respective span calls back to their original parent. Since a parent can have many child span calls, we need to record the caller_event_id to link it deterministically.

This PR:
- adds `caller_event_id` to the `requestSpanEnd` protobuf
- adds `caller_event_id` to the go runtime traceparser for handling `caller_event_id`
- adds `caller_event_id` to the ts runtime for handling `caller_event_id`

open questions:
- do we only need to use the `RequestSpanEnd` events or do we want to record the field to each of the events, such as the AuthSpan, PubSub, Test? See https://github.com/encoredev/encore/pull/2223#discussion_r2662036360. Answer we **ONLY** make it for `RequestSpanEnd`

### DIFF
--- a/runtimes/core/src/trace/protocol.rs
+++ b/runtimes/core/src/trace/protocol.rs
@@ -983,7 +983,7 @@ struct SpanEndEventData<'a> {
 
 impl SpanEndEventData<'_> {
     pub fn into_eb(self) -> EventBuffer {
-        let mut eb = EventBuffer::with_capacity(8 + 12 + 8 + 8 + self.extra_space);
+        let mut eb = EventBuffer::with_capacity(8 + 12 + 8 + self.extra_space);
 
         eb.duration(self.duration);
         eb.api_err_with_legacy_stack(self.err);


proto/encore/engine/trace2/trace2.proto
```diff
@@ -81,6 +81,7 @@ message SpanEnd {
   optional StackTrace panic_stack = 3;
   optional TraceID parent_trace_id = 4;
   optional uint64 parent_span_id = 5;
+  optional uint64 caller_event_id  = 6;
```
eandre: You also need to set this when constructing SpanEnd events in the trace parser, and also update the wire protocol to send this from both runtimes. And is this relevant for all types of SpanEnd events, or only RequestSpanEnd?

eleijonmarck: ah so both runtimes, would be finding where we write this for ts/go in the traceparser packages. gotcha 👍

pkg/traceparser/parser.go
```diff
@@ -179,12 +180,17 @@ func (tp *traceParser) spanEndEvent() spanEndEvent {
 	panicStack := tp.formattedStack()
 	parentTraceID := tp.traceID()
 	parentSpanID := tp.Uint64()
+	var callerEventID trace2.EventID
+	if tp.version >= 16 {
```
eleijonmarck: do not parse out callereventid if version is not 16

proto/encore/engine/trace2/trace2.proto
eleijonmarck: it is relevant for RequestSpanEnd AND PubsubMessageSpanEnd.

`RequestSpanEnd` is logical that it is needed to get the RPC of the exection. I am not sure about our `PubsubMessageSpanEnd` but i am not sure how we would link to the message processing execution.

AuthSpanEnd and TestSpanEnd is more questionable. I think AuthSpan is still somewhat relevant although not critical at all. TestSpan can for sure be neglected.


So if we proceed smaller surface, we could start with Request and PubSub only

runtimes/go/appruntime/exported/trace2/events.go
```diff
@@ -177,6 +178,7 @@ func (l *Log) newSpanEndEvent(data spanEndEventData) EventBuffer {
 
 	tb.Bytes(data.ParentTraceID[:])
 	tb.Bytes(data.ParentSpanID[:])
+	tb.UVarint(uint64(data.CallerEventID))
```
eandre: Now you're emitting it for all types of span end events. We want to only modify `RequestSpanEnd` (and same in the rust version)

proto/encore/engine/trace2/trace2.proto
eandre: Per the discussion, add it only to the `RequestSpanEnd` message

eleijonmarck: ah gotcha. so only on the requestspanend in protobuf as well

runtimes/core/src/trace/protocol.rs
```diff
@@ -981,7 +983,7 @@ struct SpanEndEventData<'a> {
 
 impl SpanEndEventData<'_> {
     pub fn into_eb(self) -> EventBuffer {
-        let mut eb = EventBuffer::with_capacity(8 + 12 + 8 + self.extra_space);
+        let mut eb = EventBuffer::with_capacity(8 + 12 + 8 + 8 + self.extra_space);
```
eandre: Revert this line since this function is not being otherwise modified to output more data

eleijonmarck: ah missed that line

PR #2231, opened 2026-01-09
runtimes/core: support equal paths with different path param names

fredr:


runtimes/core/src/api/paths.rs
```diff
@@ -112,6 +112,46 @@ where
...
 }
 
+/// Normalizes a path by replacing all parameter names with a canonical placeholder.
+/// This is used to group endpoints with the same structure but different param names.
+///
+/// Examples:
+/// - `/my/path/:name` -> `/my/path/:_`
+/// - `/my/path/:id` -> `/my/path/:_`
+/// - `/foo/:bar/baz/:qux` -> `/foo/:_/baz/:_`
+pub fn normalize_path(path: &str) -> String {
```
eandre: Wouldn't it be easier to modify the  `compute` fn above, where we compute the axum paths?

fredr: Ah nice catch!

fredr: fixed

runtimes/core/src/api/schema/path.rs
```diff
@@ -90,28 +92,34 @@ impl Path {
...
                 } => {
                     capacity += 10; // assume path parameters on average are 10 characters long
                     dynamic_segments.push((*typ, validation.clone()));
+                    dynamic_segment_names.push(name.clone());
                 }
-                Wildcard { validation, .. } | Fallback { validation, .. } => {
+                Wildcard { name, validation } | Fallback { name, validation } => {
                     // Assume path parameters on average are 10 characters long.
                     capacity += 10;
                     dynamic_segments.push((jsonschema::Basic::String, validation.clone()));
```
eandre: Would be nicer to add the name to this tuple instead?

fredr: can try that, makes sense!

fredr: fixed!

PR #2235, opened 2026-01-15
git: adds gitignore for self docs

eleijonmarck:


eandre:
Don't think we need personal docs in the public gitignore? Add them to `.git/info/exclude` if you want local .gitignore directories

eleijonmarck:
added to my exclude instead

PR #2238, opened 2026-01-16
Update ts_llm_instructions.txt

simon-johansson:
Optimized llm_instructions.txt to reduce token usage while preserving all functionality:

  - 45% size reduction (51,741 → 32,082 bytes) through removal of excessive XML nesting, verbose wrapper tags, and unnecessary whitespace
  - Added testing section covering Vitest setup, encore test command, and test patterns for APIs, databases, Pub/Sub, and Cron Jobs
  - All code examples, import statements, and CLI commands preserved exactly as-is
  - Flattened verbose structures like <core_concepts><concept> to inline prose
  - Consolidated repetitive patterns (e.g., error codes now in compact list format)

### DIFF
--- a/ts_llm_instructions.txt
+++ b/ts_llm_instructions.txt
@@ -431,6 +431,7 @@ Built-in request validation using TypeScript types for both runtime and compile-
 
 Example:
 import { Header, Query, api } from "encore.dev/api";
+import { Min, Max, MinLen, MaxLen, IsEmail, IsURL, StartsWith, EndsWith, MatchesRegexp } from "encore.dev/validate";
 
 interface Request {
   limit?: Query<number>;               // Optional query parameter


simon-johansson:
Reminder to check if all of the import statements are correct. 
`import { middleware, Service } from "encore.dev/api"` is not for example

bwehrle:
Wanted to comment since I saw this PR raised in Encore Discord:

* I frequently see that writing tests is a problem.  Often the LLM will write vittest tests for "integration" tests and create encore mocks instead of using the encore dependencies directly and running with `encore test.  Can we fix that?  I don't see any hints for that in the current LLM instructions.  Happy to make a separate PR.

* Wouldn't it be good to have some tests for these types of features?  I mean there should be some code generation validation if the instructions are changed.

simon-johansson:
> Wanted to comment since I saw this PR raised in Encore Discord:
> 
> * I frequently see that writing tests is a problem.  Often the LLM will write vittest tests for "integration" tests and create encore mocks instead of using the encore dependencies directly and running with `encore test.  Can we fix that?  I don't see any hints for that in the current LLM instructions.  Happy to make a separate PR.
> * Wouldn't it be good to have some tests for these types of features?  I mean there should be some code generation validation if the instructions are changed.

Thanks for the comments! I have added a testing section covering Vitest setup, encore test command, and test patterns for APIs, databases, Pub/Sub, and Cron Jobs. Also, I can recommend checking out the Encore Claude Code skills: https://github.com/encoredev/skills

PR #2240, opened 2026-01-18
cli/daemon/engine/trace2/sqlite: fix updateSpanEndIndex

eandre:
A recent change added a placeholder but no corresponding argument.


ivancernja (approved):
👍

PR #2256, opened 2026-01-28
docs: add docs for build hooks

fredr:


docs/platform/deploy/deploying.md
```diff
@@ -134,8 +134,9 @@ with their default values:
 
         // Docker-related configuration
         "docker": {
-        	// The Docker base image to use when deploying the application.
-        	// It must be a publicly accessible image, and defaults to "scratch".
+            // The Docker base image to use when deploying the application.
+            // It must be a publicly accessible image. It defaults to "scratch" for go apps
+            // and "node:24-trixie" for typescript apps.
```
eleijonmarck, line 139: this seems like we need to update or will get out of date fast. maybe we dont change node version that frequent

fredr: we dont update so often, but yeah, would be nice to get this some other way, at least it is in the docs now

PR #2262, opened 2026-01-29
runtimes/core: add support for trace sampling rate

fredr:


runtimes/core/src/api/reqauth/mod.rs
```diff
@@ -191,32 +196,36 @@ impl CallMeta {
...
             // In the future we should be able to remove this check and read the traceparent header for all requests
             // to interopt with other tracing systems.
-            if let Some(traceparent) = headers.get_meta(MetaKey::TraceParent) {
-                // Parse the traceparent.
-                if let Ok((trace_id, parent_span_id)) = parse_traceparent(traceparent) {
-                    meta.trace_id = trace_id;
-                    meta.caller_trace_id = Some(trace_id);
-                    meta.parent_span_id = Some(parent_span_id);
-                };
+            if meta.internal.is_some() {
```
fredr, line 199: this matches how we do things in the go runtime, so I assume this was just a miss? if we dont have this we will overwrite the sample rate from gcp etc

runtimes/core/src/api/call.rs
```diff
@@ -127,20 +127,29 @@ impl ServiceRegistry {
...
     ) -> impl Future<Output = APIResult<ResponsePayload>> + 'static {
         let tracer = self.tracer.clone();
-        let call = model::APICall { source, target };
-        let start_event_id = tracer.rpc_call_start(&call);
+        let start_event_id = source.as_deref().and_then(|source| {
+            tracer.rpc_call_start(crate::trace::protocol::RPCCallStartData {
+                source,
+                target: &target,
+            })
+        });
```
eandre: Why the change here? Not sure I understand

runtimes/core/src/api/reqauth/mod.rs
eandre: Can you explain in more detail? Not sure what you mean by "we will overwrite the sample rate from gcp etc"

fredr: if the request coming in via the gcp infra structure and not a service-to-service call, it will already have a traceparent header set, e.g if running on cloud run, and the sample bit will then override our sample rate and always sample

Reading the comment above this, I read it like this is how we want it to work? we also do the same in the golang runtime here: https://github.com/encoredev/encore/blob/57dc056b235be4363c4ee1f364dca4eedf8455cb/runtimes/go/appruntime/apisdk/api/call_meta.go#L233-L240

runtimes/core/src/api/call.rs
fredr: I changed the call_end to have a specific type instead of a lot of parameters, and I changed this one as well just to make their signatures look the same, but yeah, this one only have one parameter so doesnt really make sense, will change that back and check if source is none in the function instead

fredr: updated, and simplified the rpc_call_end call as well so we do the checks inside the function instead, makes it easier to call, let me know if you dont like it and I'll move the event_start_id none-check outside again

PR #2284, opened 2026-02-12
Add integration docs for Better Auth, Polar, and Resend

ivancernja:
Add documentation pages for Better Auth, Polar, and Resend integrations. Add top-level Resources section in sidebar.

### DIFF
--- a/docs/menu.cue
+++ b/docs/menu.cue
@@ -1032,6 +1032,26 @@
 		},
 		{
 			kind: "section"
+			text: "Resources"
+			items: [{
+				kind: "basic"
+				text: "Better Auth"
+				path: "/ts/develop/integrations/better-auth"
+				file: "ts/develop/integrations/better-auth"
+			}, {
+				kind: "basic"
+				text: "Polar"
+				path: "/ts/develop/integrations/polar"
+				file: "ts/develop/integrations/polar"
+			}, {
+				kind: "basic"
+				text: "Resend"
+				path: "/ts/develop/integrations/resend"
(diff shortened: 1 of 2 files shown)


docs/menu.cue
```diff
@@ -874,6 +874,30 @@
 				}]
 			}]
 		},
+		{
+			kind: "section"
+			text: "Resources"
+			items: [{
+				kind: "accordion"
```
marcuskohlberg, line 881: Should we stick stick this new section below the self-hosting section instead?
Feels a bit odd to have resources so high up.
And maybe drop the accordion? Feels very hidden. We can add it when we have too many guides to show.

docs/ts/develop/integrations/better-auth.md
```diff
@@ -0,0 +1,231 @@
...
+lang: ts
+---
+
+[Better Auth](https://www.better-auth.com) is a comprehensive TypeScript authentication library that supports email/password, OAuth, two-factor, magic links, and sessions. Combined with Encore's automatic [database provisioning](https://encore.dev/docs/ts/primitives/databases) and [secrets management](https://encore.dev/docs/ts/primitives/secrets), you get production-ready auth without managing any infrastructure.
+
+<Callout type="info">
+
+If you haven't installed Encore yet, see the [installation guide](https://encore.dev/docs/ts/install) first.
+
+</Callout>
```
marcuskohlberg, line 22: The guide assumes you have an app already I guess?
Maybe worth making that explicit?

PR #2285, opened 2026-02-13
feat(secrets): [encore] Add `delete` command to secrets

eleijonmarck:
Depends on platform - https://github.com/encoredev/platform/pull/2311

### DIFF
--- a/docs/go/cli/cli-reference.md
+++ b/docs/go/cli/cli-reference.md
@@ -18,18 +18,18 @@ $ encore run [--debug] [--watch=true] [--port=<number>] [--listen=<addr>] [flags
 
 **Flags**
 
-| Flag              | Description                                                            | Default       |
-| ----------------- | ---------------------------------------------------------------------- | ------------- |
-| `-w, --watch`     | Watch for changes and live-reload                                      | `true`        |
-| `--listen`        | Address to listen on (e.g. `0.0.0.0:4000`)                             |               |
-| `-p, --port`      | Port to listen on                                                      | `4000`        |
-| `--json`          | Display logs in JSON format                                            | `false`       |
-| `-n, --namespace` | Namespace to use (defaults to active namespace)                        |               |
-| `--color`         | Whether to display colorized output                                    | auto-detected |
-| `--redact`        | Redact sensitive data in traces when running locally                   | `false`       |
-| `-l, --level`     | Minimum log level to display (`trace\|debug\|info\|warn\|error`)       |               |
-| `--debug`         | Compile for debugging (`enabled\|break`)                               |               |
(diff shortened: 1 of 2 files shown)


cli/cmd/encore/secrets/archive.go
```diff
@@ -36,21 +36,12 @@ func doArchiveOrUnarchive(groupID string, archive bool) {
...
-		Archived: &archive,
-	})
-	if err != nil {
-		cmdutil.Fatal(err)
-	}
-	if archive {
-		fmt.Printf("Successfully archived secret group %s.\n", groupID)
+	if version.Compare("1.2.521") < 0 {
+		// older version
+		fmt.Printf("Please update your encore CLI to a newer version")
```
eandre: I don't understand why this is needed. If this code is running then we know for a fact we're already running the newer version?

cli/cmd/encore/secrets/delete.go
```diff
@@ -0,0 +1,62 @@
...
+	Short:                 "Deletes a secret value",
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Check if --yes / --force flag was passed to skip confirmation
+		if !forceFlag {
+			fmt.Printf("Are you sure you want to delete secret %q? [y/N]: ", args[0])
+			reader := bufio.NewReader(os.Stdin)
+			response, _ := reader.ReadString('\n')
+			response = strings.TrimSpace(strings.ToLower(response))
```
eandre: Use
```suggestion
			var response string
			_, _ = fmt.Scanln(&response)
```

docs/go/cli/cli-reference.md
```diff
@@ -277,15 +277,23 @@ Lists secrets, optionally for a specific key
...
+#### Archive (Deprecated)
 
 Archives a secret value
 
 ```shell
 $ encore secret archive <id>
 ```
 
-#### Unarchive
+#### Unarchive (Deprecated)
```
eandre: Let's just remove this section from the docs

docs/ts/cli/cli-reference.md
```diff
@@ -270,15 +270,23 @@ Lists secrets, optionally for a specific key
...
+#### Archive (Deprecated)
 
 Archives a secret value
 
 ```shell
 $ encore secret archive <id>
 ```
 
-#### Unarchive
+#### Unarchive (Deprecated)
```
eandre: Same

cli/cmd/encore/secrets/archive.go
eleijonmarck: oh i am sorry, this was there from the beginning. this should def not be in the PR. this is something that i experimented with in the beginning.

this is not needed 👍

PR #2375, opened 2026-03-31
runtimes/go: don't emit sub events when request is not traced

fredr:


fredr:
This is already handled, `curr.Trace` will be nil if `Traced` is false

PR #2390, opened 2026-04-08
runtimes: auth handler sampling decision

fredr:
Fixes the go local auth handler so that it uses the same trace sampling decision as from the endpoint. The same with remote auth handlers for both core and go runtime (though we dont actually use the remote auth handler in core)

### DIFF
--- a/runtimes/core/src/api/auth/remote.rs
+++ b/runtimes/core/src/api/auth/remote.rs
@@ -82,17 +82,10 @@ impl RemoteAuthHandler {
         let caller = Caller::APIEndpoint(EndpointName::new("gateway", "__encore/authhandler"));
 
         let meta = &req.call_meta;
-        // Always include the trace ID so traceparent is sent, even without a parent span.
-        // This ensures the tracestate (with the sampling decision) is read by the receiver,
-        // which is necessary on non-GCP environments where Cloud Run doesn't inject a traceparent.
-        let zero_span = crate::model::SpanId([0u8; 8]);
-        let parent_span = Some(
-            meta.trace_id
-                .with_span(meta.parent_span_id.unwrap_or(zero_span)),
-        );
         let desc: CallDesc<()> = CallDesc {
             caller: &caller,
-            parent_span,
+            trace_id: Some(meta.trace_id),
+            parent_span: meta.parent_span_id.map(|sp| meta.trace_id.with_span(sp)),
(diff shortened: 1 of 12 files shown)


runtimes/core/src/api/reqauth/mod.rs
```diff
@@ -227,8 +231,18 @@ impl CallMeta {
...
+                    // Also clear for remote auth handler calls, which are
+                    // conceptually root requests from the gateway.
                     if let Some(internal) = &meta.internal {
-                        if matches!(internal.caller, Caller::Gateway { .. }) {
+                        let clear_parent = match &internal.caller {
+                            Caller::Gateway { .. } => true,
+                            Caller::APIEndpoint(name) => {
+                                name.service() == "gateway"
+                                    && name.endpoint() == "__encore/authhandler"
+                            }
```
eandre: This part seems very brittle to me. Is there no better way of accomplishing this?

fredr: this was based on https://github.com/encoredev/encore/blob/66c3ca179438417b32d65845d87eed87b8a4388f/runtimes/core/src/api/auth/remote.rs#L79-L82

but we can probably check if tracestate doesnt contain a span_id and reset it then, if it is an internal call

fredr: updated, let me know what you think about that

runtimes/core/src/api/auth/remote.rs
```diff
@@ -82,9 +82,17 @@ impl RemoteAuthHandler {
         let caller = Caller::APIEndpoint(EndpointName::new("gateway", "__encore/authhandler"));
 
         let meta = &req.call_meta;
+        // Always include the trace ID so traceparent is sent, even without a parent span.
+        // This ensures the tracestate (with the sampling decision) is read by the receiver,
+        // which is necessary on non-GCP environments where Cloud Run doesn't inject a traceparent.
```
eandre: Seems a bit hacky. If we want to always send the tracestate header, isn't it better to change the code to always send it even if the parent span id is None? Instead of changing to set the parent to Some(zero)?

fredr: updated the code so we pass the trace_id even if we dont have a parent span

eandre (approved):

PR #2392, opened 2026-04-08
feat(sqldb): support ENCORE_DOCKER_NETWORK env var

### LINKED ISSUE #2382: Support configurable Docker network for test databases (ENCORE_DOCKER_NETWORK)
## Problem

`encore test` spawns Postgres containers via `docker run -d -p 5432` with no `--network` flag, so they always land on Docker's default bridge network.

In Google Cloud Build, build steps run on the `cloudbuild` Docker network. The test container and the Postgres container end up on different networks, making `127.0.0.1:<mapped-port>` unreachable:

```
failed to start cluster: database did not come up: failed to connect to
`user=postgres database=postgres`:
    127.0.0.1:32768 (127.0.0.1): dial error: timeout: context deadline exceeded
```

This affects any CI system that runs build steps on a non-default Docker network (Cloud Build, some GitLab CI runners, Drone, etc.).

## Proposed solution

Read an environment variable (e.g. `ENCORE_DOCKER_NETWORK`) in `CreateCluster` and pass `--network` to the `docker run` invocation when set.

Rough sketch in `cli/daemon/sqldb/docker/docker.go`:

```go
if net := os.Getenv("ENCORE_DOCKER_NETWORK"); net != "" {
    args = append(args, "--network", net)
}
```

**Usage in Cloud Build:**

```yaml
- id: test
  name: my-builder-image
  env:
    - ENCORE_DOCKER_NETWORK=cloudbuild
  args: [-c, encore test ./...]
```

## Workarounds

The only current workaround is polling `docker ps` and running `docker network connect cloudbuild <container>` after Encore spawns the Postgres container, which is fragile and race-prone.

odsod:
## Summary

- Pass `--network` to `docker run` when `ENCORE_DOCKER_NETWORK` is set
- Fixes #2382: Postgres containers are unreachable in Google Cloud Build, where build steps run on the `cloudbuild` Docker network

## Usage

```yaml
# Google Cloud Build
- id: test
  name: my-builder-image
  env:
    - ENCORE_DOCKER_NETWORK=cloudbuild
  args: [-c, encore test ./...]
```

### DIFF
--- a/cli/daemon/sqldb/docker/docker.go
+++ b/cli/daemon/sqldb/docker/docker.go
@@ -140,6 +140,9 @@ func (d *Driver) CreateCluster(ctx context.Context, p *sqldb.CreateParams, log z
 			"-e", "PGDATA=" + defaultDataDir,
 			"--name", cnames[0],
 		}
+		if net := os.Getenv("ENCORE_DOCKER_NETWORK"); net != "" {
+			args = append(args, "--network", net)
+		}
 		if p.Memfs {
 			args = append(args,
 				"--mount", "type=tmpfs,destination="+defaultDataDir,


fredr:
Thanks for adding this! 

I don't think it is enough to just set the network name, we connect to 127.0.0.1 down in clusterStatus, and for this to work I think we need to use the container name instead. So maybe if the docker network env var is set, connect to containerName:port instead of 127.0.0.1:port

PR #2394, opened 2026-04-10
runtimes/go: add metrics.SetServiceLabels for per-service label enrichment

twindebank:
Adds `metrics.SetServiceLabels(serviceName, labels)` which registers additional labels to be included with all metrics exported for a given service. This enables enriching built-in metrics (like `e_requests_total`) with custom metadata without duplicating the metric pipeline via OTel.

All four exporters are updated (Prometheus, GCP Cloud Monitoring, Datadog, CloudWatch). Reserved keys and empty values are filtered at registration time.


bkach (approved):
Adding service labels makes sense, nice one!

bkach (approved):

PR #2421, opened 2026-04-30
cli: render OpTracker as plain lines on non-interactive stderr

dstotijn:
## Summary

Encore's build-progress UI is rendered by `internal/optracker.OpTracker`, which uses ANSI cursor save/restore for in-place spinner updates. That works on a TTY, but when stderr is piped — `mise run`, log aggregators, CI, `tee`, Docker logs — every 100ms refresh becomes a new line and floods the output.

#2423 added a `lineMode` rendering path to `OpTracker` (via `NewLineMode(w)`) and uses it for the new `encore check` command. This PR extends that same rendering to the existing `encore run` and `encore exec` flows when stderr isn't a terminal.

## Fix

- New proto field `non_interactive` on `RunRequest`, `ExecScriptRequest`, `ExecSpecRequest`. Defaults to false → existing behavior, fully backwards compatible.
- The CLI auto-sets it from `!term.IsTerminal(int(os.Stderr.Fd()))`.
- The daemon constructs the tracker via `optracker.NewLineMode(stderr)` instead of `optracker.New(stderr, stream)` when the request is `non_interactive`.

Example non-interactive output:

```
encore: Building Encore application graph...
encore: Analyzing service topology...
encore: Building Encore application graph done (1.2s)
encore: Analyzing service topology done (0.4s)
...
```

## Alternatives considered

- `--quiet` flag / `ENCORE_QUIET` env var — rejected. The flooding is an output-format bug, not a verbosity preference; surfacing it as user config asks people to opt out of broken behavior.
- Total suppression when non-TTY — rejected. Users still want to see what's happening; they only want to stop the spinner spam.

## Test plan

- [ ] Interactive: `encore run` from a TTY shows the spinner UI as before.
- [ ] Non-interactive: `encore run 2>&1 | cat` produces line-mode output, no spinner / cursor escapes.
- [ ] Same for `encore exec` (Go and TS apps).


eandre:
Created #2428 which is this PR but with the protobuf generated using the correct versions. Thanks!

dstotijn:
Closing, change is handled in https://github.com/encoredev/encore/pull/2428.

PR #2422, opened 2026-04-30
Fix DummySocket for latest Node socket types

amikad09:
Add getTypeOfService and setTypeOfService to DummySocket.

These methods were added to net.Socket in the latest Node type definitions. Without them, the DummySocket cast to Socket fails during type checking.

The implementations are no-ops, matching the rest of DummySocket.


fredr (approved):
Thanks!

PR #2423, opened 2026-05-04
cli/cmd/encore: new 'encore check' command

eandre:
This adds a new `encore check` command that starts the app and runs a list of curl commands against it.

This is useful for LLM agents to be able to test that the app works without having to deal with `encore run` in the background.

### DIFF
--- a/cli/cmd/encore/check.go
+++ b/cli/cmd/encore/check.go
@@ -1,60 +1,295 @@
 package main
 
 import (
+	"bufio"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/signal"
+	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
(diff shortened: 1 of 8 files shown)


cli/cmd/encore/check.go
```diff
@@ -1,60 +1,294 @@
...
-)
+var checkTimeout time.Duration
 
-var checkCmd = &cobra.Command{
-	Use:   "check",
-	Short: "Checks your application for compile-time errors using Encore's compiler.",
+func init() {
+	checkCmd := &cobra.Command{
+		Use:   "check [spec]",
+		Short: "Runs a list of HTTP requests against a fresh app instance",
```
ekerfelt, line 25: Nit, but probably worth mentioning that the app is shutdown after checks

ekerfelt (dismissed):

ekerfelt (approved):

PR #2428, opened 2026-05-07
cli: render OpTracker as plain lines on non-interactive stderr

eandre:
Updated version of #2421 with correct protoc versions. Thanks @dstotijn!


fredr (approved):
🚢

fredr (approved):

PR #2436, opened 2026-05-08
clientgen: add optional CallParameters to generated TypeScript endpoint methods

0x4A756E65:
Generated TypeScript endpoint methods don't currently accept `CallParameters`, so there's no way to pass an `AbortSignal` through the generated client. Raw endpoints already support this — this brings typed endpoints to parity.

Without this, when a user navigates away from a page in a SPA, in-flight requests can't be canceled. The backend continues processing the abandoned request and logs "context canceled" errors when the response tries to write to a closed connection.

### Before / After

```typescript
// Before — no way to pass options
await client.svc.GetStatus(req)

// After — optional CallParameters as trailing argument
await client.svc.GetStatus(req, { signal: controller.signal })
```

### Changes

- `pkg/clientgen/typescript.go`: non-raw, non-stream endpoints now accept an optional trailing `options?: CallParameters`, forwarded to `callTypedAPI`. When headers/query are present, options are spread first (`{...options, headers, query}`) so endpoint-specific values take precedence.
- 8 golden test files updated to match.
- `sharedTypes`, stream, and raw code paths are unchanged.
- Fully backwards-compatible.

### DIFF
--- a/pkg/clientgen/testdata/goapp/expected_baseauth_typescript.ts
+++ b/pkg/clientgen/testdata/goapp/expected_baseauth_typescript.ts
@@ -120,15 +120,15 @@ export namespace svc {
         /**
          * DummyAPI is a dummy endpoint.
          */
-        public async DummyAPI(params: Request): Promise<void> {
-            await this.baseClient.callTypedAPI("POST", `/svc.DummyAPI`, JSON.stringify(params))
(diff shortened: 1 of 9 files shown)


0x4A756E65:
Any interest here Encore team?

PR #2440, opened 2026-05-11
cli/daemon/mcp: synchronous Pub/Sub probe + retry-aware call_endpoint

simon-johansson:
## Summary

Adds two MCP tools to eliminate the "async-introspection loop" failure mode AI agents hit when working with Encore Pub/Sub — where the agent publishes a message, polls a read endpoint, sees empty results, then burns turns introspecting via `get_pubsub`/`get_metadata`/`get_traces`.

- **`wait_for_subscription_message`** — synchronous probe that blocks until the next message on a topic is fully processed by a subscription (handler returns or errors), then returns the outcome (handler error message, message payload, duration, trace ID). Reuses the existing `trace2.Store` event listener via a new fan-out broker so we register exactly one listener regardless of concurrent MCP calls.
- **`retry_until` on `call_endpoint`** — optional parameter that makes the daemon poll the endpoint server-side until a predicate matches, instead of the agent looping turn-by-turn. Predicates: `status` (HTTP code), `body_path` (dot-path JSON equality), `body_jq` (minimal subset: `.path | length <op> N` and `.path` truthy).

Backwards compatible: when `retry_until` is omitted, `call_endpoint` behaves byte-for-byte as before (pinned by regression test).

## What's in here

- `cli/daemon/mcp/trace_broker.go` — single listener registered with `trace2.Store`, fan-out to per-call subscribers, hardened against close-then-subscribe and double-close
- `cli/daemon/mcp/pubsub_wait.go` — `wait_for_subscription_message` handler with topic/sub validation, since-watermark dedup, shallow payload match, auto-start of the app
- `cli/daemon/mcp/predicate.go` — predicate evaluator (status, body_path, body_jq subset)
- `cli/daemon/mcp/api_tools.go` — refactored `callEndpoint` into `ensureAppRunning` + `singleShotCall` + `stringifyBody`, then wrapped in `runRetryLoop`
- 57 new unit tests across the broker, payload match, span details, wait orchestration (happy/timeout/since), topic validation, all predicate forms, retry parsing/loop (including transient status flips and timeout shapes)

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./cli/daemon/mcp/...` passes all 57 tests
- [x] `go vet ./cli/daemon/mcp/...` clean
- [x] Backwards-compat regression test for `call_endpoint` without `retry_until`
- [ ] **Manual smoke test (reviewer):** run an Encore app with a Pub/Sub topic + subscription; verify
  - publish → `wait_for_subscription_message` returns `matched: true` with handler outcome
  - subscriber that throws returns `outcome: "error"` with the error message
  - non-existent topic returns the validation error
  - `call_endpoint` without `retry_until` returns the legacy shape
  - `call_endpoint` with `retry_until: { predicate: { body_jq: ".events | length > 0" } }` polls server-side until data arrives
  - `fail_on_timeout: true` returns `{error, last_response, retry: {...}}`

## Known follow-ups (not blocking)

- Retry loop currently fails fast on transient `singleShotCall` errors (connection refused mid-loop kills retry). Worth a `retry_on_transient` flag.
- Broker drops events silently when a slow subscriber's 16-slot buffer fills. Counter or larger buffer would help under load.
- `match` (top-level eq, wait tool) and `body_path` (deep eq, retry predicate) overlap conceptually — could unify if a third filter feature lands.
- Build-tagged e2e scaffold at `pubsub_wait_e2e_test.go` is a `t.Skip` placeholder; fill in when a Pub/Sub fixture is wired into the daemon e2e harness.

### DIFF
--- a/cli/daemon/mcp/api_tools.go
+++ b/cli/daemon/mcp/api_tools.go
@@ -37,7 +37,51 @@ func (m *Manager) registerAPITools() {
 		mcp.WithString("auth_token", mcp.Description("Optional authentication token to include in the request. This is used for endpoints that require authentication.")),
 		mcp.WithString("auth_payload", mcp.Description("Optional authentication payload in JSON format. This is used for custom authentication schemes.")),
 		mcp.WithString("correlation_id", mcp.Description("Optional correlation ID to track the request through the system. Useful for debugging and tracing.")),
-		mcp.WithObject("retry_until", mcp.Description("Optional. If set, the MCP server repeatedly calls the endpoint until the predicate matches OR the timeout elapses, then returns the final response. Predicate forms: status (int), body_path ({path, equals}), body_jq (subset: '<path>' or '<path> | length <op> N'). Other fields: timeout_ms (int, default 10000), interval_ms (int, default 250), fail_on_timeout (bool, default false). Accepts a JSON object or a JSON-encoded string.")),
+		mcp.WithObject("retry_until",
+			mcp.Description("Optional. If set, the MCP server repeatedly calls the endpoint until the predicate matches OR the timeout elapses, then returns the final response. Accepts a JSON object or a JSON-encoded string."),
+			mcp.Properties(map[string]any{
+				"predicate": map[string]any{
+					"type":        "object",
+					"description": "Predicate the response must satisfy to stop retrying. Exactly one of status, body_path, or body_jq should be set; precedence is body_jq > body_path > status.",
+					"properties": map[string]any{
+						"status": map[string]any{
+							"type":        "integer",
+							"description": "Stop when the HTTP response status equals this code (e.g. 200).",
+						},
+						"body_path": map[string]any{
+							"type":        "object",
+							"description": "Stop when the value at the given dot-path equals the expected JSON value.",
+							"properties": map[string]any{
+								"path": map[string]any{
+									"type":        "string",
+									"description": "Dot-path into the JSON response body (e.g. '.events.0.orderID').",
+								},
+								"equals": map[string]any{
+									"description": "Expected value at the path (any JSON value).",
+								},
+							},
+							"required": []string{"path", "equals"},
+						},
+						"body_jq": map[string]any{
+							"type":        "string",
+							"description": "Minimal jq subset: '<path> | length <op> N' where op is one of > >= == != < <=, or '<path>' for a truthy check.",
+						},
+					},
+				},
+				"timeout_ms": map[string]any{
+					"type":        "integer",
+					"description": "Max total wait in milliseconds. Default 10000.",
+				},
+				"interval_ms": map[string]any{
+					"type":        "integer",
+					"description": "Delay between retries in milliseconds. Default 250.",
+				},
+				"fail_on_timeout": map[string]any{
+					"type":        "boolean",
+					"description": "If true, return an MCP error on timeout instead of the last response. Default false.",
+				},
+			}),
+		),
 	), m.callEndpoint)
 
 	// Add tool for getting all services and endpoints
--- a/cli/daemon/mcp/api_tools.go
+++ b/cli/daemon/mcp/api_tools.go
@@ -3,6 +3,7 @@ package mcp
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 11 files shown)


simon-johansson:
## Manual smoke test results — both tools work end-to-end ✅

Verified against the `automativity-ztf2` app via the daemon's MCP SSE endpoint. All 9 tests in the reviewer checklist passed.

### Setup confirmation

- `tools/list` now reports **20 tools** (was 19) — `wait_for_subscription_message` registered.
- `call_endpoint` input schema includes the new `retry_until` property (description mentions `status` / `body_path` / `body_jq` predicate forms, `timeout_ms` / `interval_ms` / `fail_on_timeout`, and "Accepts a JSON object or a JSON-encoded string").

### Test results

| # | Test | Result |
|---|------|---|
| 1 | `wait_for_subscription_message` against a non-existent topic | ✅ `{"error":"topic \"nope-...\" not found in current app"}` |
| 2 | `wait_for_subscription_message` timeout (no publish, `timeout_ms:2000`) | ✅ `{matched:false, timeout:true, topic, waited_ms:2000, messages_seen_during_wait:0}` — wall-clock 2.12s |
| 3 | `call_endpoint` without `retry_until` (backwards-compat) | ✅ Legacy shape returned: `{body, status, status_code, trace_id}` — no `retry` key |
| 4 | `call_endpoint` `retry_until: {predicate: {status: 200}}` (matches immediately) | ✅ `retry: {matched:true, attempts:1, total_wait_ms:12}` |
| 5 | `retry_until` `body_jq: ".bits | length > 0"` against a response with `.Bits` (PascalCase) | ✅ `matched:false` (correctly — wrong key); attempts:12 over 3026ms, body returned for inspection |
| 5b | Same but with the right key `.Bits | length > 0` | ✅ `matched:true, attempts:1, total_wait_ms:56` |
| 6 | `retry_until` `fail_on_timeout:true` predicate that never matches | ✅ `{error: "retry_until predicate never matched within 1500ms", last_response, retry}` — 1.69s for a 1500ms budget |
| **7** | **Real publish → wait, subscriber handler errors** | ✅ **The big one — see below** |
| 8 | `wait_for_subscription_message` `match: {requestID: "WANTED-B"}` across two publishes | ✅ Returned the WANTED-B message; noise-A was correctly skipped |

### Test 7 — handler-error visibility (the case this PR was designed for)

`POST /releaser/webhook` with a garbage `run_id: "NOT_A_NUMBER"` — the subscriber handler calls `strconv.ParseInt` on it and returns an `errs.B().Cause(err).Msg(...)` error. The wait tool returned in **0.66s** with:

```json
{
  "matched": true,
  "topic": "release-webhook",
  "subscription": "subscription",
  "handler": {
    "outcome": "error",
    "duration_ms": 0,
    "error": "unknown code: failed to parse run id: strconv.ParseInt: parsing \"NOT_A_NUMBER\": invalid syntax"
  },
  "message": {
    "id": "d80tqkv6nm701vfn5md0",
    "published_at": "2026-05-11T13:44:51.177768Z",
    "delivery_attempt": 1,
    "payload": { "event": "workflow_run", "data": {"run_id":"NOT_A_NUMBER",...}, "requestID":"probe-fire", ... }
  },
  "trace_id": "cot93c9ov5pli3lmrhg3pute10"
}
```

The literal Go error string from the subscriber handler is propagated end-to-end through the broker, message payload is fully deserialized, `delivery_attempt`, `published_at`, and `trace_id` all populated. This is exactly the loop that was burning 80–120 turns / $0.66–$1.35 per cell on `*/03-pubsub-events` prompts in the benchmark — collapsed to one call.

### Minor note (not a blocker)

The MCP SSE query parameter is `?app=<id>`, not `?appID=<id>` — my first attempt used the wrong key and got `app not found in context`. Worth a mention in user-facing docs if there isn't already one; `encore mcp start` itself prints the right URL.

### Recommendation

Ship. Implementation matches the spec on every acceptance criterion including the tricky bits (handler-error visibility, `since` dedup, `match` shallow-key filter, two distinct response shapes for `fail_on_timeout` true/false, byte-compat on `call_endpoint` without `retry_until`).

cli/daemon/mcp/api_tools.go
```diff
@@ -34,6 +37,7 @@ func (m *Manager) registerAPITools() {
 		mcp.WithString("auth_token", mcp.Description("Optional authentication token to include in the request. This is used for endpoints that require authentication.")),
 		mcp.WithString("auth_payload", mcp.Description("Optional authentication payload in JSON format. This is used for custom authentication schemes.")),
 		mcp.WithString("correlation_id", mcp.Description("Optional correlation ID to track the request through the system. Useful for debugging and tracing.")),
+		mcp.WithObject("retry_until", mcp.Description("Optional. If set, the MCP server repeatedly calls the endpoint until the predicate matches OR the timeout elapses, then returns the final response. Predicate forms: status (int), body_path ({path, equals}), body_jq (subset: '<path>' or '<path> | length <op> N'). Other fields: timeout_ms (int, default 10000), interval_ms (int, default 250), fail_on_timeout (bool, default false). Accepts a JSON object or a JSON-encoded string.")),
```
ekerfelt: There are ObjectProperties you can use to define the object fields. Will be more standardized than this ad-hoc description in Description.

ekerfelt (approved):

PR #2444, opened 2026-05-12
Docs: generate TS runtime API reference with TypeDoc

simon-johansson:
## Summary

Adds a TypeDoc-based pipeline that generates per-module API reference MDX under `docs/ts/runtime/`, wired into the existing `#EncoreTS` menu as an accordion under the "Development" section. CI regenerates and verifies the committed output stays in sync with JSDoc on the runtime source.

### Pipeline shape

- `runtimes/js/encore.dev/typedoc.json` — TypeDoc + `typedoc-plugin-markdown` + `typedoc-plugin-frontmatter`, covering all 14 public entry points (`encore.dev`, `encore.dev/api`, `encore.dev/pubsub`, ..., `encore.dev/storage/sqldb`, etc.). `gitRevision: "main"` is pinned so source links are reproducible across CI runs.
- `runtimes/js/encore.dev/scripts/postprocess-docs.mjs` — runs after typedoc. Responsibilities:
  - File renames (`pubsub.mod.mdx` → `pubsub.mdx`, `storage.cache.mod.mdx` → `storage-cache.mdx`)
  - Injects per-page `title:` frontmatter (`encore.dev/pubsub`, etc.) matching the docs site convention
  - Strips `.mdx` from internal links so they match the file-based router
  - Drops the runtime API index page (the accordion makes it redundant)
  - Wraps each top-level symbol in `<!-- symbol-start: Name -->` / `<!-- symbol-end -->` markers for the renderer to pair into a `<SymbolCard>` wrapper
  - Collapses single-line `\`\`\`ts` fences to inline backticks (drops trailing semicolons; sidesteps MDX-v1's JSX-in-fence parser on trivial signatures)
  - Limits source links to top-level symbols (H3) and methods (H5 ending in `()`, or literally `Constructor`) — properties, parameters, enum values, and type parameters no longer carry a per-line source link
  - Inlines the surviving source link into the heading itself, as a `<SymbolSource href="..." />` component the www repo will register and style
- `docs/menu.cue` — new "Encore API Reference" accordion under "Development", matching the shape of the existing "ORMs" accordion. 14 entries (one per public entry point).
- `.github/workflows/ci.yml` — new `ts_runtime_docs` job runs `npm run docs` and `git diff --exit-code -- docs/ts/runtime` so PRs fail if a JSDoc change wasn't accompanied by a regenerate. Also adds `repo-token` to all `arduino/setup-protoc` invocations (drive-by fix for unrelated rate-limit flakes).

To regenerate locally:
```
cd runtimes/js/encore.dev && npm install && npm run docs
```

## Changes inside `runtimes/js/encore.dev/`

TypeDoc only reads `/** */` JSDoc, so the docs pipeline doesn't work without source changes. The bulk of the runtime edits fall into four buckets, all driven by what TypeDoc surfaced as warnings or what looked broken in the generated output:

### 1. JSDoc migration (`//` → `/** */`)

Roughly 50 exported symbols across `api/`, `app_meta.ts`, `req_meta.ts`, `pubsub/`, `storage/sqldb/`, `storage/objects/`, `storage/cache/` had `//`-style comments above them. TypeDoc ignores `//` comments, so without the migration the generated pages would show signatures with no prose. The comment **text** is preserved verbatim — the migration is mechanical, just changing the comment delimiter so TypeDoc picks it up.

### 2. Re-exports for types leaking through public signatures

A type that appears in a public method/field signature but isn't itself exported from the module's `mod.ts` shows up in the docs with a broken link. The fix is to add the type to the existing `export type { ... }` block:

- `storage/sqldb/mod.ts`: `Connection`, `Transaction`, `Primitive`, `SQLMigrationsConfig`
- `storage/objects/mod.ts`: `ListEntry`, `SignedDownloadUrl`, `SignedUploadUrl`, and the six `*Options` types
- `pubsub/mod.ts`: `RetryPolicy`
- `api/mod.ts`: `ErrDetails`, `HttpStatusValues`
- `mod.ts` (root): `HostedService`, `BaseRequestMeta`

These types were always part of the public API surface in practice — users could reach them through the methods they appeared on. The PR makes that explicit.

### 3. Promoting genuinely-public types out of `internal/`

Three symbols were already re-exported as part of the public API (`encore.dev/api` exposed `RawRequest`/`RawResponse`; `encore.dev/pubsub` exposed `DurationString`), but their definitions lived in `internal/` which the doc pipeline excludes. Result: type names in the docs with no documentation page to follow. The fix is to move the source files into their public homes:

- `internal/types/mod.ts` → `types/mod.ts` (DurationString — the file's only content)
- `internal/api/node_http.ts` → `api/node_http.ts` (RawRequest, RawResponse, plus their private DummySocket helper)

Five import paths updated (`cron/mod.ts`, `pubsub/subscription.ts`, `pubsub/mod.ts`, `api/mod.ts`, `internal/appinit/mod.ts`). No runtime behavior change.

### 4. `@internal` tags on truly-private machinery

A handful of types are referenced from public signatures but are deliberately not user-facing — type-level branding machinery and compile-time-only utility types. These get `/** @internal */` JSDoc tags so TypeDoc filters them out of the generated docs even when surfaced in type signatures (the `excludeInternal: true` config option handles the filtering).

## Outcome of the `intentionallyNotExported` discussion

The earlier draft of this PR asked reviewers to weigh in on 7 entries in TypeDoc's `intentionallyNotExported` list. After feedback, the list resolved as follows:

| Symbol | Verdict | What changed |
|---|---|---|
| `DurationString` | Promoted to public | Moved from `internal/types/mod.ts` → `types/mod.ts`; now part of `encore.dev/types` and `encore.dev/pubsub` with full doc page |
| `RawRequest` | Documented in place | Already exported from `encore.dev/api`; moved source from `internal/api/node_http.ts` → `api/node_http.ts` so TypeDoc can scan it |
| `RawResponse` | Same as above | Same |
| `brandedAttribute` | Kept private | Type-level brand for `Attribute<T>` — users only ever write `Attribute<number>`, never the brand |
| `allBrandedTypes` | Kept private | Same brand machinery |
| `HandlerFn` | Kept private | Internal alias for the `api()` handler shape; not user-facing |
| `StringLiteral` | Kept private | Compile-time constraint used to require literal-string names in `secret()` |

One new addition: `durationComponent`, the private template-literal helper that backs `DurationString`. Same shape as the other branding-machinery entries.

Net effect: the list shrunk from 7 → 5 entries, and the three user-facing types (`DurationString`, `RawRequest`, `RawResponse`) now have real doc pages instead of dangling type references.

## Renderer-side dependencies (www repo)

Two component slots need to be registered in the docs site's MDX provider:

- `<SymbolSource href="..." />` — renders inline next to symbol/method headings. The encore repo emits the URL; the www repo decides the visual treatment (icon, position, hover state).
- `<!-- symbol-start: Name --> ... <!-- symbol-end -->` markers — paired by a remark plugin into a `<SymbolCard symbol="Name">` wrapper around each top-level symbol.

The www team has these tracked separately.

## Test plan

- [x] `npm run docs` runs locally with 0 warnings, 0 errors
- [x] Generated MDX renders correctly when inspected
- [x] `cue vet docs/menu.cue` passes
- [x] `arduino/setup-protoc` rate-limit issue fixed
- [ ] Verify the new menu entries render in the docs site preview
- [ ] Verify intra-page anchors (e.g. `/ts/runtime/pubsub#topic`) resolve correctly
- [ ] Verify `<SymbolSource />` renders once the www repo registers the component
- [ ] CI's `ts_runtime_docs` job passes on this PR


runtimes/js/encore.dev/storage/sqldb/database.ts
```diff
@@ -310,6 +316,11 @@ export class SQLDatabase extends BaseQueryExecutor {
   }
 }
 
+/**
+ * Represents a database transaction.
+ *
+ * Make sure to always call `rollback` or `commit` to prevent hanging transactions.
```
eandre: Suggest using the AsyncDisposable pattern instead, whenever possible

eandre (approved):

PR #2447, opened 2026-05-13
share cache keyspaces between services

fredr:


proto/encore/parser/meta/v1/meta.proto
```diff
@@ -61,6 +61,7 @@ message Service {
   bool has_config = 6; // true if the service has uses config
   repeated BucketUsage buckets = 7; // buckets this service uses
   repeated string metrics = 8; // metrics this service uses
+  repeated string cache_clusters = 9; // cache clusters this service connects to
```
eandre: Hmm, it would be better if we track which keyspaces a service uses instead of which cache clusters

fredr: yeah, sure, I'll change!

fredr: I've updated to keyspace tracking per service now

PR #2471, opened 2026-06-04
docs/go/self-host: fix inverted databases map-key and name descriptions

### LINKED ISSUE #2461: Docs: self-host infra config `databases` map key and `name` descriptions are inverted vs runtime behavior
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What part(s) of Encore does this bug report apply to?

- [ ] Encore.ts (TypeScript)
- [x] Encore.go (Go)
- [ ] Encore CLI
- [ ] Local Development Dashboard
- [ ] Encore Cloud

### Current behavior

On the [Configure Infrastructure](https://encore.dev/docs/go/self-host/configure-infra#6-sql-database-configuration) docs (section *6. SQL Database Configuration*), the `databases` **map key** and the **`name`**
  field are described with inverted meanings vs. what the runtime actually does.

  Docs say ([configure-infra.md#L303-L304](https://github.com/encoredev/encore/blob/79a1afbbae09117368ab521cf3595aac7199c794/docs/go/self-host/configure-infra.md?plain=1#L303-L304)):
  > - `my-database`: This is the name of the database as it is declared in your Encore app.
  > - `name`: The name of the database on the database server. Defaults to the declared Encore name.

  But the runtime does the opposite — the **map key** is the physical DB name used in the connection string, and **`name`** is the Encore-declared name:

  - [`parse.go#L243-L247`](https://github.com/encoredev/encore/blob/79a1afbbae09117368ab521cf3595aac7199c794/runtimes/go/appruntime/exported/config/parse.go#L243-L247): `EncoreName:
  orDefault(db.Name, dbName)` (the `name` field), `DatabaseName: dbName` (the map key).
  - [`config.go#L348-L349`](https://github.com/encoredev/encore/blob/79a1afbbae09117368ab521cf3595aac7199c794/runtimes/go/appruntime/exported/config/config.go#L348-L349): `EncoreName //
  the Encore name`, `DatabaseName // the actual database name as known by the SQL server`.
  - [`db.go#L151-L153`](https://github.com/encoredev/encore/blob/79a1afbbae09117368ab521cf3595aac7199c794/runtimes/go/storage/sqldb/db.go#L151-L153): `dbName = db.DatabaseName` →
  `dbname=%s` in the DSN (the map key is the connection target).
  - [`manager_internal.go#L84`](https://github.com/encoredev/encore/blob/79a1afbbae09117368ab521cf3595aac7199c794/runtimes/go/storage/sqldb/manager_internal.go#L84): app code resolves a
  DB by `EncoreName` (the `name` field).

  When `name` is omitted it defaults to the key, so key and `name` coincide and the inversion is invisible (as in the overview example's `my-db` block). The detailed example in section 6 (`my-database` with `"name": "my-postgres-db-name"`) uses *different* values — which under the runtime connects to `my-database` (the key) and matches the declared DB by `my-postgres-db-name` (the `name`), i.e. the reverse of the descriptions directly beneath it.

  **Practical impact:** because the connection-target dbname *is* the map key (`DatabaseName`), there is no way to map multiple Encore-declared databases onto the same physical database.
  Two declared databases that should share one physical DB would both need the same map key, which a JSON object cannot represent (keys must be unique); the differing `name`/`EncoreName` does not change the connection target. This makes schema-based separation (several logical databases sharing one Postgres database, isolated by schema) inexpressible via the infra config.

### Minimum reproduction code

_No response_

### Steps to reproduce

1. Declare a database in app code: `var db = sqldb.NewDatabase("users", ...)`.
2. Following the docs, write infra config with the map key as the *declared* name and `name` as the *server* name:
     ```jsonc
     "databases": {
       "users": {            // docs: "name as declared in your Encore app"
         "name": "myapp_users" // docs: "name of the database on the database server"
       }
     }
3. Run self-hosted with this infra config.
4. Observe: Encore connects with dbname=users (the map key), not myapp_users, and matches the declared DB by the name field — i.e. the reverse of the docs. Following the docs literally connects to the wrong database.

### Expected behavior

The documentation should match the implemented behavior: the `databases` **map key** is the physical database name on the server (what goes into the connection `dbname=`), and **`name`** is the Encore-declared database name that matches `sqldb.NewDatabase(...)` / `sqldb.Named(...)`, defaulting to the key.

  Additionally, there should be a way to map multiple Encore-declared databases onto a single physical database — e.g. a field for the connection-target dbname that is independent of the (unique) map key — so that several logical databases can share one Postgres database and be isolated by schema. Today this is impossible, since the connection target is the map key and keys must be unique.

### Encore CLI version

1.56.6

### Node.js version

_No response_

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

Confirmed present on `main` @ `79a1afbbae09117368ab521cf3595aac7199c794`. Runtime module `encore.dev v1.52.1`.

harsha-cpp:
fixes #2461.

**problem:** the docs described `my-database` (the map key) as the Encore-declared name and `name` (the field) as the physical server name. the runtime does the opposite — the map key is the physical database name (used as `dbname=` in the DSN), and the `name` field is the Encore-declared name.

**fix:** swap the two descriptions to match runtime behavior.

### DIFF
--- a/docs/go/self-host/configure-infra.md
+++ b/docs/go/self-host/configure-infra.md
@@ -300,8 +300,8 @@ There must be exactly one database configuration for each declared database. You
 }
 ```
 
-- `my-database`: This is the name of the database as it is declared in your Encore app.
-- `name`: The name of the database on the database server. Defaults to the declared Encore name.
+- `my-database`: The name of the database on the database server.
+- `name`: The name of the database as declared in your Encore app. Defaults to the map key name.
 - `host`: SQL server host, optionally including the port.
 - `tls_config`: TLS configuration for secure connections. If the server uses TLS with a non-system CA root, or requires a client certificate, specify the appropriate fields as PEM-encoded strings. Otherwise, they can be left empty.
 - `databases`: List of databases, each with connection settings.


harsha-cpp:
can anyone review this and gimme remarks

ekerfelt (approved):

ekerfelt:
Looks good. The real issue is that the mapping is done incorrectly, but we can't easily change it without breaking backwards compatibility. I think updating the docs makes sense for now and we might revisit later to fix the mapping issue.

harsha-cpp:
Cool, is there anything i can do?

PR #2479, opened 2026-06-11
Add support for external secret vaults

ekerfelt:


proto/encore/runtime/v1/infra.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.10
-// 	protoc        v6.32.1
+// 	protoc-gen-go v1.36.11
+// 	protoc        v7.34.0
```
fredr: could you use the same versions as before so we dont get lots of diffs when switching back and forth?

ekerfelt: ah, annoying. yeah, I'll try to downgrade

proto/encore/runtime/v1/infra.proto
```diff
@@ -22,6 +22,29 @@ message Infrastructure {
...
+// GCP). Explicit credentials may be added later via additional fields here.
+message SecretProvider {
+  // The unique resource id for this provider.
+  string rid = 1;
+
+  // Human-readable identifier, used in logs.
+  string encore_name = 2;
+
+  oneof provider {
+    GCPSecretManager gcp_sm = 10;
```
fredr, line 42: is this suppose to be number 10? or was that a typo?

ekerfelt: It's a pattern we use for other oneofs. I assumed it was because we wanted to have the oneofs sequential, but I don't have a strong opinion

fredr: ah didn't know! then all good!

proto/encore/runtime/v1/infra.pb.go
fredr: I can run it if you want, I have my versions locked :key:

Cargo.lock
```diff
@@ -373,6 +367,28 @@ dependencies = [
  "zeroize",
 ]
 
+[[package]]
+name = "aws-lc-rs"
```
fredr: this might cause problems in zigbuild, see https://github.com/encoredev/encore/pull/2201

that issue referenced have been fixed though, so might not be an issue anymore

proto/encore/runtime/v1/infra.pb.go
ekerfelt: Fixed

Cargo.lock
ekerfelt: Removed it again.

proto/encore/runtime/v1/infra.pb.go
fredr: 🫸️🫷️

fredr (approved):
🚢

PR #2492, opened 2026-06-22
fix(db): clarify error message when invalid environment name is passed to db shell

### LINKED ISSUE #2252: Encore CLI database shell command gives confusing error message
If you provide an incorrect environment name when trying to use the Encore CLI database shell command, an misleading error message is returned stating "database appDB not found".   

In my case, the Encore Cloud environment is "staging" and Development is a type of environment.  This is compounded by the fact that if you look at secrets, the name "Development" is an environment.  The value "staging" doe create the shell and connect to the correct DB

My ask is that If an incorrect environment name is given the CLI should respond with a clearer error message to highlight the user's mistake.

```sh
bwehrle@mac encore-test % encore db shell appDB --env=development -v               
psql: error: connection to server at "127.0.0.1", port 58103 failed: FATAL:  database appDB not found
8:19AM FTL psql failed error="exit status 2"
```
```sh
bwehrle@mac encore-test % encore secrets list                                
Secret Key             Production   Development   Local   Preview   Specific Envs   
APISecretKey1           ✗            ✓             ✓       ✗                         
APISecretKey2           ✗            ✓             ✓       ✗                         
```

venturefm639:
## Problem
When an incorrect environment name is passed via `--env` to `encore db shell`, 
the CLI returns a misleading error message blaming the database name instead of 
the environment name.

For example, passing `--env stagng` (typo) would produce:
    could not connect to db myapp: environment not found

This incorrectly implies the database is missing when the real problem 
is the environment name.

## Fix
Check the gRPC status code returned by the daemon. When `codes.NotFound` 
is returned, report that the environment name was not found instead of 
the database, and suggest the correct usage.

The fallback error message now also includes both the database name and 
environment name for full context.

Fixes #2252

### DIFF
--- a/cli/cmd/encore/db.go
+++ b/cli/cmd/encore/db.go
@@ -138,8 +138,15 @@ when using tools like Prisma.
 			Namespace:   nonZeroPtr(nsName),
 			Role:        getDBRole(),
 		})
-		if err != nil {
-			fatalf("could not connect to db  %s: %v", dbName, err)
+		if err != nil { // if there is an error
+			st , ok :=status.FromError(err) // we check the error statud what kind of erro and where is it coming from
+			if ok && st.Code()== codes.NotFound{ // IF THE gRPC daemon status (st) is equals to NOT FOUND and 
+			// the ok (boolean) returns true
+			fatalf("Environment %q not found you can check your environment name and try again ",dbEnv) 
+			// now it shows where the exact error is coming from instead of wrongly pointing to the databse 
+
+			}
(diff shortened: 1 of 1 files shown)


ivancernja:
Duplicate of: https://github.com/encoredev/encore/pull/2301

venturefm639:
"Thanks for the review  sir ,
 I can see #[their PR number] takes a more thorough approach by validating the environment upfront via ListEnvs rather than handling the error after the fact. Closing this in favour of that fix. I learned a lot from reading the codebase on this one."

PR #2496, opened 2026-06-26
dockerbuild: exclude volatile pnpm metadata from the deps layer

Zei33:
## What

pnpm writes bookkeeping files into `node_modules` whose contents change on every install even when the installed dependency set is identical:

- `node_modules/.modules.yaml` records a `prunedAt` timestamp (and a machine-specific `storeDir`)
- `node_modules/.pnpm-workspace-state-v1.json` records a `lastValidatedTimestamp`

This change skips those files when building the image so the dependency layer's digest stays stable across installs.

## Why

Since #2483 the image is split into layers by change frequency, with `node_modules` in its own dependency layer so that "unchanged layers keep identical digests across builds and can be skipped when pushing and pulling images". For pnpm apps that benefit never materialises: because `.modules.yaml` changes on every install, the dependency layer gets a new digest on every build, so the whole `node_modules` layer is re-uploaded on `--push` and re-pulled on deploy even when nothing changed. For large pnpm trees this dominates deploy time.

These files are written directly inside a `node_modules` directory and are only consumed by the package manager during a later install; nothing reads them at runtime. Dropping them from the image makes the dependency layer reproducible across installs without affecting the running app, which is the same rationale as the existing always-on `.git` exclusion.

## How

`CopyDir` skips a small denylist of pnpm metadata files, matched only when they sit directly inside a `node_modules` directory, so a coincidentally-named file shipped deep inside a package is left untouched:

- `.modules.yaml`, `.pnpm-workspace-state.json`, `.pnpm-workspace-state-v1.json`

## Tests

- `TestBuildImage` fixture now includes the pnpm metadata files and asserts they are absent from the dependency layer.
- `TestBuildImage_DeterministicDepsLayerAcrossInstalls` builds twice with differing `.modules.yaml` content and asserts the dependency layer digest is identical. It extends `TestBuildImage_ReproducibleLayers`, which only varied the build time.
- `TestIsVolatilePnpmMetadata` covers the path matching, including the immediate-parent guard.

Context: https://github.com/pnpm/pnpm/issues/9474, https://github.com/pnpm/pnpm/issues/11669

### DIFF
--- a/pkg/dockerbuild/tarcopy.go
+++ b/pkg/dockerbuild/tarcopy.go
@@ -208,6 +208,32 @@ func nodeModulesPath(relPath HostPath) (within, isRoot bool) {
 	return false, false
 }
 
+// volatilePnpmMetadata are pnpm bookkeeping files written directly inside a
+// node_modules directory whose contents change on every install even when the
+// installed dependencies are identical: pnpm records a prunedAt timestamp in
+// .modules.yaml and a lastValidatedTimestamp in .pnpm-workspace-state-v1.json.
+// They are not read at runtime, and including them gives the dependency layer a
+// new digest each build. That defeats layer caching, so the whole node_modules
+// layer is re-pushed and re-pulled on every deploy.
+// See https://github.com/pnpm/pnpm/issues/9474.
+var volatilePnpmMetadata = map[string]bool{
+	".modules.yaml":                 true, // volatile prunedAt + machine-specific storeDir
+	".pnpm-workspace-state.json":    true, // older pnpm: lastValidatedTimestamp
+	".pnpm-workspace-state-v1.json": true, // current pnpm: lastValidatedTimestamp
+}
+
+// isVolatilePnpmMetadata reports whether relPath is a volatilePnpmMetadata file
+// located directly inside a node_modules directory (so a coincidentally-named
+// file shipped deep inside a package is not skipped).
+func isVolatilePnpmMetadata(relPath HostPath) bool {
+	components := strings.Split(string(relPath.ToUnix()), "/")
+	n := len(components)
+	if n < 2 {
+		return false
+	}
+	return components[n-2] == "node_modules" && volatilePnpmMetadata[components[n-1]]
+}
+
 // shouldInclude returns true if the path should be included in the tar.
 func shouldInclude(desc *dirCopyDesc, path HostPath) bool {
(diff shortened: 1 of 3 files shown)


fredr (changes requested):
Thanks for opening this pr. Lets keep it as minimal as possible and only fix the pnpm issue by ignoring the problematic pnpm files. The rest I believe should not be volatile.

Zei33:
@fredr thanks for the tip. I've altered the PR and code to be specific to pnpm. 

I also altered the name of the exclusion functions, variables and tests (`volatileDepMetadata` -> `volatilePnpmMetadata` style). This narrows the scope of the functionality being implemented. We can always rename it later if we want to expand the scope to include some future package manager that has a similar issue. Yarn and npm related inclusions have been removed.

fredr:
@Zei33 Sounds good! The diff doesn't reflect those changes though, have you pushed them?

Zei33:
@fredr Sorry that's my bad, I forgot to stage my changes before amending the PR 😳

That should all be sorted now. Let me know if anything else is amiss.

pkg/dockerbuild/tarcopy.go
```diff
@@ -208,6 +208,32 @@ func nodeModulesPath(relPath HostPath) (within, isRoot bool) {
...
+var volatilePnpmMetadata = map[string]bool{
+	".modules.yaml":                 true, // volatile prunedAt + machine-specific storeDir
+	".pnpm-workspace-state.json":    true, // older pnpm: lastValidatedTimestamp
+	".pnpm-workspace-state-v1.json": true, // current pnpm: lastValidatedTimestamp
+}
+
+// isVolatilePnpmMetadata reports whether relPath is a volatilePnpmMetadata file
+// located directly inside a node_modules directory (so a coincidentally-named
+// file shipped deep inside a package is not skipped).
+func isVolatilePnpmMetadata(relPath HostPath) bool {
```
fredr, line 225: I think this should be possible to write a bit cleaner, something like this maybe:

```go
func isVolatilePnpmMetadata(relPath HostPath) bool {
    s := string(relPath)
    return volatilePnpmMetadata[filepath.Base(s)] && filepath.Base(filepath.Dir(s)) == "node_modules"
}
```

fredr (approved):

eandre (approved):

PR #2520, opened 2026-07-24
[daemon] Fix unbounded memory/fd growth from eager watching, codegen-regen, and leaked runs

djeer:
## Problem

Running `encore` locally against a large monorepo (~200 services, multiple git
worktrees checked out alongside the main clone) causes the daemon
(`encore daemon`) to leak memory and file descriptors without bound over the
life of the process:

- RAM usage climbing to **25-50GB+**, forcing macOS into heavy swap
- File descriptor exhaustion (`too many open files`) even when running
  `encore run --watch=false` and `encore check`/`encore test` - commands that
  shouldn't need to watch anything, also causing  `read tcp: connection reset by peer` errors everywhere
- None of this recovers by restarting the daemon: the very next boot
  reproduces the same cost, because the set of app roots the daemon considers
  "known" is persisted in sqlite (`~/Library/Application Support/encore/encore.db`,
  table `app`) and survives across restarts.

We profiled the daemon live (`pprof` on the `debug` HTTP endpoint) against a
real large monorepo and traced this to three separate, compounding causes.

## Root causes

**1. File watching was eager, not lazy.**
`apps.Manager.resolve()` called `Instance.beginWatch()` unconditionally for
every app root it resolved - for *any* command, not just `encore run`. Since
resolved instances are cached for the daemon's entire lifetime, every git
worktree ever `encore check`ed or `encore test`ed accumulated a permanent
recursive fsnotify watch that was never released. `encore run --watch=false`
didn't help either: that flag only ever gated the live-reload subscription
(`Manager.watch(run)`), not the underlying watch itself, which the daemon
starts unconditionally as soon as an app is resolved.

**2. Codegen regeneration ran for every app root the daemon has ever seen, on every boot.**
`Server.watchApps()` (run once at daemon startup) calls `WatchAll`, which
calls `Manager.List()` - which iterates *every app root in the sqlite `app`
table* and resolves each one. Each resolution fires an `appListeners`
callback (`regenerateUserCode`) that does a full parse of the app to
regenerate `encore.gen.go` and update `.gitignore`. With 14 worktrees/
workspaces tracked from past sessions, every single daemon boot triggered 14
full monorepo parses before a single user command ran - each costing
multiple GB.

**3. `run.Manager.runs` never released exited runs.**
Every `encore run` invocation adds a `*Run` entry to `mgr.runs`, keyed by a
unique ID. Nothing ever removed an entry once its process exited - confirmed
by the fact that every consumer of the map (`FindRunByAppID`) has to manually
skip over dead entries via a `select { case <-run.Done(): }` check, rather
than the map only containing live runs. Each `*Run` holds its own `Builder`,
whose internal `pkginfo.Loader` package cache stays reachable for as long as
the entry survives. Restarting the same app 5 times grew daemon memory
linearly and permanently (~1.5GB per restart, ~7.7GB total, never released).

## Fix

Three commits, each addressing one of the above:

- **`daemon: make file watching lazy, only for apps running with live-reload`**
  Removed the eager `beginWatch()` call from `resolve()`. `beginWatch()` is
  now reachable only through `Instance.Watch()`, which is only called via
  `Manager.watch(run)`, which is only invoked when `Params.Watch` is true
  (i.e. `encore run` without `--watch=false`). `encore check`, `encore test`,
  and `encore run --watch=false` now never hold a watch open.

- **`daemon: gate codegen-regen behind active Watch(), matching file-watching`**
  Moved the `appListeners` notification (which triggers `regenerateUserCode`)
  from `resolve()` into `beginWatch()`, so it fires under the exact same
  condition as file-watching. Apps that are only ever checked/tested no
  longer trigger a background parse at all, and the startup `WatchAll` pass
  no longer reparses every historically-tracked app root - only apps actively
  being run with live-reload do.

  Verified this doesn't affect correctness: `encore check`/`encore run` never
  read the codegen-regen output back (`bld.Parse()` is always called fresh);
  the only consumers of the underlying metadata cache are the MCP server
  tools and the local dev dashboard, which we confirmed via a full grep of
  all callers.

- **`daemon: return parse garbage to the OS, and stop leaking exited runs`**
  Two independent fixes: (a) `debug.FreeOSMemory()` after the parse/compile
  pass completes in both `run.go`'s `buildAndStart` and `check.go`'s `Check`,
  so the large amount of short-lived AST/type-checking data is returned to
  the OS immediately rather than waiting on the background scavenger -
  relevant on macOS, where pages the scavenger hasn't reclaimed yet still
  count against the process's reported memory footprint (and system memory
  pressure, i.e. what actually pushed us into swap). (b) Added
  `Manager.removeRun`, called at both points a run's `exited` channel closes,
  so `mgr.runs` no longer accumulates dead entries forever.

## Testing

All measurements taken against our actual monorepo (~200 services), using
`footprint` (macOS's `phys_footprint`, the same metric Activity Monitor
reports) rather than `ps`'s RSS, since Go's scavenger on Darwin uses
`madvise(MADV_FREE)`, which `ps` stops counting as resident before the kernel
has actually reclaimed the pages, but which `phys_footprint` (correctly)
still charges to the process.

Each fix targets a separate, non-overlapping code path, so we could isolate
and verify each one independently by measuring with only the relevant fix(es)
applied. The table below reports each root cause's own before/after in
isolation - these numbers are **not cumulative with each other** (e.g. the
codegen-regen row and the leaked-runs row were measured at different points
in the rollout, so don't add them together):

| Root cause | Before | After |
|---|---|---|
| #1: fsnotify watch held for every resolved app root | dozens+ watch fds per worktree, growing forever | 1 fd, held only while an app is actively `run` with live-reload |
| #2: codegen-regen for every tracked app root, on every daemon boot | 6-9GB at every boot (14 tracked roots) | 209MB |
| #3: `mgr.runs` never releasing exited runs | ~1.5GB per restart, unbounded (7.7GB after 5 restarts of the same app) | ~1.4GB regardless of restart count |

Screenshot 2026-07-24 at 08 18 12

---
<a href="https://app.blacksmith.sh/encoredev/codesmith/encore/pr/2520"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787492088&installation_model_id=427204&pr_number=2520&repository=encoredev%2Fencore&return_to=https%3A%2F%2Fgithub.com%2Fencoredev%2Fencore%2Fpull%2F2520&signature=0b662fd968e48c424e4e68fef7f008bc27cb29e1474f8d129da5b4807c4f559c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/cli/daemon/apps/apps.go
+++ b/cli/daemon/apps/apps.go
@@ -244,9 +244,10 @@ func (mgr *Manager) resolve(appRoot string) (*Instance, error) {
 	i := NewInstance(appRoot, man.LocalID, platformID)
 	i.tutorial = man.Tutorial
 	i.mgr = mgr
-	if err := i.beginWatch(); err != nil && !errors.Is(err, fs.ErrNotExist) {
-		log.Error().Err(err).Str("id", i.PlatformOrLocalID()).Msg("unable to begin watching app")
-	}
+	// Note: we deliberately don't start the file watcher here. Watching is
+	// lazy, started only via Watch() (i.e. only for apps actually running
+	// with live-reload enabled), so that commands like `encore check` and
+	// `encore run --watch=false` never hold a recursive fsnotify watch open.
 	mgr.instances[appRoot] = i
 
 	// Notify any listeners about the new app
--- a/cli/daemon/apps/apps.go
+++ b/cli/daemon/apps/apps.go
@@ -174,7 +174,7 @@ func (mgr *Manager) listRoots() ([]string, error) {
 }
 
 // RegisterAppListener registers a callback that gets invoked every time
-// an app is tracked.
+// an app starts being actively watched (i.e. run with live-reload enabled).
 func (mgr *Manager) RegisterAppListener(fn func(*Instance)) {
 	mgr.instanceMu.Lock()
 	defer mgr.instanceMu.Unlock()
@@ -183,9 +183,22 @@ func (mgr *Manager) RegisterAppListener(fn func(*Instance)) {
 	mgr.appListeners = append(mgr.appListeners, fn)
 	mgr.appRegMu.Unlock()
 
-	// Call the handler for all existing apps
+	// Call the handler for all apps that are already being watched.
 	for _, inst := range mgr.instances {
-		fn(inst)
+		if inst.watcher != nil {
+			fn(inst)
+		}
+	}
+}
+
+// notifyAppListeners invokes all registered app listeners for i.
+func (mgr *Manager) notifyAppListeners(i *Instance) {
+	mgr.appRegMu.Lock()
+	listeners := mgr.appListeners
+	mgr.appRegMu.Unlock()
+
+	for _, fn := range listeners {
+		fn(i)
 	}
 }
 
@@ -244,16 +257,13 @@ func (mgr *Manager) resolve(appRoot string) (*Instance, error) {
 	i := NewInstance(appRoot, man.LocalID, platformID)
 	i.tutorial = man.Tutorial
 	i.mgr = mgr
-	// Note: we deliberately don't start the file watcher here. Watching is
-	// lazy, started only via Watch() (i.e. only for apps actually running
-	// with live-reload enabled), so that commands like `encore check` and
-	// `encore run --watch=false` never hold a recursive fsnotify watch open.
+	// Note: we deliberately don't start the file watcher (or notify app
+	// listeners) here. Both are lazy, triggered only via Watch() (i.e. only
+	// for apps actually running with live-reload enabled), so that commands
+	// like `encore check` and `encore run --watch=false` never hold a
+	// recursive fsnotify watch open, and never trigger a codegen-regen parse
+	// for apps that aren't being actively developed against.
 	mgr.instances[appRoot] = i
-
-	// Notify any listeners about the new app
-	for _, fn := range mgr.appListeners {
-		fn(i)
-	}
 
 	return i, nil
 }
@@ -481,6 +491,13 @@ func (i *Instance) beginWatch() error {
 				}
 			}
 		}()
+
+		// Now that this app is actively watched (i.e. running with
+		// live-reload), notify listeners so they can e.g. regenerate
+		// user-facing codegen and keep it fresh as files change.
+		if i.mgr != nil {
+			i.mgr.notifyAppListeners(i)
+		}
 
 		return nil
 	})
--- a/cli/daemon/apps/apps.go
+++ b/cli/daemon/apps/apps.go
@@ -205,7 +205,8 @@ func (mgr *Manager) notifyAppListeners(i *Instance) {
 // WatchFunc is the signature of functions registered as app watchers.
 type WatchFunc func(*Instance, []watcher.Event)
 
-// WatchAll watches all apps for changes.
+// WatchAll registers fn to receive change events from every app that is
+// being actively watched, now or in the future.
 func (mgr *Manager) WatchAll(fn WatchFunc) error {
 	err := mgr.setupWatch.Do(func() error {
 		// Begin tracking all known apps by calling List (since it calls resolve).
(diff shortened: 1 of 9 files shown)


eandre:
Thanks very much! Note that this is more of a behavior change than a bugfix; the behavior is sort of intentional to allow for code generation to happen immediately, even if `encore run` is not running. That said, it's clearly become a bigger problem with AI-centric workflows involving many workspaces/worktrees. Additionally, code generation is somewhat less important for Encore.go than it is for Encore.ts, so you may not be feeling that pain as much since you use Encore.go. 

All that said, I think I'm okay with going this direction, but with a few tweaks. The `--watch` flag is really more intended to be about live reload, so I think we should still have file watching running while `encore run --watch=false` is running so that code regeneration still happens, but I'm okay with de-registering the file watcher after `encore run` exits. Does that sound like a reasonable compromise?

djeer:
@eandre Thanks - that makes sense, this would be OK for agentic workflows in my case - I usually only do `encore check` in the worktrees, and run a single instance of the app in the main checkout, when I need to validate something manually. I've hit another descriptors leak bug when implementing, but it looks good now.

----

**`94e2a08`** - `Manager.watch` takes a `liveReload` param: every `encore run` now watches, and the flag only gates `run.Reload()`. So `--watch=false` still gets codegen regeneration, just no app reload. To allow de-registering on exit, the watcher lifetime is refcounted (`beginWatch`/`endWatch`) instead of a one-shot `syncutil.Once`; the existing `<-run.Done(); Unwatch(sub)` goroutine already drives teardown. `check`/`test` still never watch.

**`d9c9b86`** - while verifying teardown, descriptors weren't released despite `Close()` being called. Not an Encore bug: fsnotify's kqueue backend marks the watcher closed *before* removing its watches, and `remove()` starts with `if w.isClosed() { return nil }` - so `unix.Close(info.wd)` never runs. It also opens an fd per watched *file*, not just per directory. Repros standalone on v1.8.0 and v1.9.0, 1818 dirs:

```
after watching:             13,224
after Close():              13,221   <- nothing released
after explicit Remove()s:       13   <- released
```

Fix: remove the watches ourselves before closing, while still open (we already track every watched dir)

cli/daemon/apps/apps.go
```diff
@@ -437,52 +456,92 @@ func (i *Instance) Unwatch(id WatchSubscriptionID) {
...
+	if i.mgr != nil {
+		i.mgr.notifyAppListeners(i)
+	}
+
+	return nil
+}
+
+// endWatch releases one reference on the app's file watcher, closing the
+// underlying watcher once the last reference has been released. Safe to
+// call even if beginWatch never succeeded.
```
eandre: This seems incorrect, in that if you call this when `beginWatch` fails it will result in the ref count going out of sync

djeer: Replaced the refcount with subscription map

eandre (approved):
Thanks!

djeer:
I now realised `--watch=false` is in fact `--autoreload=false`, but renaming is a breaking change :(

eandre:
I suppose we could add it and alias the old one and hide/deprecate it

PR #2522, opened 2026-07-27
feat(sqldb): make Postgres Docker image configurable via environment …

### LINKED ISSUE #2212: Configuring Postgres version for local development
Currently `encore run...` defaults to creating a docker container based on `encoredotdev/postgres:15`. Now that Postgres 18 has been released it'd be good to look at providing more recent versions of the database.

Happy to help with the implementation of such feature if that helps and I can be pointed in the right direction.

venturefm639:
…variable

Previously, the Postgres Docker image used for local development was hardcoded as a constant:

    const Image = "encoredotdev/postgres:18"

This meant developers had no way to match their local Postgres version to their production environment, which could cause subtle bugs that only appear in production.

This change replaces the constant with a function GetPostgresImage() that reads from the ENCORE_POSTGRES_IMAGE environment variable, falling back to the default image if not set.

Usage:
    ENCORE_POSTGRES_IMAGE=encoredotdev/postgres:17 encore run

If the environment variable is not set, behaviour is unchanged and the default image encoredotdev/postgres:18 is used.

Fixes #2212

---
<a href="https://app.blacksmith.sh/encoredev/codesmith/encore/pr/2522"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787757223&installation_model_id=427204&pr_number=2522&repository=encoredev%2Fencore&return_to=https%3A%2F%2Fgithub.com%2Fencoredev%2Fencore%2Fpull%2F2522&signature=54b767f321f52497404c65139ca94223ee95e339b6c3feb2e7ee9d579d8f4145"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/cli/cmd/encore/db.go
+++ b/cli/cmd/encore/db.go
@@ -159,7 +159,7 @@ when using tools like Prisma.
 				}
 			}
 
-			cmd = exec.Command("docker", "run", "-it", "--rm", "--network=host", docker.Image, "psql", dsn)
+			cmd = exec.Command("docker", "run", "-it", "--rm", "--network=host", docker.GetPostgresImage(), "psql", dsn)
 		}
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
--- a/cli/daemon/sqldb/docker/docker.go
+++ b/cli/daemon/sqldb/docker/docker.go
@@ -144,7 +144,7 @@ func (d *Driver) CreateCluster(ctx context.Context, p *sqldb.CreateParams, log z
 		if p.Memfs {
 			args = append(args,
 				"--mount", "type=tmpfs,destination="+defaultDataDir,
-				Image,
+				GetPostgresImage(),
 				"-c", "fsync=off",
 			)
 		} else {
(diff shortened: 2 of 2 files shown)


eandre (commented):
We'd rather not make this an environment variable, as typically you want everyone working on the same application to use the same postgres version. Not sure exactly what is a better approach, though.

venturefm639:
"That's a fair point sir, an environment variable would let individual developers diverge from each other. Would configuring the image via encore.app make more sense? That way the version is committed to the repository and consistent across the whole team. Happy to rework the approach if so."      Also this seems to be a better logical approach

PR #2556, opened 2026-08-29
runtime/{core,js}: adds caller data into requests

murilopl:
Adds the name of the calling service to request metadata, for service-to-service calls.

The field is optional: it is only set when the caller is another API endpoint. It is not set when the request comes from outside the application (e.g. through an API gateway) or when the caller is unknown.

## TypeScript

`currentRequest()` now includes `callerService` on API call metadata:

```ts
import { currentRequest } from "encore.dev";

const meta = currentRequest();
if (meta?.type === "api-call") {
  // "service1" when called from service1, undefined for external requests
  console.log(meta.callerService);
}
```

## Go

`encore.CurrentRequest()` now includes `CallerService`:

```go
req := encore.CurrentRequest()
// "service1" when called from service1, "" for external requests
log.Println(req.CallerService)
```

Covered by e2e tests in both runtimes (`ts_app_test.go` and `testscript/encore_currentrequest.txt`), asserting the field is set on service-to-service calls and absent on external requests.

### DIFF
--- a/docs/ts/runtime/encore-dev.mdx
+++ b/docs/ts/runtime/encore-dev.mdx
@@ -19,6 +19,15 @@ Describes an API call being processed.
 
 Describes the API Endpoint being called.
 
+##### callerService?
+
+`optional callerService?: string`
+
+The name of the service that made the API call, for
+service-to-service calls. Not set when the request comes from
+outside the application (e.g. through an API gateway) or when
+the caller is unknown.
+
 ##### headers
 
 `headers: Record<string, string | string[]>`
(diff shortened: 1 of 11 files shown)


eandre:
Overall this is reasonable, but I think we should simplify it to only include the name of the service, if any. It also needs to be made optional since it's not guaranteed to be present, and we'd want to include support in Encore.go as well

murilopl:
> Overall this is reasonable, but I think we should simplify it to only include the name of the service, if any. It also needs to be made optional since it's not guaranteed to be present, and we'd want to include support in Encore.go as well

Comment addressed.

